PR #15, opened 2020-10-29
Implement Custom Transformer to Model Deployment

ariefrahmansyah:
This PR allows users to deploy their Transformer Docker image that contains pre/post-processing code.

Changes:
- [x] API
  - [x] Append Transformer Spec to InferenceService Spec
  - [x] Update Model Version Endpoint API (Deploy, Redeploy, Get, List)
  - [x] Add unit test
  - [x] Update Swagger spec
- [x] UI - Update Model Version Deployment page to include Transformer configuration
- [x] SDK - Update `deploy()` function to accept new optional arg: `transformer`
  - [x] Add E2E test

Model version deployment page preview:

transformer-deploy


api/cluster/controller.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 type Controller interface {
-	Deploy(modelService *models.Service) (*models.Service, error)
+	Deploy(modelService *models.Service, transformer models.Transformer) (*models.Service, error)
```
pradithya, line 35: Can we put `models.Transformer` as a field within `models.Service`?

ui/src/version/deployment/components/EndpointDeployment.js
```diff
@@ -24,15 +24,18 @@ import {
   EuiFlexItem,
```
pradithya, line 24: Can you provide gif and/or screenshot for the UI changes?

pradithya:
You'll need to include end to end test to the to do list

ui/src/version/deployment/components/EndpointDeployment.js
ariefrahmansyah: Added to the PR description

ariefrahmansyah:
> You'll need to include end to end test to the to do list

Yes. The e2e tests will be included in Python SDK test

api/client/model_transformer.go
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package client
+
+import (
+	"time"
+)
+
+type Transformer struct {
+	Enabled         bool             `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
ariefrahmansyah: The enabled flag allows users to switch the Transformer on/off while maintaining the previous configuration value from the UI. For example, I've deployed a transformer with image X and min replica 2. Then I disable it. Next time, when I want to enable the transformer, I will still have the previous configuration (image X and min replica 2).

pradithya:
After a model+transformer is deployed, is there a way to see the configuration in UI?

ariefrahmansyah:
> After a model+transformer is deployed, is there a way to see the configuration in UI?

Great question. There's no special page for displaying model version configuration, but we can use the redeploy page to display it. Something like this:

image

We can have separate PR for dedicated model version page.

pradithya:
> > After a model+transformer is deployed, is there a way to see the configuration in UI?
> 
> Great question. There's no special page for displaying model version configuration, but we can use the redeploy page to display it. Something like this:
> 
> image
> 
> We can have separate PR for dedicated model version page.

Can you create an issue for this?

api/client/model_transformer.go
```diff
@@ -0,0 +1,25 @@
...
+	Image           string           `json:"image,omitempty"`
+	Command         string           `json:"command,omitempty"`
```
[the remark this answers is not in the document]
ariefrahmansyah: The main reason I choose to use string is that I'm using [Text FIeld](https://elastic.github.io/eui/#/forms/form-controls#text-field) instead of [Combo Box](https://elastic.github.io/eui/#/forms/combo-box) as Combo Box doesn't allow us to input the same value more than once. Although not often, there may be a case when we need to specify the same argument twice. So I decide to let the user write the command and args as one string. The splitting from string to an array of string happens at the creation of Transformer spec.

api/models/transformer.go
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+
+package models
+
+import "github.com/google/uuid"
+
+// Transformer is a service for pre/post-processing steps.
+type Transformer struct {
+	ID                string    `json:"id"`
+	Enabled           bool      `json:"enabled"`
```
pradithya, line 22: Can you remove the `Enabled` field?

ariefrahmansyah: I've answered this question here: https://github.com/gojek/merlin/pull/15#discussion_r513907426

Do you have a better suggestion regarding enabling our user to reuse their previous transformer config?

charts/merlin/values.yaml
```diff
@@ -23,6 +23,9 @@ merlin:
   apiHost: http://merlin.mlp/v1
   docsURL: ""
 
+  # Comma-separated value of Docker registries that can be chosen in deployment page
+  dockerRegistries: ghcr.io/gojek,ghcr.io/your-company
```
pradithya, line 27: I am not sure about this feature. I am thinking we should just ask user the fully qualified docker image name in the UI instead of have a suggestion as combo box. 
It's one less feature and configuration to maintain.

python/sdk/merlin/model.py
```diff
@@ -946,6 +955,43 @@ def deploy(self, environment_name: str = None,
...
+
+        target_env_vars = []
+        if transformer.env_vars is not None:
+            if not isinstance(transformer.env_vars, dict):
+                raise ValueError(
+                    f"transformer.env_vars should be dictionary, got: {type(transformer.env_vars)}")
+
+            if len(transformer.env_vars) > 0:
+                for name, value in transformer.env_vars.items():
+                    target_env_vars.append(client.EnvVar(name, value))
```
pradithya, line 988: Can we handle properly when the `name` and `value` is not string? I think we should add string casting here.

api/models/transformer.go
pradithya: Oh got it, I missed that reply

charts/merlin/values.yaml
ariefrahmansyah: This feature makes sure the user only uses the supported Docker registry by Merlin. It prevents the failure of Image pulling if the model cluster doesn't have the credential to access it. Turing also has this feature and by having this in Merlin we can provide the same experience across MLP. If the user wants to use another Docker registry, he/she needs to ask the administrator to add it to Merlin and to make sure the model cluster can access it (by adding Image Pull Secret). For default, we have public Docker Hub + dockerRegistries specified here.

python/sdk/merlin/model.py
ariefrahmansyah: Updated

charts/merlin/values.yaml
pradithya: Ok got it

api/cluster/resource.go
```diff
@@ -165,6 +181,67 @@ func createPredictorSpec(modelService *models.Service, config *config.Deployment
 	return predictorSpec
 }
 
+func createTransformerSpec(modelService *models.Service, transformer *models.Transformer, config *config.DeploymentConfig) *kfsv1alpha2.TransformerSpec {
```
pradithya, line 184: Can we also specify target concurrency of the transformer?
The reason is that many of transformer will contain IO code (fetch from feast, db, etc) and it's wasteful to have target concurrency = 1.

ariefrahmansyah: Unfortunately, we cannot do it at this moment. The annotation for configuring the target concurrency is specified in InferenceService that will be inherited by both Predictor and Transformer.

pradithya: Let's keep this in mind, especially when upgrading the KFserving

pradithya:
LGTM

pradithya (approved):

PR #21, opened 2020-11-03
Automatically generate version in GitHub Actions

ariefrahmansyah:

This PR uses [vertagen](https://github.com/gojek/merlin/blob/main/scripts/vertagen/vertagen.sh) script to automatically generate a valid semver to tag our Docker images whether we're on a branch or a tag. It will help the development and testing process as we will have a new Docker image published with a new tag.

The current workflows publish Docker images using a branch name or tag. If we've introduced a new change in our branch during development, a new Docker image will still use the same branch name. This can lead to Kubernetes not pulling the new Docker image if we set ImagePullPolicy to not Always.


pradithya:
Can you add description about this PR?

PR #23, opened 2020-11-11
InferenceService URL as target host of Model VS

ariefrahmansyah:
Previously, we use InferenceService Predictor URL for Model Endpoint VirtualService target host when serving traffic to model version.
E.g., when we serve model version 1, we will have this URL (model-version-1-predictor-default.domain.com) as host.

The problem would arise if we deploy a model with a transformer. If we keep using Predictor URL as Host, we will skip sending requests to the Transformer. To fix it, we can do one of these:
1. Check if transformer exists, then use Transformer URL as Model Endpoint VirtualService host, or
2. Use InferenceService Host URL, e.g. (model-version-1.domain.com)

This PR uses option #2.

### DIFF
--- a/api/models/version_endpoint.go
+++ b/api/models/version_endpoint.go
@@ -74,6 +74,19 @@ func (e *VersionEndpoint) IsServing() bool {
 	return e.Status == EndpointServing
 }
 
+func (e *VersionEndpoint) HostURL() string {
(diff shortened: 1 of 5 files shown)


pradithya:
LGTM

pradithya (approved):

PR #25, opened 2020-11-12
Reduce required python version to run Merlin SDK

ariefrahmansyah:
Some of the clients are still using the old version of Python and it would be not straightforward for them to update their Python version.

```release-note
NONE
```


ariefrahmansyah:
I'm closing this PR as this solution can break other clients that already use Python 3.7 or newer.

PR #28, opened 2020-11-18
Fix disable merlin transformer

ariefrahmansyah:
There's an issue with the redeployment of a model version with an enabled transformer. When the user tries to disable a transformer using Model Version Endpoint Redeploy API, the model version will still have the transformer enabled. This is due to unhandled logic inside the API/

This PR fixes the issue describes above.

**Checklist**

- [x] Tested locally


pradithya (approved):
LGTM

PR #30, opened 2020-12-04
Upgrade KFServing, Istio Client

tiopramayudi:
1. Upgrade KFServing Client into v0.4.1 . Our current KFServing is old (v0.1.4)
2. Modify inference url returned by latest inferenceservice manifest. Latest inferenceservice's url remove `v1/models/{inferencename}` from the url, we need to append `v1/models/{inferencename} to returned url
3. Upgrade spark-on-k8s-operator library into version v1beta2-1.1.2-2.4.5 . We need to update this because previously it is depended on k8s.io library from version 1.13
4. Add pytest-dependency for sdk. It is required because current e2e test failing because dependency between test is not working
5. Upgrade istio to version 1.4.6

**Checklist**

- [x] Tested locally
- [x] E2E test run successful


api/cluster/resource.go
```diff
@@ -176,7 +176,7 @@ func createPredictorSpec(modelService *models.Service, config *config.Deployment
 	}
 
 	predictorSpec.DeploymentSpec = kfsv1alpha2.DeploymentSpec{
-		MinReplicas: modelService.ResourceRequest.MinReplica,
+		MinReplicas: &(modelService.ResourceRequest.MinReplica),
```
pradithya, line 179: Why is it the case?

api/istio/istio.go
```diff
@@ -84,8 +84,9 @@ func (c *client) PatchVirtualService(ctx context.Context, namespace string, vs *
 	if err != nil {
 		return nil, err
 	}
-
-	return c.networking.VirtualServices(namespace).Patch(vs.ObjectMeta.Name, types.MergePatchType, vsJSON)
+	xx, yy := c.networking.VirtualServices(namespace).Patch(vs.ObjectMeta.Name, types.MergePatchType, vsJSON)
+	return xx, yy
+	// return c.networking.VirtualServices(namespace).Patch(vs.ObjectMeta.Name, types.MergePatchType, vsJSON)
```
pradithya, line 89: Can you remove the commented code

api/istio/istio.go
```diff
@@ -84,8 +84,9 @@ func (c *client) PatchVirtualService(ctx context.Context, namespace string, vs *
 	if err != nil {
 		return nil, err
 	}
-
-	return c.networking.VirtualServices(namespace).Patch(vs.ObjectMeta.Name, types.MergePatchType, vsJSON)
+	xx, yy := c.networking.VirtualServices(namespace).Patch(vs.ObjectMeta.Name, types.MergePatchType, vsJSON)
```
pradithya, line 87: Maybe a better variable naming?

api/istio/istio_test.go
```diff
@@ -22,13 +20,17 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/gojek/merlin/istio/client-go/pkg/apis/networking/v1alpha3"
-	networkingv1alpha3 "github.com/gojek/merlin/istio/client-go/pkg/clientset/versioned/typed/networking/v1alpha3"
-	"github.com/gojek/merlin/istio/client-go/pkg/clientset/versioned/typed/networking/v1alpha3/mocks"
+	// "github.com/gojek/merlin/istio/client-go/pkg/clientset/versioned/typed/networking/v1alpha3/mocks"
```
pradithya, line 23: Remove dead code

pradithya (commented):
A few changes required, but overall LGTM.

api/cluster/resource.go
tiopramayudi: it needs pointer of int instead of int

api/istio/istio.go
tiopramayudi: it is not supposed to there, forgot to remove

Issue #29, opened 2020-11-26
Rework Fetching Model Version List


ariefrahmansyah:
## Is your feature request related to a problem? Please describe.

The current implementation of List Model Version API is using preloading GORM and without limitation. This would become an issue if we have a huge list of the model versions that can exceed Postgres' `max_stack_depth` configuration value. In most cases, users don't need to get all model versions and filtering or limitation is needed here.

It also affects the model version UI page since it's ended up fetching all of the model versions and displaying all of them at once.

## Describe the solution you'd like

In API side, it will be good to use basic SQL instead of relying to GORM. It will also be huge improvement to allow user to specify his own filter query.

In UI side, we can add pagination in model version table.

PR #40, opened 2020-12-21
Support API Pagination On Model Version API

tiopramayudi:
The current implementation of List Model Version API is using preloading GORM and without limitation. This would become an issue if we have a huge list of the model versions that can exceed Postgres' max_stack_depth configuration value. In most cases, users don't need to get all model versions and filtering or limitation is needed here.

It also affects the model version UI page since it's ended up fetching all of the model versions and displaying all of them at once.

This PR fixed partial of this issue https://github.com/gojek/merlin/issues/29

 Yes
***Screenshots***
Screen Shot 2021-01-07 at 07 54 12
Screen Shot 2021-01-07 at 07 54 57
**Checklist**

- [x] Tested locally


api/service/version_service.go
```diff
@@ -54,20 +71,72 @@ func (service *versionsService) query() *gorm.DB {
 		Select("versions.*")
 }
 
-func (service *versionsService) ListVersions(ctx context.Context, modelID models.ID, monitoringConfig config.MonitoringConfig) (versions []*models.Version, err error) {
-	err = service.query().
-		Where(models.Version{ModelID: modelID}).
-		Order("created_at DESC").
-		Find(&versions).
-		Error
+func decodeCursor(cursor string) (cursorPagination, bool) {
```
pradithya, line 74: It will be better to move this method to its own file/package, it will be reused for other resources pagination too.

pradithya: Can we make it generalized though? somehow now it dependens to `versionID` and `versionModelID`

python/sdk/merlin/model.py
```diff
@@ -333,12 +334,33 @@ def list_version(self) -> List['ModelVersion']:
 
         :return: list of ModelVersion
         """
+        result: List['ModelVersion'] = []
+        versions, cursor = self.list_version_pagination()
+        result = result + versions
+        while cursor != "":
+            versions, cursor = self.list_version_pagination(cursor=cursor)
+            result = result + versions
+            sleep(0.05)
```
pradithya, line 343: Why do we need sleep?

python/sdk/merlin/model.py
```diff
@@ -333,12 +334,33 @@ def list_version(self) -> List['ModelVersion']:
...
+        return result
+
+    def list_version_pagination(self, limit=DEFAULT_MODEL_VERSION_LIMIT, cursor="", search="") -> Tuple[List['ModelVersion'], str]:
```
pradithya, line 346: In what case this function will be useful for?

api/api/response.go
```diff
@@ -49,6 +54,14 @@ func Ok(data interface{}) *Response {
 	}
 }
 
+func OkWithCustomHeaders(data interface{}, headers map[string]string) *Response {
```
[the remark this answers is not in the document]
tiopramayudi: hmm because by default we write standard headers like Content-Type in the response. I'm ok with `OkWithHeaders`

api/client/model_alert_condition_metric_type.go
```diff
@@ -13,9 +13,9 @@ type AlertConditionMetricType string
 
 // List of AlertConditionMetricType
 const (
-	ThroughputAlertConditionMetricType AlertConditionMetricType = "throughput"
-	LatencyAlertConditionMetricType    AlertConditionMetricType = "latency"
-	ErrorRateAlertConditionMetricType AlertConditionMetricType = "error_rate"
-	CPUAlertConditionMetricType        AlertConditionMetricType = "cpu"
-	MemoryAlertConditionMetricType     AlertConditionMetricType = "memory"
+	THROUGHPUT_AlertConditionMetricType AlertConditionMetricType = "throughput"
```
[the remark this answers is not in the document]
tiopramayudi: I think it is generated by swagger client, I'm not sure why it uses those convention

api/client/model_model.go
```diff
@@ -14,13 +14,13 @@ import (
...
-	ProjectID          int32  `json:"project_id,omitempty"`
-	MlflowExperimentID int32  `json:"mlflow_experiment_id,omitempty"`
+	Id                 int32  `json:"id,omitempty"`
+	ProjectId          int32  `json:"project_id,omitempty"`
+	MlflowExperimentId int32  `json:"mlflow_experiment_id,omitempty"`
 	Name               string `json:"name,omitempty"`
 	// Model type
-	Type      string          `json:"type,omitempty"`
-	MlflowURL string          `json:"mlflow_url,omitempty"`
+	Type_     string          `json:"type,omitempty"`
```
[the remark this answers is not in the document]
tiopramayudi: it is also generated by swagger

api/service/version_service.go
tiopramayudi: sure, I'll try to generalized this approach

python/sdk/merlin/model.py
tiopramayudi: I used this for test, will remove this

python/sdk/merlin/model.py
tiopramayudi: it is called on `list_version` function, to list all the versions

api/client/model_alert_condition_metric_type.go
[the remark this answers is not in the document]
pradithya: I think it's due to swagger codegen version, previously I use 2.4.14 to generate the code

python/sdk/merlin/model.py
pradithya: In that case it will be better to make it private

api/api/response.go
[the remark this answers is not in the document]
pradithya: Yes please

api/api/versions_api.go
```diff
@@ -79,13 +80,24 @@ func (c *VersionsController) PatchVersion(r *http.Request, vars map[string]strin
 func (c *VersionsController) ListVersions(r *http.Request, vars map[string]string, _ interface{}) *Response {
 	ctx := r.Context()
 
+	var query service.VersionQuery
+	if err := decoder.Decode(&query, r.URL.Query()); err != nil {
```
pradithya, line 84: What will happen if `.Query()` is empty?

api/api/response.go
```diff
@@ -49,6 +61,14 @@ func Ok(data interface{}) *Response {
 	}
 }
 
+func OkWithHeaders(data interface{}, headers map[string]string) *Response {
```
pradithya, line 64: Can you add comment to the exported method `OkWithHeaders`

api/api/versions_api.go
tiopramayudi: it will use default value of VersionQuery struct

pradithya:
LGTM

pradithya (approved):

PR #42, opened 2021-01-06
Add local installation document

ariefrahmansyah:

Add documentation for local deployment of Merlin..

```release-note
NONE
```

**Checklist**

- [x] Updated documentation

### DIFF
--- a/mlflow/Dockerfile
+++ b/mlflow/Dockerfile
@@ -3,8 +3,10 @@ FROM python:3.7.2-slim
 RUN pip install google-cloud-storage
 RUN pip install psycopg2-binary
 
+ARG BOTO3_VERSION=1.7.12
 ARG MLFLOW_VERSION=1.3.0
 
+RUN pip install boto3==${BOTO3_VERSION}
 RUN pip install mlflow==${MLFLOW_VERSION}
 
 ENV BACKEND_STORE_URI="/data/mlruns"
--- a/scripts/quick_install.sh
+++ b/scripts/quick_install.sh
@@ -1,0 +1,230 @@
+#!/bin/sh
+
+export CLUSTER_NAME=dev
+
+export ISTIO_VERSION=1.5.4
+export KNATIVE_VERSION=v0.14.3
+export KNATIVE_NET_ISTIO_VERSION=v0.15.0
+export CERT_MANAGER_VERSION=v1.1.0
+export KFSERVING_VERSION=v0.4.0
+
+export VAULT_VERSION=0.7.0
+export MINIO_VERSION=7.0.2
+
+export OAUTH_CLIENT_ID="<put your oauth client id here>"
+export MERLIN_VERSION=v0.9.0
+
+# Install Istio
+curl --location https://git.io/getLatestIstio | sh -
+
+cat << EOF > ./istio-config.yaml
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  values:
+    global:
+      proxy:
+        autoInject: disabled
+      useMCP: false
+      jwtPolicy: first-party-jwt
+      k8sIngress:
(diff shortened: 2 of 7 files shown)


scripts/quick_install.sh
```diff
@@ -0,0 +1,230 @@
...
+          - path: /*
+            backend:
+              serviceName: mlp
+              servicePort: 8080
+EOF
+
+# Install Merlin
+kubectl create secret generic vault-secret --namespace=mlp --from-literal=address=http://vault.vault.svc.cluster.local --from-literal=token=root
+
+helm install merlin ../charts/merlin --namespace=mlp --values=../charts/merlin/values-e2e.yaml \
```
jmelinav, line 216: I followed the same steps to try it out on minikube with virtual box. I run into below error `/src/api/cmd/main.go:126 +0x7f7
{"level":"panic","ts":1610364390.0461166,"caller":"cmd/main.go:554","msg":"unable to retrieve secret for cluster dev from vault","stacktrace":"github.com/gojek/merlin/log.Panicf\n\t/src/api/log/log.go:56\nmain.initImageBuilder\n\t/src/api/cmd/main.go:554\nmain.main\n\t/src/api/cmd/main.go:126\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"}
panic: unable to retrieve secret for cluster dev from vault`. All other pods are in running state except this. Vault pod is running and I am able to a get the secret using the command `kubectl exec vault-0 --namespace=vault -- vault kv get secret/${CLUSTER_NAME}`. I also tried setting external vault but no luck. Any help is appreciated.

ariefrahmansyah: What's the output of this command: `export CLUSTER_NAME=dev && kubectl exec vault-0 --namespace=vault -- vault kv get secret/${CLUSTER_NAME}`?
It's expected that the output contains the Kubernetes cluster credentials

Otherwise, it's possible that the script fails to write the credential to vault (https://github.com/gojek/merlin/blob/local-dev-doc/scripts/quick_install.sh#L131-L147).

jmelinav: the command outputs the credentials.

jmelinav: Should the address in the below command have port as well? 
`kubectl create secret generic vault-secret --namespace=mlp --from-literal=address=http://vault.vault.svc.cluster.local --from-literal=token=root`

ariefrahmansyah: > Should the address in the below command have port as well?
No need. Merlin can connect to Vault.

Can you please re-run this part of scripts: https://github.com/gojek/merlin/blob/local-dev-doc/scripts/quick_install.sh#L131-L147 ?
Seems like there's a discrepancy on Vault secret engine that Merlin needs. After that, please verify if secret exist then try to redeploy Merlin pod (or reinstall Merlin helm).

jmelinav: Trying that. I will let you know in a while.

jmelinav: It works after re-deployment. Not sure what I had missed in the 1st step.

ariefrahmansyah: Nice!

Seems like the vault isn't ready yet when we run these (https://github.com/gojek/merlin/blob/local-dev-doc/scripts/quick_install.sh#L132-L133)

mlflow/Dockerfile
```diff
@@ -3,8 +3,10 @@ FROM python:3.7.2-slim
 RUN pip install google-cloud-storage
 RUN pip install psycopg2-binary
 
+ARG BOTO3_VERSION=1.7.12
```
pradithya, line 6: Why is it needed?

ariefrahmansyah: So MLflow can use minio as artifact storage

PR #45, opened 2021-01-12
Request/Response logging

tiopramayudi:
This PR aims to enable users deploy version endpoint with request/response logging feature. This PR only touch API and SDK implementation. UI implementation will be raised in separate PR

None

**Checklist**

- [x] Tested locally


pradithya (approved):
LGTM

api/models/logger.go
```diff
@@ -0,0 +1,69 @@
...
+	Model          *LoggerConfig `json:"model"`
+	Transformer    *LoggerConfig `json:"transformer"`
+}
+
+type LoggerConfig struct {
+	Enabled bool       `json:"enabled"`
+	Mode    LoggerMode `json:"mode"`
+}
+
+func GetLoggerMode(mode LoggerMode) kfsv1alpha2.LoggerMode {
```
ariefrahmansyah, line 44: nitpick: using GetLoggerMode is confusing for me when the object returned is KFServing's LoggerMode. i prefer to rename it to `ToKFServingLoggerMode()`

api/models/logger.go
```diff
@@ -0,0 +1,69 @@
...
+	modeMapping := map[LoggerMode]kfsv1alpha2.LoggerMode{
```
ariefrahmansyah, line 45: nitpick: we can move this map object outside of the function so it will only initialize once

ariefrahmansyah (approved):
I have some nitpick comments. But overall, LGTM

PR #47, opened 2021-01-19
Pyspark image builder issue

tiopramayudi:
Fix issues when building pyspark application image. There are 2 cases that trigger this issue:
1. Users use `protobuf` dependency which version is less than v3.12.0 in their model. Since currently we generate proto using protoc version above v3.13.0, it will cause error `AttributeError: module 'google.protobuf.descriptor' has no attribute '_internal_create_key'`.  In order to fix this we regenerate proto using older version of protoc(3.11.4)
2.  Users use `grpcio` version 1.22.0. It will cause error `AttributeError: module 'google.api_core' has no attribute 'gapic_v1'`, because automatically we will use latest `google-api-core` which has dependency rule `"grpc": "grpcio >= 1.29.0, < 2.0dev"`. Hence we need to pin `google-api-core` and `google-cloud-core` also `google-cloud-storage` 

Fixes #

NONE

```release-note

```

**Checklist**

- [x] Tested locally
- [-] Updated documentation
- [-] Update Swagger spec if the PR introduce API changes
- [-] Regenerated Golang and Python client if the PR introduce API changes


python/batch-predictor/requirements.txt
```diff
@@ -2,5 +2,5 @@ pyspark==2.4.5
 mlflow==1.6.0
 cloudpickle==1.2.2
 pyarrow==0.14.1
-merlin-sdk==0.9.1
+merlin-sdk @ git+https://github.com/gojek/merlin.git@pyspark-image-builder-issue#egg=merlin-sdk\&subdirectory=python/sdk
```
pradithya, line 5: Can we use relative directory instead?

pradithya (approved):
LGTM

PR #53, opened 2021-01-27
Feast Transformer implementation

ariefrahmansyah:

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/pkg/transformer/server/server.go
```diff
@@ -17,42 +21,87 @@ type Options struct {
...
+		return nil, err
+	}
 
-	fmt.Fprintf(w, string(out))
+	return body, nil
 }
 
 // Run serves the HTTP endpoints.
 func (s *Server) Run() {
+	// TODO: to be implemented
```
tiopramayudi, line 104: it will be implemented in the next PR?

api/pkg/transformer/server/server.go
```diff
@@ -17,42 +21,87 @@ type Options struct {
...
+}
 
-	// TODO: call postprocesshandler
+func (s *Server) predict(ctx context.Context, request []byte) ([]byte, error) {
+	req, err := http.NewRequest("POST", s.options.ModelPredictURL, bytes.NewBuffer(request))
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("Content-Type", "application/json")
```
tiopramayudi, line 86: I think you should pass the headers from the original request. Because if logging enabled, original request to transformer will has requestID headers generated which we need to pass to model prediction request

api/pkg/transformer/feast/transformer_test.go
```diff
@@ -0,0 +1,210 @@
...
+							},
+						},
+					},
+				},
+			},
+			want:    []byte(`{"driver_id":"1001","feast_features":{"driver_id":{"columns":["driver_trips:average_daily_rides"],"data":[1.1]}}}`),
+			wantErr: false,
+		},
+		{
+			name: "two configs: each retrieve one entity, one feature",
```
pradithya, line 96: Can we add more test case:
- 2 configs: 2 entities and 2 or more features
- Entity type that's not string
- The entity type in json body is different than the entity type (i.e. integer in the request, but the entity type is string)

And edge cases:
- Json path is incorrect (field is empty or not found)

.github/workflows/ci.yml
```diff
@@ -129,6 +129,20 @@ jobs:
           build_args: GITHUB_TOKEN=${{ secrets.DSP_REPO_TOKEN }}
           tag_with_ref: true
 
+  publish-merlin-transformer-docker:
```
pradithya, line 132: Can you add benchmark on the component that might take time? excluding external call though

api/pkg/transformer/feast/transformer.go
```diff
@@ -40,33 +43,22 @@ type FeastFeature struct {
...
-	feastFeatures := make(map[string]FeastFeature, len(f.config.TransformerConfig.Feast))
-
-	for _, config := range f.config.TransformerConfig.Feast {
-		// TODO: validate the config
-		log.Println(config)
+// Transform retrieves the Feast features values and add them into the request.
+func (t *Transformer) Transform(ctx context.Context, request []byte) ([]byte, error) {
+	feastFeatures := make(map[string]FeastFeature, len(t.config.TransformerConfig.Feast))
 
+	for _, config := range t.config.TransformerConfig.Feast {
```
pradithya, line 50: we can call it featureTable

api/pkg/transformer/feast/transformer.go
```diff
@@ -114,29 +104,44 @@ func (f *FeastTransformer) TransformHandler(w http.ResponseWriter, r *http.Reque
 		return nil, err
 	}
 
-	out, err := jsonparser.Set(body, feastFeatureJSON, "feast_features")
+	out, err := jsonparser.Set(request, feastFeatureJSON, "feast_features")
```
pradithya, line 107: Make `"feast_features"` constant

api/pkg/transformer/feast/transformer.go
```diff
@@ -82,19 +74,17 @@ func (f *FeastTransformer) TransformHandler(w http.ResponseWriter, r *http.Reque
 			Entities: entities,
 			Features: features,
 		}
+		t.logger.Debug("feast_request", zap.Any("feast_request", feastRequest))
 
-		feastResponse, err := f.feastClient.GetOnlineFeatures(ctx, &feastRequest)
+		feastResponse, err := t.feastClient.GetOnlineFeatures(ctx, &feastRequest)
```
pradithya, line 79: This will make the call to feast sequential. We should use scatter and gather pattern for the external call.

pradithya:
I think the critical part that are still missing:
- Making parallel call instead of sequential to feast
- Caching
- Batching
- Some edge cases testing

ariefrahmansyah (commented):
lgtm!

PR #65, opened 2021-02-04
Revert the validation function for validate endpoint deployment request

ariefrahmansyah:


ui/src/version/deployment/components/EndpointDeployment.js
```diff
@@ -79,12 +79,9 @@ const targetRequestStatus = currentStatus => {
 };
 
 const isRequestConfigured = request => {
-  if (!request.transformer || !request.transformer.enabled) {
-    return false;
-  }
-
   if (
-    request.transformer.transformer_type !== "standard" &&
+    request.transformer &&
```
tiopramayudi, line 83: what if user use `standard` transformer there will be no image parameter that will be passed right?

PR #69, opened 2021-02-04
Fix transformer.env_vars value if switching transformer type

ariefrahmansyah:

How to reproduce the issue:
1. Go to the existing deployed model with custom transformer redeployment page
2. On Transformer Type component, change from Custom Transformer to Standard Transformer
3. Then change back to Custom Transformer
4. Click deploy
5. The deployed transformer will lose all it's previous environment variables

This issue is because switching from custom to standard transformer will replace all existing environment variables with a standard transformer configuration.

**Checklist**

- [x] Tested locally


pradithya (approved):
Lgtm

PR #73, opened 2021-02-10
Refactor how to build feast request to support multiple entities

ariefrahmansyah:

The current implementation of the standard transformer cannot retrieve features for multiple entities at once due to an invalid FeastRequest schema.

For given entities config in transformer config:
```yaml
entities:
  - name: merchant_uuid
    valueType: STRING
    jsonPath: $.merchants[*].id
  - name: customer_id
    valueType: INT64
    jsonPath: $.customer_id
```

It will create this request:
```
"Entities": [
    {
      "customer_id": {
        "Val": {
          "Int64Val": 1111
        }
      }
    },
    {
      "merchant_uuid": {
        "Val": {
          "StringVal": "M111"
        }
      }
    }
  ],
```

This PR fixes this so the request will be like this:
```
"Entities": [
    {
      "customer_id": {
        "Val": {
          "Int64Val": 1111
        }
      },
      "merchant_uuid": {
        "Val": {
          "StringVal": "M111"
        }
      }
    }
  ],
```

Note that merchant_uuid is sibling of customer_id field instead of entire different feast row.

**Checklist**

- [x] Tested locally

### DIFF
--- a/api/pkg/transformer/feast/transformer.go
+++ b/api/pkg/transformer/feast/transformer.go
@@ -225,48 +225,35 @@ func (t *Transformer) getFeastFeature(ctx context.Context, request []byte, confi
 }
 
 func buildEntitiesRequest(request []byte, configEntities []*transformer.Entity) ([]feast.Row, error) {
-	allEntityRows := [][]feast.Row{}
-	for _, entity := range configEntities {
-		vals, err := getValuesFromJSONPayload(request, entity)
+	entities := []feast.Row{}
+
+	for _, configEntity := range configEntities {
+		vals, err := getValuesFromJSONPayload(request, configEntity)
 		if err != nil {
-			return nil, fmt.Errorf("unable to extract entity %s: %v", entity.Name, err)
+			return nil, fmt.Errorf("unable to extract entity %s: %v", configEntity.Name, err)
 		}
 
-		rows := []feast.Row{}
-		for _, val := range vals {
-			rows = append(rows, feast.Row{
-				entity.Name: val,
-			})
-		}
-		allEntityRows = append(allEntityRows, rows)
-	}
-
-	if len(allEntityRows) == 0 {
-		return nil, fmt.Errorf("no entity extracted")
-	}
-
-	entities := allEntityRows[0]
-
-	entityIndex := 1
-	for entityIndex < len(configEntities) {
-		entityName := configEntities[entityIndex].Name
-		rows := allEntityRows[entityIndex]
-
-		newEntities := []feast.Row{}
-		for _, entity := range entities {
-			for _, row := range rows {
-				newFeastRow := feast.Row{}
-				for key, value := range entity {
-					newFeastRow[key] = value
-				}
-
-				newFeastRow[entityName] = row[entityName]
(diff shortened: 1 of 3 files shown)


api/pkg/transformer/feast/transformer.go
```diff
@@ -233,6 +224,51 @@ func (t *Transformer) getFeastFeature(ctx context.Context, request []byte, confi
...
+		rows := allEntityRows[entityIndex]
+
+		newEntities := []feast.Row{}
+		for _, entity := range entities {
+			for _, row := range rows {
+				newFeastRow := feast.Row{}
+				for key, value := range entity {
+					newFeastRow[key] = value
+				}
+				newFeastRow[configEntities[entityIndex].Name] = row[configEntities[entityIndex].Name]
```
tiopramayudi, line 261: it is better to extract `configEntities[entityIndex].Name` into a variable

ariefrahmansyah: Can you elaborate? Are you saying we need validation as well?

tiopramayudi: something like this
```
entityName = configEntities[entityIndex].Name
newFeastRow[entityName] = row[entityName]

```

api/pkg/transformer/feast/transformer.go
```diff
@@ -233,6 +224,53 @@ func (t *Transformer) getFeastFeature(ctx context.Context, request []byte, confi
...
+	}
+
+	if len(allEntityRows) == 0 {
+		return nil, fmt.Errorf("no entity extracted")
+	}
+
+	entities := allEntityRows[0]
+
+	entityIndex := 1
+	for entityIndex < len(configEntities) {
```
pradithya, line 251: I think you can simplify this method by removing the second loop and include it in the first loop. It help avoiding too many tight loop.

ariefrahmansyah: Yup, you're right. I've simplified this logic to only have one loop.

tiopramayudi:
lgtm

api/pkg/transformer/feast/transformer_test.go
```diff
@@ -475,3 +476,510 @@ func TestTransformer_Transform(t *testing.T) {
 		})
 	}
 }
+
+func Test_buildEntitiesRequest(t *testing.T) {
```
pradithya, line 480: Can you add test that involves 3 entities? Want to see how it generalize

pradithya (approved):
LGTM

api/pkg/transformer/feast/transformer_test.go
ariefrahmansyah: Already added: https://github.com/gojek/merlin/pull/73/files/232cd0692aa1f1d2902c67757bae766b8681620c#diff-a6950993560c99c56bcc7cc2aaa7e80747301e01c16056044ee0703151f3022dR716-R914

I also add test for 4 entities: https://github.com/gojek/merlin/pull/73/files/232cd0692aa1f1d2902c67757bae766b8681620c#diff-a6950993560c99c56bcc7cc2aaa7e80747301e01c16056044ee0703151f3022dR916

PR #76, opened 2021-02-15
Improve standard transformer performance

ariefrahmansyah:

To have a more stable performance of standard transformer, I added or updated the following items:
1. Use gorilla/mux as router
2. Add HTTP server timeout
3. Add HTTP client timeout -- previously, we are using empty `&http.Client{}`. The HTTP client without timeout can hijack goroutines.
4. Add Jaeger tracing

```release-note
NONE
```

**Checklist**

- [x] Tested locally


pradithya (approved):
LGTM

PR #80, opened 2021-02-22
Geohash udf for Feast entity value extraction

khorshuheng:

jsonpath is not sufficient for use cases where the Feast entity values are derived from one or more fields in json. To achieve this, this PR introduce the concept of UDF. Geohash computation function is added to demonstrate how users can derived geohas from latitude and longitude found in the json body.

Fixes #

yes
```release-note
UDF is now supported for entity value extraction for Feast.
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


ariefrahmansyah:
Please add some test cases on `Test_buildEntitiesRequest` as well: https://github.com/gojek/merlin/blob/main/api/pkg/transformer/feast/transformer_test.go#L483

api/pkg/transformer/feast/transformer.go
```diff
@@ -86,13 +89,24 @@ func NewTransformer(feastClient feast.Client, config *transformer.StandardTransf
...
+				if err != nil {
+					return nil, fmt.Errorf("unable to compile jsonpath for entity %s: %s", configEntity.Name, configEntity.GetJsonPath())
+				}
+				compiledJsonPath[configEntity.GetJsonPath()] = c
+			case *transformer.Entity_Udf:
+				c, err := expr.Compile(configEntity.GetUdf(), expr.Env(UdfEnv{}))
+				if err != nil {
+					return nil, err
+				}
+				compiledUdf[configEntity.Name] = c
```
pradithya, line 107: Using entity name as the map key might cause issue when there are same entity but in different retrieval table.

api/pkg/transformer/feast/udf.go
```diff
@@ -0,0 +1,110 @@
...
+		floatValue = float64(o.(int))
+	default:
+		return 0, errors.New(fmt.Sprintf("%v cannot be parsed into float64", reflect.TypeOf(o)))
+	}
+	return floatValue, nil
+
+}
+
+func extractGeohash(jsonRequestBody interface{}, latitudeJsonPath, longitudeJsonPath string) (interface{}, error) {
+	latitude, err := jsonpath.JsonPathLookup(jsonRequestBody, latitudeJsonPath)
```
pradithya, line 63: `jsonpath.JsonPathLookup` will call json path Compile, it will be better if the compiled json path can be cached.

api/pkg/transformer/feast/udf.go
```diff
@@ -0,0 +1,110 @@
...
+		for _, long := range longitudes {
+			latFloat, err := toFloat64(lat)
+			if err != nil {
+				return nil, err
+			}
+			longFloat, err := toFloat64(long)
+			if err != nil {
+				return nil, err
+			}
+			geohashes = append(geohashes, geohash.Encode(latFloat, longFloat))
```
pradithya, line 92: `geohash.Encode` uses 12 character precision by default. I think we should allow user to specify the character precision. IIRC jaeger uses 7 characters.

api/pkg/transformer/feast/udf.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}
+
+	if len(geohashes) == 1 {
```
pradithya, line 96: I think we should maintain the type returned by `JsonPathLookup` for `latitude` or `longitude`. I.e. if it was `[]interface{}` then `extractGeohash` should return the same.
This is to avoid issue where there is only 1 entry in the original array.

api/pkg/transformer/feast/udf.go
```diff
@@ -0,0 +1,110 @@
...
+	if len(latitudes) == 0 {
+		return nil, errors.New("empty arrays of latitudes provided")
+	}
+
+	longitude, err := jsonpath.JsonPathLookup(jsonRequestBody, longitudeJsonPath)
+	if err != nil {
+		return nil, err
+	}
+	longitudes := toList(longitude)
+	if len(longitudes) == 0 {
```
pradithya, line 77: Might need to assert `len(latitudes) == len(longitudes)`

api/pkg/transformer/feast/udf.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, errors.New("empty arrays of longitudes provided")
+	}
+
+	geohashes := make([]string, 0)
+	for _, lat := range latitudes {
```
pradithya, line 82: This loop will process all combination of latitude and longitude. I don't think it's expected behaviour. Usually we treat lat and lon as a pair.

api/pkg/transformer/feast/udf_test.go
```diff
@@ -0,0 +1,98 @@
...
+				geohash.Encode(1.0, 2.0),
+				geohash.Encode(2.0, 2.0),
+			},
+		},
+		{
+			name:              "latitude and longitude arrays",
+			latitudeJsonPath:  "$.latitudeArrays",
+			longitudeJsonPath: "$.longitudeArrays",
+			expValue: []string{
+				geohash.Encode(1.0, 1.0),
```
pradithya, line 76: It should expect
```
				geohash.Encode(1.0, 1.0),
				geohash.Encode(2.0, 2.0),
```

api/pkg/transformer/feast/validator.go
```diff
@@ -52,8 +52,11 @@ func ValidateTransformerConfig(ctx context.Context, coreClient core.CoreServiceC
 				return NewValidationError("entity not found: " + entity.Name)
 			}
 
-			if len(entity.JsonPath) == 0 {
-				return NewValidationError(fmt.Sprintf("json path for %s is not specified", entity.Name))
+			switch entity.Extractor.(type) {
```
pradithya, line 55: Can you add validation logic for the udf config provided by user?

khorshuheng:
> Please add some test cases on `Test_buildEntitiesRequest` as well: https://github.com/gojek/merlin/blob/main/api/pkg/transformer/feast/transformer_test.go#L483

The test case in the original commit is for Test_buildEntitiesRequest. I added new test case to TestTransformer_Transform instead.

api/pkg/transformer/feast/validator.go
khorshuheng: Is it suffice to verify if the string is non empty, or do we need to make sure that it can be compiled without error as well? If the latter, right now initializing new transformer would have thrown error if the udf expression can't be compiled, so not sure whether we need an additional verification.

api/pkg/transformer/feast/udf.go
khorshuheng: Yes, it's definitely better to cache compiled json path. But implementation wise, that's a little tricky: the json path for latitude and longitude are part of the UDF expression. If we want to stored the compiled json path during transformer initialization, we would have to somehow extract the arguments from the expression, then store it in a map. It is also hard to generalize to all udfs, and the udf would have different arguments.

khorshuheng: Do we know how much latency is introduced by compiling the json path on the fly? If it is very significant, i will try to figure out some ways to cache the json path, but probably as a separate pull request?

pradithya: I am not too sure the additional processing time, but you can lazily create cache in this method itself, thus, only the first call to this method will call `Compile`. You'll need to ensure it's concurrency-safe.

api/pkg/transformer/feast/validator.go
pradithya: It's better to call `Compile` too. These validation steps are performed when user configure the standard transformer from the UI. If we chose the later, the error only show up as deployment failure and might confuse user. I think we forgot to add the validation for `jsonPath` too, so it will be great if you can add it here.

pradithya (approved):
Apart from caching json path in udf, everything else is LGTM

PR #82, opened 2021-02-25
Add labels to model Versions in Create Version endpoint

ashwinath:
Versions numbers do not tell us exactly what the model does. Most data scientists seem to keep these version numbers in a separate sheet/paper in order to tell which model does what. This PR adds in the ability to add labels to each model version via the CreateVersion endpoint so that data scientist will be able to add in metadata about each version.

This is a new feature

```release-note
Ability to add labels to create version endpoint.
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


davidheryanto:
@ashwinath Wonder why the CI job (GitHub Actions) do not run?

Oh and is it possible maybe to put those generated code changes in a different pull request? Unless it is necessary to pass the integration or end-to-end test (or it is tighly related to the current change), so that the changes are easier to review.

db-migrations/23_version_labels.down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE versions DROP COLUMN labels;
+
+DROP INDEX version_labels_idx;
```
davidheryanto, line 3: I think we need to change this order. Drop index should be performed before dropping the column because dropping the labels column will usually delete the related index, so the drop index command will fail.

davidheryanto: You can dry run the schema rollback locally like so:
> Using migrate command from https://github.com/golang-migrate/migrate
```
migrate -source file://./db-migrations -database postgres://USER:PASSWORD@localhost:5432/DATABASE_NAME?sslmode=disable down 1
```

api/api/versions_api.go
```diff
@@ -100,17 +99,22 @@ func (c *VersionsController) ListVersions(r *http.Request, vars map[string]strin
 	return OkWithHeaders(versions, responseHeaders)
 }
 
-func (c *VersionsController) CreateVersion(r *http.Request, vars map[string]string, _ interface{}) *Response {
+func (c *VersionsController) CreateVersion(r *http.Request, vars map[string]string, body interface{}) *Response {
 	ctx := r.Context()
 
+	versionPost, ok := body.(*models.VersionPost)
+	if !ok {
+		return InternalServerError("Unable to parse request body")
```
davidheryanto, line 107: Do you think BadRequest is a more appropriate HTTP status here since it is error due to failure in parsing request payload?

db-migrations/23_version_labels.down.sql
ashwinath: will edit this.

api/cmd/main.go
```diff
@@ -194,6 +195,7 @@ func main() {
 		DB:                        db,
 		Enforcer:                  authEnforcer,
 		FeastCoreClient:           coreClient,
+		MlFlowClientFactory:       mlflowClientFactory,
```
davidheryanto, line 198: hmm i think `ClientFactory` is not really necessary here? Do you think it's cleaner if this is just `MLflowClient` field which accepts an object that satisfies `mlflow.Client` interface?

ashwinath:
@davidheryanto yes the E2E test is important for the actual feature developed. for the Github CI maybe @ariefrahmansyah will know something about this?

You can omit this commit: 12595b4ff1942b6fc3eae30716b631edc9bcd1e1 when reviewing, that's all autogenerated code.

api/api/versions_api.go
ashwinath: agreed. this should be a 4xx rather than 5xx error

api/cmd/main.go
ashwinath: See line 213, they call it MlpURL so its kinda inconsistent. But if you follow the pattern it seems that outside names seem to be initial caps and small characters for the rest of the acronym.

ashwinath: now i know what you mean, let me change it.

api/api/versions_api_test.go
```diff
@@ -511,3 +513,137 @@ func TestPatchVersion(t *testing.T) {
 		})
 	}
 }
+
+type mockClient struct{}
```
davidheryanto, line 517: With the current project structure, I think it's cleaner if this is defined in a different package, e.g `mlflow.MockClient`, `mocks.MLflowClient` or smth similar.

Currently it's defined in `api` package and `api.mockClient` can be quite misleading when this is actually a fake / mock **MLflow** client. If we want to keep it here, at least we need to maybe prefix the name like `mockMLflowClient`

api/cmd/main.go
ashwinath: I'm not sure how to get away with not using a factory pattern in order to keep recreating a new client. I need to be able to mock the mlflow client in order to write unit tests.

We cant instantiate similar to how the core feast client does it as we are unable to keep the same state, i.e. mlflowTrackingUrl always changing per request. A singleton pattern would not suffice and a prototype pattern is required for this.

api/api/versions_api_test.go
ashwinath: alright.

ashwinath:
Closing this in favour of using this PR: https://github.com/gojek/merlin/pull/84

PR #84, opened 2021-02-25
Add labels to model Versions in Create Version endpoint

ashwinath:
Versions numbers do not tell us exactly what the model does. Most data scientists seem to keep these version numbers in a separate sheet/paper in order to tell which model does what. This PR adds in the ability to add labels to each model version via the CreateVersion endpoint so that data scientist will be able to add in metadata about each version.

This is a new feature

```release-note
Ability to add labels to create version endpoint.
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


api/api/versions_api.go
```diff
@@ -100,17 +99,22 @@ func (c *VersionsController) ListVersions(r *http.Request, vars map[string]strin
...
+
 	modelID, _ := models.ParseID(vars["model_id"])
 
 	model, err := c.ModelsService.FindByID(ctx, modelID)
 	if err != nil {
 		return NotFound(fmt.Sprintf("Model with given `model_id: %d` not found", modelID))
 	}
 
-	mlflowClient := mlflow.NewClient(nil, model.Project.MlflowTrackingUrl)
+	mlflowClient := c.MlflowClientFactory.NewClient(nil, model.Project.MlflowTrackingUrl)
```
pradithya, line 117: I think it's better to make `VersionsController` hold reference to `mlflowClient` directly. 
To give you some context, this code
```
mlflowClient := mlflow.NewClient(nil, model.Project.MlflowTrackingUrl)
```
was intended to for multiple MLFlow deployments. I don't think it's the case and we always have only 1 MLflow per merlin instance. Thus, it's better to instantiate the mlflow client during start up and store the reference in `VersionsController`

api/client/model_alert_condition_metric_type.go
```diff
@@ -13,9 +13,9 @@ type AlertConditionMetricType string
 
 // List of AlertConditionMetricType
 const (
-	ThroughputAlertConditionMetricType AlertConditionMetricType = "throughput"
-	LatencyAlertConditionMetricType    AlertConditionMetricType = "latency"
-	ErrorRateAlertConditionMetricType  AlertConditionMetricType = "error_rate"
-	CPUAlertConditionMetricType        AlertConditionMetricType = "cpu"
-	MemoryAlertConditionMetricType     AlertConditionMetricType = "memory"
+	THROUGHPUT_AlertConditionMetricType AlertConditionMetricType = "throughput"
```
pradithya, line 16: Can you exclude modification that changes the constant name? The swagger generated constant names are strangely not following convention.

api/cmd/main.go
```diff
@@ -176,9 +177,9 @@ func main() {
 	}
 	tracker.Start()
 
+	mlflowClientFactory := mlflow.NewClientFactory()
```
pradithya, line 180: Refer to this https://github.com/gojek/merlin/pull/84/files#r582762090 ,
It's better to instantiate mlflow client directly here.

api/api/versions_api.go
```diff
@@ -120,6 +124,7 @@ func (c *VersionsController) CreateVersion(r *http.Request, vars map[string]stri
 		ModelID:     modelID,
 		RunID:       run.Info.RunID,
 		ArtifactURI: run.Info.ArtifactURI,
+		Labels:      versionPost.Labels,
 	}
 
 	version, _ = c.VersionsService.Save(ctx, version, c.MonitoringConfig)
```
pradithya, line 130: Can you add test for `Save` operation

pradithya: The test should use a real database instance. You can refer to https://github.com/gojek/merlin/tree/main/api/storage for reference

api/api/versions_api.go
ashwinath: Do we want to break this compatibility? Because right now its reading the value from databases.

api/client/model_alert_condition_metric_type.go
ashwinath: Sorry to trouble you but how do I go about doing this?

api/api/versions_api.go
ashwinath: done

api/api/versions_api.go
ashwinath: Also, are we pulling this out into a configuration file?

api/client/model_alert_condition_metric_type.go
pradithya: Just don't submit the changes. @ariefrahmansyah Do you remember which swagger version that produce the correct convention?

ariefrahmansyah: Swagger produces constants with this format. We have CamelCase previously because we manually fix the linting error: https://github.com/gojek/merlin/commit/dc6f4c3bc3730c67cccdb7e0b106014abf90fd48#diff-bcbc448bedb0c6b16d1c8635ecb2fcd9567044f36d628d6a695ad0477f413361

I think what we can do is:
1. Do not commit these API's client change (We can cherry-pick the change)
2. Entirely ignore API's client directory from linting

I prefer to choose option number 1. I think we still need to check and lint the API's client to make sure the generated code is correct and expected.

api/cmd/main.go
ashwinath: i have addressed this

api/client/model_alert_condition_metric_type.go
ashwinath: alright i have changed this.

pradithya (approved):
LGTM

pradithya:
Oh I just noticed the e2e-test failed. Can you have a look? This step is currently just try to deploy merlin so probably the error is due to some configuration not correctly setup

PR #85, opened 2021-02-26
Add support to run GitHub Action for PR raised from fork repositories

jials:
Add support to run GitHub Actions for PRs raised from fork repositories.

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


jials:
Resolved merged conflict

pradithya (approved):
LGTM, I'll wait for @ariefrahmansyah

scripts/e2e/deploy-merlin.sh
```diff
@@ -6,12 +6,12 @@ CHART_PATH="$1"
 
 helm install merlin ${CHART_PATH} --namespace=mlp \
   --values=${CHART_PATH}/values-e2e.yaml \
-  --set merlin.image.tag=${GITHUB_REF#refs/*/} \
+  --set merlin.image.tag=${GITHUB_HEAD_REF} \
```
ariefrahmansyah, line 9: From the documentation [1](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables), GITHUB_HEAD_REF is "Only set for pull request events. The name of the head branch."
So what is the value of GITHUB_HEAD_REF once the GitHub Actions triggered after merged to master? Is it empty? If it's empty, I think we should fallback to using `${GITHUB_REF#refs/*/}`

.github/workflows/ci.yml
```diff
@@ -93,28 +143,6 @@ jobs:
...
-      - name: Build and publish
-        env:
-          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
-          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-        working-directory: ./python/sdk
-        run: |
-          python setup.py sdist bdist_wheel
-          twine upload dist/*
-
   publish-merlin-docker:
```
ariefrahmansyah, line 146: Why is there no `if` condition here while other publish jobs have? If this jobs is triggered and run from a fork, to which GHCR account the docker will be published? GitHub or accounts who forked the repo?

scripts/e2e/deploy-merlin.sh
jials: Currently we are only running this deployment shell script when during the `push_request` event as shown [here](https://github.com/gojek/merlin/pull/85/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR2) so this won't be a problem.

ariefrahmansyah: This script is part of [e2e-test job](https://github.com/gojek/merlin/blob/f268d8cf75ac04970441d86f0b95accf49727b4c/.github/workflows/ci.yml#L286), which doesn't have pull_request and pull_request_target validation. And also, we should still running this e2e test job on the main branch.

.github/workflows/ci.yml
jials: Ah yes, good catch. The `if` condition should be here as well. To answer your question, the same `${{ secrets.GHCR_TOKEN }}` will be used to push the docker from fork repository based on the `if` condition.

scripts/e2e/deploy-merlin.sh
jials: So what you are saying is that you want these GH actions to support `push` as well right? https://github.com/gojek/merlin/blob/f268d8cf75ac04970441d86f0b95accf49727b4c/.github/workflows/ci.yml#L2

ariefrahmansyah: yes, we can do it that way. or we can also have separate actions spec (or we can update `publish-python-sdk.yml`) that consists of jobs that need to run on push.

jials: Hmm should we allow direct push to the branch though? I think all changes should be from PR (support for `push` event is not needed here). I think to support the scenario you mentioned we need to do something like manual triggers with workflow_dispatch. wdyt?

ariefrahmansyah: > should we allow direct push to the branch though?
no. push to `main` branch is not allowed, we need PR for that. but we should be able to push another branch to repository. and when I push my changes to my branch, I expect the github actions to run without needing me to make a PR. it offers more flexibility to the developer.

> I think all changes should be from PR (support for push event is not needed here)
yes, all changes should be from PR but the push trigger is still needed. let's say 2 devs working on 2 features and both of them created a separate PR. each PR will have 2 different e2e test scenarios. but once it's merged to the main branch, there's no e2e-test for both scenarios. that's why we still need e2e-test to run on push trigger.

workflow_dispatch will be great but once we have `push` trigger, I don't think we need it.

pradithya:
@ariefrahmansyah @jials Is it mergable now?

ariefrahmansyah:
> @ariefrahmansyah @jials Is it mergable now?

I think @jials still working on ${GITHUB_HEAD_REF} fallback value in deploy-merlin.sh

ariefrahmansyah (approved):
LGTM! Thanks @jials

PR #94, opened 2021-03-03
Support UDF syntax in Entities UI

jials:
**What this PR does / why we need it**:
Add a dropdown column to allow user to specify field type to be UDF or JSONPath.

UDF
image

image

JSONPath
image

image

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

- [ ] Added unit test, integration, and/or e2e tests
- [X] Tested locally
- [ ] Updated documentation
- [ ] Update Swagger spec if the PR introduce API changes
- [ ] Regenerated Golang and Python client if the PR introduce API changes

PR #95, opened 2021-03-04
Fix UI bug where existing env var is not read properly for standard transformer

jials:
Fix UI bug where existing env var is not read properly for standard transformer

Fixes #94

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


ariefrahmansyah (approved):
LGTM

PR #99, opened 2021-03-08
[WIP]: Enable external host from merlin-chart

imjuanleonard:

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/charts/merlin/templates/_helpers.tpl
+++ b/charts/merlin/templates/_helpers.tpl
@@ -20,7 +20,11 @@
 {{- end -}}
 
 {{- define "postgresql.host" -}}
+{{- if .Values.postgresql.enabled -}}
 {{- printf "%s-postgresql.%s.svc.cluster.local" .Release.Name .Release.Namespace -}}
+{{- else -}}
+{{- printf "%s" .merlin.database.host -}}
+{{- end -}}
 {{- end -}}
 
 {{- define "mlflow.backendStoreUri" -}}
--- a/charts/merlin/values.yaml
+++ b/charts/merlin/values.yaml
@@ -135,6 +135,9 @@ merlin:
     # This should be the actual DNS registered
     trackingURL: "http://www.example.com"
 
+  database:
+    host: "localhost"
+
 mlflow:
   image:
     registry: ghcr.io


charts/merlin/values.yaml
```diff
@@ -135,6 +135,11 @@ merlin:
     # This should be the actual DNS registered
     trackingURL: "http://www.example.com"
 
+  database:
```
pradithya, line 138: Can you comment these out and add description such as what is it and when is it being used.

charts/merlin/values-e2e.yaml
```diff
@@ -66,7 +66,13 @@ merlin:
     # This should be the actual DNS registered
     trackingURL: "http://www.example.com"
 
+  database:
```
pradithya, line 69: Same as https://github.com/gojek/merlin/pull/99/files#r590982104

charts/merlin/templates/deployment.yaml
```diff
@@ -79,17 +79,26 @@ spec:
...
+            value: "{{ .Values.postgresql.postgresqlDatabase }}"
+          {{- else -}}
+          - name: DATABASE_HOST
+            value: "{{ .Values.merlin.database.host }}"
+          - name: DATABASE_USER
+            value: "{{ .Values.merlin.database.user }}"
+          - name: DATABASE_NAME
+            value: "{{ .Values.merlin.database.name }}"
+          {{- end }}
           - name: DATABASE_PASSWORD
```
pradithya, line 97: How are you planning to configure the `DATABASE_PASSWORD` ?
It seems to me this part is still reusing the built-in postgres.

pradithya: I prefer to user to specify the source entirely:
```
            valueFrom:
              secretKeyRef:
                name: {{ .Values.merlin.database.password.secretName }}
                key:  {{ .Values.merlin.database.password.secretKey }}
```
Can you check with any open source helm chart whether this is best practise.

charts/merlin/values.yaml
```diff
@@ -147,6 +152,7 @@ mlflow:
     postgresqlUsername: mlflow
 
 postgresql:
+  enabled: false
```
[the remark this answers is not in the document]
imjuanleonard: Done

charts/merlin/values.yaml
imjuanleonard: Done

charts/merlin/values-e2e.yaml
imjuanleonard: Done

charts/merlin/templates/deployment.yaml
```diff
@@ -79,17 +79,26 @@ spec:
             value: "{{ .Values.merlin.loggerDestinationURL }}"
           - name: PORT
             value: "{{ .Values.merlin.service.internalPort }}"
+          {{- if .Values.postgresql.enabled }}
```
ariefrahmansyah, line 82: instead of having conditions here, i prefer to use _helpers.tpl

in _helpers.tpl, I will have this:

```
{{- define "database.host" -}}
{{- if .Values.postgresql.enabled -}}
{{- printf "%s-postgresql.%s.svc.cluster.local" .Release.Name .Release.Namespace -}}
{{- else -}}
{{- .Values.merlin.database.host -}}
{{- end -}}
{{- end -}}

and 3 more for db name, user, and password
```

then in deployment.yaml, there will be no logic, but simply:

```
- name: DATABASE_HOST
   value: {{ template "database.host" . }}
```

imjuanleonard: At first, I plan to do this however after implementing it, the helper's class become bloated.
Meanwhile, I have seen some logic in the templates like the sentry enable, etc. It is not hard to read.

WDYT?

I don't mind to revert it back to helpers template though

ariefrahmansyah: Sentry's case is different from database config. For sentry, we simply check if we are going to use it or not. But for database, we need to decide which value to use for a given postgresql provider.

_helpers.tpl supposed to be "bloated" because it's abstract the complexity of other template files.

imjuanleonard: Ah understood, I saw similar pattern in postgresql as well mas.
Thank you for the review already refactor it

charts/merlin/templates/deployment.yaml
imjuanleonard: I tried to look at various open source, but not found any resemblance except for feast.

For Feast, there is an option to use an existing secret (which deployed internally and separated resources with the helm chart)

I have refactored the database password, rather than using postgres one, have added your code snippets and create a secret resources should postgres is not enabled

charts/merlin/templates/_helpers.tpl
```diff
@@ -19,10 +19,40 @@
 {{- end -}}
 {{- end -}}
 
-{{- define "postgresql.host" -}}
+{{- define "database.host" -}}
+{{- if .Values.postgresql.enabled }}
 {{- printf "%s-postgresql.%s.svc.cluster.local" .Release.Name .Release.Namespace -}}
+{{- else -}}
+
```
ariefrahmansyah, line 26: Why is this empty?

charts/merlin/values-e2e.yaml
```diff
@@ -66,7 +66,20 @@ merlin:
     # This should be the actual DNS registered
     trackingURL: "http://www.example.com"
 
+  # Database connection used by Merlin backend when postgres is disabled
+  #
+  # database:
```
ariefrahmansyah, line 71: You can merge these two example into one.

```
# Specify database configuration
# database:
#   # Database host
#   database: "localhost"
...
```

charts/merlin/values.yaml
```diff
@@ -135,6 +135,26 @@ merlin:
     # This should be the actual DNS registered
     trackingURL: "http://www.example.com"
 
+  # Database connection used by Merlin backend when postgres is disabled
+  #
+  # database:
```
ariefrahmansyah, line 140: same as above

ariefrahmansyah (changes requested):

PR #103, opened 2021-03-17
Change postgres helm Chart to allow external database

imjuanleonard:
- Revert the changes of the previous helm chart release
- Change the Postgres helm chart to add external postgres type (Cloudsql) and enable to configure postgres host database


charts/merlin/requirements.yaml
```diff
@@ -2,7 +2,6 @@ dependencies:
   - name: postgresql
     repository: "@stable"
     version: 7.0.0
-    condition: postgresql.enabled
```
pradithya, line 5: If this is remove does it mean it will always deploy a prostgres instance?

charts/merlin/values.yaml
```diff
@@ -157,13 +147,16 @@ mlflow:
...
 
 postgresql:
-  enabled: true
   resources:
     requests:
       memory: 2Gi
       cpu: 500m
   persistence:
     size: 100Gi
+  ## Option to use postgresql "internal" or "external" database type, by default internal is chosen
```
pradithya, line 156: I thought we agree to have "cloudsql" and "postgresql" as separate type? Why it's becoming "internal" and "external"

charts/merlin/values.yaml
```diff
@@ -157,13 +147,16 @@ mlflow:
...
+  type: internal
+  ## To use an external PostgreSQL instance uncomment the line below:
+  ## postgresqlServer: ""
```
pradithya, line 159: Use `postgresqlHost`

charts/merlin/values.yaml
pradithya: I think it's still better to use `enabled` : true/false, rather than `internal` vs `external`

charts/merlin/templates/database-password-secret.yaml
```diff
@@ -1,23 +0,0 @@
-{{- if not .Values.postgresql.enabled }}
-apiVersion: v1
-kind: Secret
```
pradithya, line 3: How is database password being stored when it's using "external"?

imjuanleonard: It will be created using the same way as the previous one. using the `postgresqlPassword` parameter

charts/merlin/values.yaml
imjuanleonard: Done

charts/merlin/requirements.yaml
imjuanleonard: Yes it is, have re-added this back

charts/merlin/templates/_helpers.tpl
```diff
@@ -19,40 +19,14 @@
 {{- end -}}
 {{- end -}}
 
-{{- define "database.host" -}}
-{{- if .Values.postgresql.enabled }}
+{{- define "postgresql.host" -}}
+{{- if empty .Values.postgresql.postgresqlHost -}}
```
pradithya, line 23: Why not using `{{- if .Values.postgresql.enabled }}` ?

imjuanleonard: to allow backward compatibility with current merlin deployment. 
In the current merlin deployment, the MLFlow DB is disabled, therefore if it is checking based on postgresql.enabled, this value won't be found out.

Rather than complicating the if in the helpers, checking whether postgresqlHost set up or not would be a cleaner way to use internal or ezternal.

pradithya (approved):
LGTM, let's make the E2E test pass

PR #104, opened 2021-03-18
Fix standard transformer validation

ariefrahmansyah:

We fix standard transformer validation function to support user to use non-default project. Previously, The ListEntities call didn't specify the project name, so by default, Feast will return the "default" project. See https://github.com/feast-dev/feast/blob/v0.9.2/sdk/go/protos/feast/core/CoreService.pb.go#L1489-L1492

We also have fix for E2E tests:
1. Fix CLI tests to use dynamic environment target value
2. Fix standard transformer tests to use Merlin Test's entities and features

Fixes #

```release-note
NONE
```

**Checklist**

- [x] Updated unit test, integration, and/or e2e tests
- [x] Tested locally


python/sdk/test/cli_integration_test.py
```diff
@@ -31,9 +31,10 @@ def deployment_info():
     filename = os.path.join(dirname, 'test/sklearn-model')
     url = os.environ.get("E2E_MERLIN_URL", default="http://127.0.0.1:8080")
     project = os.environ.get("E2E_PROJECT_NAME", default="integration-test")
+    env = os.environ.get("E2E_MERlIN_ENVIRONMENT", default="id-dev")
```
pradithya, line 34: It's a typo `E2E_MERLIN_ENVIRONMENT`

ariefrahmansyah: Nice catch! Updated

pradithya (approved):
LGTM

PR #107, opened 2021-03-24
Queue deployment

tiopramayudi:

Add queue mechanism when deploying inferenceservice & batch job. This is to solve issue when merlin server is restarted by external factor and the ongoing deployment(inferenceservice or batch) will be failed and status is not updated

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


.env.sample
```diff
@@ -54,3 +54,4 @@ MONITORING_DASHBOARD_JOB_BASE_URL=http://monitoring.dev/batch-job-dashboard
 MLP_API_ENCRYPTION_KEY=password
 
 # MLFLOW_TRACKING_URL=
+NUM_OF_WORKERS=1
```
pradithya, line 57: What's the ideal number of worker?

api/cluster/controller.go
```diff
@@ -156,6 +158,13 @@ func (k *controller) Deploy(modelService *models.Service) (*models.Service, erro
 
 	s, err = k.waitInferenceServiceReady(s)
 	if err != nil {
+		// remove newly created inferenceservice when got error
```
pradithya, line 161: This is unrelated to queue implementation right? it's for the fix when deployment is considered fail but we didn;t garbage collect the resource.

api/cmd/ui_handler.go
```diff
@@ -0,0 +1,77 @@
+package main
```
pradithya, line 1: Does it make sense to have separate folder for cmd.go, setup.go, and ui_handler.go ? I.e. under cmd/api/

api/queue/dispatcher.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/gojek/merlin/log"
+	"github.com/jinzhu/gorm"
+)
+
+type Producer interface {
+	EnqueueJob(job *Job) error
+}
+
+type Consumer interface {
```
pradithya, line 18: Can you add description to these interfaces : Producer and Consumer

api/queue/dispatcher.go
```diff
@@ -0,0 +1,131 @@
...
+type JobFn func(*Job) error
+
+type Config struct {
+	NumWorkers int
+	Db         *gorm.DB
+}
+
+type Status string
+
+type Argument map[string]interface{}
```
pradithya, line 41: Maybe rename it to `Arguments` since it contains more than one Argument

api/queue/dispatcher.go
```diff
@@ -0,0 +1,131 @@
...
+type Job struct {
+	ID        int64     `json:"id" gorm:"primary_key;"`
+	Arguments Argument  `json:"arguments"`
+	Completed bool      `json:"completed"`
+	Name      string    `json:"name"`
+	CreatedAt time.Time `json:"created_at"`
+	UpdatedAt time.Time `json:"updated_at"`
+}
+
+func (a Argument) Value() (driver.Value, error) {
```
pradithya, line 52: I think `Arguments` definition should be separated into its own file.

api/queue/dispatcher.go
```diff
@@ -0,0 +1,131 @@
...
+		return errors.New("type assertion to []byte failed")
+	}
+
+	return json.Unmarshal(b, &a)
+}
+
+func NewDispatcher(cfg Config) *Dispatcher {
+	workers := make([]*worker, 0, cfg.NumWorkers)
+	jobChan := make(chan *Job)
+	jobFuncMap := make(map[string]JobFn)
```
pradithya, line 68: I think jobFuncMap should use the thread-safe version of map instead of stdlib map to avoid concurrency issue

api/queue/dispatcher.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func (d *Dispatcher) EnqueueJob(job *Job) error {
+	var savedJob Job
+	if err := d.db.Save(job).Scan(&savedJob).Error; err != nil {
+		log.Errorf("Failed to save job %d with error: %v", job.ID, err)
+		return err
+	}
+	go func() {
```
pradithya, line 87: What is this code for?

api/queue/worker.go
```diff
@@ -0,0 +1,84 @@
...
+				go w.processJob(job)
+			}
+		}
+	}()
+}
+
+func (w *worker) processJob(job *Job) {
+	var refreshedJob Job
+	tx := w.db.Begin()
+	err := tx.Raw(findJobQuery, job.ID, false).Scan(&refreshedJob).Error
```
pradithya, line 47: Is it possible to not use `SKIP_LOCKED` in the query? The reason is that `worker` become unable to identify whether the row already has a lock or really doesn't exists.

api/service/model_endpoint_service.go
```diff
@@ -35,6 +35,9 @@ import (
 )
 
 const (
+	RealtimeDeploymentJob = "realtime"
```
pradithya, line 38: I think `model service deployment` is a better name

api/service/prediction_job_service.go
```diff
@@ -42,6 +43,8 @@ type PredictionJobService interface {
 	ListContainers(env *models.Environment, model *models.Model, version *models.Version, predictionJob *models.PredictionJob) ([]*models.Container, error)
 	// StopPredictionJob deletes the spark application resource and cleans up the resource
 	StopPredictionJob(env *models.Environment, model *models.Model, version *models.Version, id models.ID) (*models.PredictionJob, error)
+	// ExecuteDeployment creates prediction job based on queue job that produce by CreatePredictionJob method
```
pradithya, line 46: I think `ExecuteDeployment` shouldn't be part of `PredictionJobService`.  
`PredictionJobService` should just submit the request to queue, and then queue+worker handle the rest.

api/service/version_endpoint_service.go
```diff
@@ -60,24 +75,121 @@ type endpointService struct {
...
+
+type EndpointJobArgs struct {
+	Endpoint *models.VersionEndpoint
+	Model    *models.Model
+	Version  *models.Version
+	Project  mlp.Project
+}
+
+// ExecuteDeployment will create inferenceservice based on job that already queued
+func (k *endpointService) ExecuteDeployment(job *queue.Job) error {
```
pradithya, line 102: Similar to comment https://github.com/gojek/merlin/pull/107/files#r601994686 . I think `ExecuteDeployment` shouldn't be part of the service. It's better to split it as separate class and closely related to `queue` package.

db-migrations/1_schema_creation.up.sql
```diff
@@ -16,7 +16,7 @@ CREATE TYPE endpoint_status as ENUM ('pending', 'running', 'serving', 'failed',
 
 CREATE TABLE IF NOT EXISTS projects
 (
-    id                  serial PRIMARY KEY,
+    id                  bigserial PRIMARY KEY,
```
pradithya, line 19: Will it become an issue?

api/cluster/controller.go
tiopramayudi: yes, it is not related to queue implementation. I changed this also because it is related to model deployment

api/cmd/ui_handler.go
tiopramayudi: yes make sense

db-migrations/1_schema_creation.up.sql
tiopramayudi: that is not supposed to be here? will revert this

api/queue/dispatcher.go
tiopramayudi: so for enqueuing job, queue will insert the job record to DB if it success it will send the job into channel, since the worker will consume job from values from channel

api/queue/worker.go
tiopramayudi: there will be no case the row will be empty because the record is not there in DB because all job already persist to DB before. I think case that I havent covered is when query return err because of issue with DB. I think for this case we can requeue the job after delay for several seconds

api/service/version_endpoint_service.go
tiopramayudi: package queue will handle registration of job and consume the job, but the logic to consume the job itself is handled in relevant package. IMO version_endpoint_service or prediction_job_service still relevant to deploy the job, or do we need another file for deployment only let say deployment_service.go?

api/service/model_endpoint_service.go
tiopramayudi: ok I'll change this

api/queue/dispatcher.go
tiopramayudi: will change this

api/queue/worker.go
pradithya: Then what is this code for?
```
	if err == gorm.ErrRecordNotFound {
		tx.Rollback()
		log.Warnf("Job records not found with id: %d", job.ID)
		return
	}
```

api/service/version_endpoint_service.go
pradithya: Yeah, better to separate these into its own package. probably under queue/work. We can name it deployment.go and batch_prediction.go

api/queue/worker.go
tiopramayudi: to return early, this indicate that job already success or locked in other process. The logging message should be improve, I'll change this

api/service/version_endpoint_service.go
```diff
@@ -60,26 +73,123 @@ type endpointService struct {
...
 }
 
+type EndpointJobArgs struct {
+	Endpoint *models.VersionEndpoint
+	Model    *models.Model
+	Version  *models.Version
+	Project  mlp.Project
+}
+
+// ExecuteDeployment will create inferenceservice based on job that already queued
```
pradithya, line 99: Dead code

api/work/batch_deployment.go
```diff
@@ -0,0 +1,85 @@
+package work
```
pradithya, line 1: Can it be part of /queue/work package ?

api/models/batch_job.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+import "github.com/gojek/merlin/mlp"
+
+type BatchJob struct {
```
pradithya, line 5: I think we shouldn't put `BatchJob` and `EndpointJob` under `models` package. It's better to put it closer to the `work` package

api/models/endpoint_job.go
```diff
@@ -0,0 +1,12 @@
+package models
+
+import (
+	"github.com/gojek/merlin/mlp"
+)
+
+type EndpointJob struct {
```
pradithya, line 7: https://github.com/gojek/merlin/pull/107/files#r603758002

pradithya (approved):
LGTM

PR #110, opened 2021-03-30
Support authentication when calling feast

tiopramayudi:

Feast already support authentication, hence we need to update our feast core client to pass required credentials. Also we need to update `Authorization-Header` when calling feast from UI

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/cmd/api/setup.go
```diff
@@ -83,8 +84,13 @@ func initMLPAPIClient(ctx context.Context, cfg config.MlpAPIConfig) mlp.APIClien
...
 
-func initFeastCoreClient(feastCoreURL string) core.CoreServiceClient {
-	cc, err := grpc.Dial(feastCoreURL, grpc.WithInsecure())
+func initFeastCoreClient(feastCoreURL, feastAuthAudience string) core.CoreServiceClient {
+	cred, err := feast.NewGoogleCredential(feastAuthAudience)
+	if err != nil {
+		panic(err)
+	}
+
+	cc, err := grpc.Dial(feastCoreURL, grpc.WithInsecure(), grpc.WithPerRPCCredentials(cred))
```
pradithya, line 93: You have to choose either `grpc.WithInsecure()` or `grpc.WithPerRPCCredentials(cred)` isn't it?
And we should add enabled/disabled flag

api/cmd/transformer/main.go
```diff
@@ -70,7 +70,14 @@ func main() {
 	if err != nil {
 		logger.Fatal("Unable to parse Feast Serving Port", zap.String("feast-serving-port", feastPort), zap.Error(err))
 	}
-	feastClient, err := feastSdk.NewGrpcClient(feastHost, feastPortInt)
+
+	cred, err := feastSdk.NewGoogleCredential(cfg.Feast.AuthAudience)
```
pradithya, line 74: How is it different from `initFeastCoreClient` ?

pradithya, line 74: We don't need auth for feast serving

charts/merlin/templates/merlin-deployment.yaml
```diff
@@ -143,6 +143,10 @@ spec:
             value: "{{ .Values.merlin.transformer.feast.servingURL }}"
           - name: FEAST_CORE_URL
             value: "{{ .Values.merlin.transformer.feast.coreURL }}"
+          - name: FEAST_CORE_AUTH_AUDIENCE
+            value: "{{ .Values.merlin.transformer.feast.coreAuthAudience }}"
+          - name: FEAST_SERVING_AUTH_AUDIENCE
```
pradithya, line 148: We don't need auth for serving

api/cmd/api/setup.go
tiopramayudi: need to add `gcpc.WIthInsecure()` because the feast credentials is not transport credentials.

pradithya: Got it

PR #117, opened 2021-04-15
Fix: Navbar docs link

ariefrahmansyah:

The docs links should use Documentations struct instead of string. Previously, I assumed we only use JSON string for passing docs link but it should be properly parsed as Documentations struct.


tiopramayudi:
LGTM

PR #119, opened 2021-04-23
Change how to build Feast request

ariefrahmansyah:

The previous implementation of building the Feast request is resulting to huge number of entity due to combination logic used.

In this commit, we change it with the assumption of entities pairing instead of combination.

See unit test for example.

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/transformer/feast/transformer.go
+++ b/api/pkg/transformer/feast/transformer.go
@@ -375,12 +375,17 @@ func (t *Transformer) buildEntitiesRequest(ctx context.Context, request []byte, 
 	}
 
 	entities := make([]feast.Row, maxLength)
+	for k := range entities {
+		entities[k] = feast.Row{}
+	}
+
 	for s, series := range allSeries {
 		entityName := configEntities[s].Name
-
 		if len(series) == 1 {
 			entities = t.broadcastSeries(entityName, series, entities)
-		} else {
+		}
+
+		if len(series) > 1 {
 			entities = t.addSeries(entityName, series, entities)
 		}
 	}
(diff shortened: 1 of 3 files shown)


api/pkg/transformer/feast/transformer.go
```diff
@@ -248,12 +248,13 @@ func (t *Transformer) getFeatureFromFeast(ctx context.Context, entities []feast.
 		}
 		batchedEntities := entityNotInCache[startIndex:endIndex]
 
-		go func(project string, entityList []feast.Row, columns []string) {
+		go func(project string, entityList []feast.Row, columns []string, i int) {
```
pradithya, line 251: It seems the variable `i` is not used?

ariefrahmansyah: You're right. It was added for debugging purposes. Removed.

api/pkg/transformer/feast/transformer.go
```diff
@@ -337,62 +339,79 @@ func getExpressionExtractor(entity *transformer.Entity) string {
...
 		}
 	}
 
 	return entities, nil
 }
 
+func (t *Transformer) broadcastSeries(entityName string, series []*types.Value, entities []feast.Row) []feast.Row {
+	for e := range entities {
+		newFeastRow := feast.Row{}
+		for k, v := range entities[e] {
```
pradithya, line 394: I think this part will make the whole `buildEntitiesRequest` slower as it will be 3 nested for loop. In worst case it will be O(m*n^2), m: number of series, n: number of entities, it might not be good when there are a lot of entities.

pradithya: This can be simply
```
	for _, entity := range entities {
		entity[entityName] = series[0]
	}
	return entities
```

api/pkg/transformer/feast/transformer.go
```diff
@@ -337,62 +339,79 @@ func getExpressionExtractor(entity *transformer.Entity) string {
...
+			newFeastRow[k] = v
+		}
+		newFeastRow[entityName] = series[0]
+		entities[e] = newFeastRow
+	}
+	return entities
+}
+
+func (t *Transformer) addSeries(entityName string, series []*types.Value, entities []feast.Row) []feast.Row {
```
pradithya, line 403: Same as https://github.com/gojek/merlin/pull/119/files#r618989906, it can be simplified to
```
	for idx, entity := range entities {
		entity[entityName] = series[idx]
	}
	return entities
```

pradithya (approved):
LGTM

PR #121, opened 2021-04-28
Standard transformer transformation

pradithya:

This PR add ability to specify transformation:
1. table transformation (select, drop, rename, update columns, and sort by columns)
2. table join

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/pkg/transformer/pipeline/compiler.go
```diff
@@ -198,10 +214,32 @@ func (c *Compiler) parseTablesSpec(tableSpecs []*spec.Table, compiledJsonPaths *
...
+	for _, step := range transformationSpecs.Steps {
+		for _, updateColumn := range step.UpdateColumns {
+			compiledExpression, err := c.compileExpression(updateColumn.Expression)
+			if err != nil {
+				return nil, err
+			}
+			compiledExpressions.Set(updateColumn.Expression, compiledExpression)
+		}
+	}
+
```
tiopramayudi, line 227: i think you need to register dummy variables for outputTable

tiopramayudi: also can you add test case when using array of transformation when the second transformation using output of previous transformation?

api/pkg/transformer/pipeline/compiler.go
```diff
@@ -198,10 +214,32 @@ func (c *Compiler) parseTablesSpec(tableSpecs []*spec.Table, compiledJsonPaths *
 	return NewCreateTableOp(tableSpecs), nil
 }
 
+func (c *Compiler) parseTableTransform(transformationSpecs *spec.TableTransformation, paths *jsonpath.Storage, compiledExpressions *expression.Storage) (Op, error) {
+	for _, step := range transformationSpecs.Steps {
```
tiopramayudi, line 218: why you don't check expressions for the inputTable?

api/pkg/transformer/pipeline/compiler.go
```diff
@@ -198,10 +214,32 @@ func (c *Compiler) parseTablesSpec(tableSpecs []*spec.Table, compiledJsonPaths *
...
+			}
+			compiledExpressions.Set(updateColumn.Expression, compiledExpression)
+		}
+	}
+
+	return NewTableTransformOp(transformationSpecs), nil
+}
+
+func (c *Compiler) parseTableJoin(tableJoinSpecs *spec.TableJoin, paths *jsonpath.Storage, expressions *expression.Storage) (Op, error) {
+	return NewTableJoinOp(tableJoinSpecs), nil
```
tiopramayudi, line 232: why you don't check expressions for `leftTable`, `rightTable`?

tiopramayudi: also I think you need to register outputTable variables / dummy table, in case it is referred in the next step of operation

api/pkg/transformer/pipeline/table_join_op.go
```diff
@@ -0,0 +1,117 @@
...
+	rightTable, err := getTable(environment, t.tableJoinSpec.RightTable)
+	if err != nil {
+		return err
+	}
+
+	var resultTable *table.Table
+	joinColumn := t.tableJoinSpec.On
+	switch t.tableJoinSpec.How {
+	case spec.JoinMethod_LEFT:
+		err := validateJoinColumn(leftTable, rightTable, joinColumn)
```
tiopramayudi, line 36: why do we need this? if we call do `LeftJoin` it will also failed right if there is no overlapping column?

pradithya: It's to get a better error message, the error message thrown by gota is not user-friendly.

api/pkg/transformer/pipeline/compiler.go
pradithya: Added the check

api/pkg/transformer/pipeline/compiler.go
pradithya: Updated

api/pkg/transformer/pipeline/compiler.go
pradithya: Updated

tiopramayudi:
LGTM

PR #124, opened 2021-05-04
Standard Transformer Output Stage

tiopramayudi:

This PR add json output capability for standard transformer

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/pkg/transformer/pipeline/compiler.go
```diff
@@ -198,6 +205,77 @@ func (c *Compiler) parseTablesSpec(tableSpecs []*spec.Table, compiledJsonPaths *
...
+				return errors.New("can't specify nested json, if field has value to set")
+			}
+
+			// Check whether format specified is valid or not
+			if val.FromTable.Format == spec.FromTable_INVALID {
+				return errors.New("table format specified is invalid")
+			}
+
+			// Check whether table name already registered or not
+			tableName := val.FromTable.TableName
```
pradithya, line 253: I think we can just simply check whether the variable is declared and doesn't need to treat it as expression

PR #134, opened 2021-05-11
Transformer: Update Content-Length header

ariefrahmansyah:

On writing HTTP response header, we are reusing (copying) the request's Content-Length header that might be different with actual response. This could raise EOF response (e.g.: `curl: (18) transfer closed with XX bytes remaining to`) from the standard transformer when there's preprocess / postprocess output modification due to a mismatch on Content-Length header. This PR sets the Content-Length to the correct request or response body size.

**Checklist**

- [x] Added tests
- [x] Tested locally


tiopramayudi:
is it possible to create test for this kind of case?

ariefrahmansyah:
> is it possible to create test for this kind of case?

I've updated the tests that check Content-Length header.

pradithya (approved):
LGTM

pradithya:
I think the PR description requires update

PR #136, opened 2021-05-19
Fix Error in SDK When Trying to Fetch VersionEndpoint

tiopramayudi:

With new release of merlin v0.12.0 , there is slight change payload logger that specified by ui. For when disabled the logger, the logger mode specified is empty string, which is invalid and sdk didn't handle it properly. So we create changes in API and SDK. 
API Changes:
* Return logger object with default logger mode if logger mode specified in body is invalid

SDK Changes:
* Fallback to default logger mode if logger mode from API response is invalid

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/models/logger.go
+++ b/api/models/logger.go
@@ -56,11 +56,10 @@ type LoggerConfig struct {
 	Mode    LoggerMode `json:"mode"`
 }
 
-func (lc *LoggerConfig) ToValidConfig() *LoggerConfig {
+func (lc *LoggerConfig) SanitizeMode() {
 	if !lc.Mode.isValid() {
 		lc.Mode = LogAll
 	}
-	return lc
 }
 
 func ToKFServingLoggerMode(mode LoggerMode) kfsv1alpha2.LoggerMode {
--- a/api/models/logger.go
+++ b/api/models/logger.go
@@ -36,6 +36,15 @@ const (
 	LogResponse LoggerMode = "response"
 )
 
+func (l LoggerMode) isValid() bool {
+	switch l {
+	case LogAll, LogRequest, LogResponse:
+		return true
+	default:
+		return false
+	}
+}
+
 type Logger struct {
 	DestinationURL string        `json:"-"`
 	Model          *LoggerConfig `json:"model"`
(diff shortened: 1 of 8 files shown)


api/models/logger.go
```diff
@@ -47,6 +56,13 @@ type LoggerConfig struct {
 	Mode    LoggerMode `json:"mode"`
 }
 
+func (lc *LoggerConfig) ToValidConfig() *LoggerConfig {
+	if !lc.Mode.isValid() {
+		lc.Mode = LogAll
```
ariefrahmansyah, line 61: Should we add `lc.Enabled = false` too?

ariefrahmansyah: I understand this function incorrectly. I think it's better to rename this function to `SetModeToDefault()` as it's more explanatory.

python/sdk/merlin/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-VERSION = "0.12.0"
+VERSION = "0.12.1"
```
ariefrahmansyah, line 15: You also need to update this line: https://github.com/gojek/merlin/blob/main/python/pyfunc-server/requirements.txt#L9

api/models/logger.go
tiopramayudi: it is not necessary set mode to default, it changed the mode to default one if the model is not valid. Does `SetModeToDefaultIfNecessary()` make sense?

ariefrahmansyah: I think it's too long 😂 
I suggest `ToValidMode` or `SanitizeMode`.

Oh ya, any reason why is the function returning the *LoggerConfig? The receiver already a pointer so the method already changes the receiver value.

tiopramayudi: no specific reason, just habit not create function that only mutate its value, I can make it returning nothing

python/sdk/merlin/version.py
tiopramayudi: since the version that used by pyfunc-server always the latest, I will remove specific version on the requirements.txt 
so the requirements look like this `merlin-sdk` only, wdyt

ariefrahmansyah: Agree with that. I like that approach!

ariefrahmansyah (approved):
LGTM

PR #139, opened 2021-05-27
Set capacity of map when creating json with records format

tiopramayudi:

When the table that needs to be converted is huge, the latency in conversion is high. By add capacity when initialize map, it will reduce the latency and memory allocation

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


pradithya (approved):
LGTM!

PR #140, opened 2021-06-01
UI Improvement: Transformer Inputs

ariefrahmansyah:

1. On specifying table inputs, users can specify columns without the need to specify the base table.

**Before**

image

**After**

image

2. On specifying feast inputs, users can group feast tables so the retrieval process started asynchronously.

**Before**

image

image

**After**

image

image

**Checklist**

- [x] Tested locally

### DIFF
--- a/ui/src/pages/version/components/forms/components/feast_config/components/FeastInputGroup.js
+++ b/ui/src/pages/version/components/forms/components/feast_config/components/FeastInputGroup.js
@@ -1,0 +1,114 @@
+import React from "react";
+import {
+  EuiDragDropContext,
+  euiDragDropReorder,
+  EuiDraggable,
+  EuiDroppable,
+  EuiFlexGroup,
+  EuiFlexItem,
+  EuiPanel,
+  EuiSpacer
+} from "@elastic/eui";
+import { get, useOnChangeHandler } from "@gojek/mlp-ui";
+import { DraggableHeader } from "../../DraggableHeader";
+import { FeastResourcesContextProvider } from "../../../../../../../providers/feast/FeastResourcesContext";
+import { AddButton } from "../../transformer/components/AddButton";
+import { FeastInput } from "../../../../../../../services/transformer/TransformerConfig";
+import { FeastInputCard } from "./FeastInputCard";
+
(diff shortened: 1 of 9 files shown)


pradithya:
So if there are 2 or more feast input that's defined subsequent to each other, these will be flattened into 1 `feast` input?
But will not be flattened if it is separated by other input?

pradithya:
> So if there are 2 or more feast input that's defined subsequent to each other, these will be flattened into 1 `feast` input?
> But will not be flattened if it is separated by other input?

I think the screenshot in description is misleading

ariefrahmansyah:
> So if there are 2 or more feast input that's defined subsequent to each other, these will be flattened into 1 `feast` input?
> But will not be flattened if it is separated by other input?

No. It's not automatically flattened. The user must manually click `+ Add Feast Table` to have another Feat retrieval on the same group.

image

pradithya:
> > So if there are 2 or more feast input that's defined subsequent to each other, these will be flattened into 1 `feast` input?
> > But will not be flattened if it is separated by other input?
> 
> No. It's not automatically flattened. The user must manually click `+ Add Feast Table` to have another Feat retrieval on the same group.
> 
> image

Yes, the button is not visible in previous screenshot. It looks good to me

pradithya (approved):
LGTM!

ariefrahmansyah:
I've added more screenshots on the description.

ui/src/pages/version/components/forms/components/feast_config/components/FeastInputGroup.js
```diff
@@ -0,0 +1,114 @@
...
+          </EuiDragDropContext>
+        </EuiFlexGroup>
+
+        <EuiFlexItem>
+          <AddButton
+            title="+ Add Feast Table"
+            description="Add another Feast table that will be created asynchronously with other tables in this group."
+            titleSize="xs"
+            onClick={() => onAddTable()}
+          />
```
tiopramayudi, line 100: Please remove this if not being used

tiopramayudi:
left one comment, the rest is LGTM

PR #141, opened 2021-06-02
Capture termination log as error message during pyfunc image build failure

khorshuheng:
Currently, user has no visibility on what went wrong when deploying model endpoint. Ideally, the SDK itself should give information on what went wrong with the deployment.

One of the possible error that a user might encounter is failure to build the pyfunc image. To capture the error, we can use the logs prior to pod termination as termination message, and propagate the information back to the users. 

Fixes #

```release-note
Model endpoint deployment will provide more descriptive log in the event of pyfunc image build failure.
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/db-migrations/25_message_varchar_length.down.sql
+++ b/db-migrations/25_message_varchar_length.down.sql
@@ -1,2 +1,1 @@
-ALTER TABLE version_endpoints DROP COLUMN message;
-ALTER TABLE version_endpoints ADD COLUMN message varchar(64);
+ALTER TABLE version_endpoints ALTER COLUMN message varchar(64) USING SUBSTR(message, 1, 64);
--- a/db-migrations/25_message_varchar_length.down.sql
+++ b/db-migrations/25_message_varchar_length.down.sql
@@ -1,0 +1,2 @@
+ALTER TABLE version_endpoints DROP COLUMN message;
+ALTER TABLE version_endpoints ADD COLUMN message varchar(64);
--- a/python/sdk/merlin/model.py
+++ b/python/sdk/merlin/model.py
@@ -59,6 +59,26 @@ V1 = "v1"
 PREDICTION_JOB = "PredictionJob"
 
 
+class ModelEndpointDeploymentError(Exception):
+
+    def __init__(self, model_name: str, version: int, details: str):
+        self._model_name = model_name
+        self._version = version
+        self._details = details
+
+    @property
+    def model_name(self):
+        return self._model_name
+
+    @property
+    def version(self):
+        return self._version
+
+    @property
+    def details(self):
+        return self._details
+
+
 @autostr
 class Project:
     def __init__(self, project: client.Project, mlp_url: str,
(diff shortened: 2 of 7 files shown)


python/sdk/merlin/model.py
```diff
@@ -1004,6 +1004,8 @@ def deploy(self, environment_name: str = None,
         bar.stop()
 
         if endpoint.status != "running":
+            print("Model deployment failed. Reason:")
```
pradithya, line 1007: I think it's better to include the error message in the `ValueError` instead of printing it.

ariefrahmansyah (approved):
Please bump the SDK version to `0.12.2` here: https://github.com/gojek/merlin/blob/main/python/sdk/merlin/version.py#L15

Other than that, LGTM!

python/sdk/merlin/model.py
khorshuheng: In fact i am not too sure personally if we really should throw the entire log to the user, though it's certainly better than the current status quo which the user must go to the console and find out what goes wrong via logging.

khorshuheng: One way would be truncating the log output before returning the error, for example. But then it becomes the problem on how do we truncate it, whether by fixed number of line or trying to infer which lines correspond to traceback / errors

khorshuheng: Alternatively, instead of a ValueError, we can have a dedicated error type, with the message as an attribute, and the caller can decide what to do with the log message.

pradithya: I see. Btw, I remember `message` was intended for a short description and the schema is limited https://github.com/gojek/merlin/blob/main/db-migrations/3_deployment_status.up.sql

pradithya: In the UI side, there is a use of `message` field here https://github.com/gojek/merlin/blob/5e08c02a9acecad0570f0a62f2011ce9a1d2803b/ui/src/version/VersionListTable.js#L217
Since this will be a long message and it's shown in a small table. I think we can avoid showing this for now by removing these part
```
                        {versionEndpoint.status === "failed" ? (
                          <EuiText size={defaultTextSize}>
                            <EuiIcon
                              type={"alert"}
                              size={defaultIconSize}
                              color="danger"
                            />
                            {versionEndpoint.message}
                          </EuiText>
                        )
```

pradithya: As for the DB schema you can add migration script to extend the column capacity

db-migrations/25_message_varchar_length.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE version_endpoints DROP COLUMN message;
```
pradithya, line 1: This might cause data loss, preferably we could just trim the content, e.g:

```
ALTER TABLE version_endpoints ALTER COLUMN message varchar(64)
USING SUBSTR(message, 1, 64)
```

pradithya:
LGTM! Thanks @khorshuheng !

PR #143, opened 2021-06-07
Handle output json when using expression which the value is series or…

tiopramayudi:

This PR will solve unexpected JSON output when one of the json field expect value from table or series. Before this PR, the field value would be `nil`

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


pradithya (approved):
LGTM!

PR #150, opened 2021-06-22
Deduplicate Entities Rows Before Sending to Feast Server

tiopramayudi:

Dedup entity rows before sending to feast. The purpose is to remove duplicate result in feast response and reduce number of unnecessary request to feast

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/transformer/feast/feature_retriever.go
+++ b/api/pkg/transformer/feast/feature_retriever.go
@@ -24,8 +24,6 @@ import (
 )
 
 const (
-	hystrixCommandName = "feast_retrieval"
-
 	maxUint = ^uint(0)
 	maxInt  = int(maxUint >> 1)
 )
@@ -56,7 +54,7 @@ func NewFeastRetriever(
 
 	defaultValues := compileDefaultValues(featureTableSpecs)
 
-	hystrix.ConfigureCommand(hystrixCommandName, hystrix.CommandConfig{
+	hystrix.ConfigureCommand(options.FeastClientHystrixCommandName, hystrix.CommandConfig{
 		Timeout:                durationToInt(options.FeastTimeout, time.Millisecond),
 		MaxConcurrentRequests:  options.FeastClientMaxConcurrentRequests,
 		RequestVolumeThreshold: options.FeastClientRequestVolumeThreshold,
@@ -85,6 +83,7 @@ type Options struct {
 	CacheTTL                time.Duration `envconfig:"FEAST_CACHE_TTL" default:"60s"`
 
 	FeastTimeout                      time.Duration `envconfig:"FEAST_TIMEOUT" default:"1s"`
+	FeastClientHystrixCommandName     string        `envconfig:"FEAST_HYSTRIX_COMMAND_NAME" default:"feast_retrieval"`
 	FeastClientMaxConcurrentRequests  int           `envconfig:"FEAST_HYSTRIX_MAX_CONCURRENT_REQUESTS" default:"100"`
 	FeastClientRequestVolumeThreshold int           `envconfig:"FEAST_HYSTRIX_REQUEST_VOLUME_THRESHOLD" default:"100"`
 	FeastClientSleepWindow            int           `envconfig:"FEAST_HYSTRIX_SLEEP_WINDOW" default:"1000"` // How long, in milliseconds, to wait after a circuit opens before testing for recovery
@@ -435,7 +434,7 @@ func (fr *FeastRetriever) getFeatureTable(ctx context.Context, entities []feast.
 		batchedEntities := entityNotInCache[startIndex:endIndex]
 
 		f := fr.newFeastCall(featureTableSpec, batchedEntities)
-		hystrix.GoC(ctx, hystrixCommandName, func(ctx context.Context) error {
+		hystrix.GoC(ctx, fr.options.FeastClientHystrixCommandName, func(ctx context.Context) error {
 			batchResultChan <- f.do(ctx, features, columns, entityIndices)
 			return nil
 		}, func(ctx context.Context, err error) error {
--- a/api/pkg/transformer/feast/feature_retriever.go
+++ b/api/pkg/transformer/feast/feature_retriever.go
@@ -3,6 +3,7 @@ package feast
 import (
 	"context"
 	"encoding/base64"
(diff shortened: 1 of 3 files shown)


ariefrahmansyah (approved):

api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -3,13 +3,15 @@ package feast
...
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
 	"strings"
 	"time"
 
 	"github.com/afex/hystrix-go/hystrix"
+	"github.com/cespare/xxhash"
```
pradithya, line 14: Is there any specific reason for using this library? I thought built in library such as https://golang.org/pkg/crypto/sha256/ should work too

api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -213,7 +214,29 @@ func (fr *FeastRetriever) buildEntityRows(ctx context.Context, symbolRegistry sy
...
-	return entities, nil
+	uniqueEntities, err := dedupEntities(entities)
+	if err != nil {
+		return nil, err
+	}
+
+	return uniqueEntities, nil
+}
+
+func dedupEntities(rows []feast.Row) ([]feast.Row, error) {
```
pradithya, line 225: Can you benchmark the code to compare before and after adding dedupEntities?

api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -213,7 +214,29 @@ func (fr *FeastRetriever) buildEntityRows(ctx context.Context, symbolRegistry sy
...
+	uniqueRows := make([]feast.Row, 0, len(rows))
+	rowLookup := make(map[uint64]bool)
```
pradithya, line 227: Is it possible to do:
```
rowLookup := make(map[feast.Row]bool)
```

pradithya: not possible since apparently `feast.Row` is a map

api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -213,7 +214,29 @@ func (fr *FeastRetriever) buildEntityRows(ctx context.Context, symbolRegistry sy
...
+	for _, row := range rows {
+		rowByte, err := json.Marshal(row)
```
pradithya, line 229: `json.Marshal` might be too expensive for this.
Alternatively we can "stringify" the row. I.e. create a string by appending all the keys and values from the Row.
The string then can be used as key of `rowLookup`
```
rowLookup := make(map[string]bool)
rowLookup[stringify(row)] = true
```

api/pkg/transformer/feast/feature_retriever.go
tiopramayudi: from their benchmark, it said that it is faster then standard go, also it is the one that used by our in memory cache library

api/pkg/transformer/feast/feature_retriever.go
tiopramayudi: there is no guarantee that keys in maps is sorted, so we need to sort the keys first. Let me check the performance of it

tiopramayudi: I add this 
`func dedupEntities_stringify(rows []feast.Row) ([]feast.Row, error) {
	uniqueRows := make([]feast.Row, 0, len(rows))
	rowLookup := make(map[string]bool)
	keys := make([]string, 0)
	if len(rows) == 0 {
		return rows, nil
	}
	for key := range rows[0] {
		keys = append(keys, key)
	}
	sort.Strings(keys)
	for _, row := range rows {
		var keyString strings.Builder
		for _, k := range keys {
			keyString.WriteString(fmt.Sprintf("%s:%v_", k, row[k]))
		}
		rowKey := keyString.String()
		if _, found := rowLookup[rowKey]; !found {
			uniqueRows = append(uniqueRows, row)
			rowLookup[rowKey] = true
		}
	}
	return uniqueRows, nil
}`

`func Benchmark_dedup_stringify(b *testing.B) {
	b.StopTimer()

	entities := make([]feast.Row, 0, 100)
	for i := 0; i < 100; i++ {
		row := feast.Row{}
		row["merchant_id"] = feast.Int32Val(int32(i))
		// row["driver_id"] = feast.Int32Val(int32((i + 1) * 2))
		entities = append(entities, row)
	}
	b.StartTimer()
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		request, _ := dedupEntities_stringify(entities)
		_ = request
	}
}`

benchmark result:
`goos: darwin
goarch: amd64
pkg: github.com/gojek/merlin/pkg/transformer/feast
cpu: Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
Benchmark_dedup_stringify-8   	    8710	    135942 ns/op	   31236 B/op	     812 allocs/op`

tiopramayudi: also compare with existing dedup method

```go
func Benchmark_dedup(b *testing.B) {
	b.StopTimer()

	entities := make([]feast.Row, 0, 100)
	for i := 0; i < 100; i++ {
		row := feast.Row{}
		row["merchant_id"] = feast.Int32Val(int32(i))
		// row["driver_id"] = feast.Int32Val(int32((i + 1) * 2))
		entities = append(entities, row)
	}
	b.StartTimer()
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		request, _ := dedupEntities(entities)
		_ = request
	}
}
```

goos: darwin
goarch: amd64
pkg: github.com/gojek/merlin/pkg/transformer/feast
cpu: Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
Benchmark_dedup-8   	   13713	     87502 ns/op	   29856 B/op	     618 allocs/op

pradithya: Got it, I am wondering if there is a way to convert `feast.Row` into `byte[]` which can be more efficient than using json.

pradithya (approved):
LGTM

PR #153, opened 2021-06-23
Fix spec feast validator for case of using expression in entity extraction

pradithya:

Fix standard transformer spec validation for case when expression is used within feast entity extraction. E.g.
```
transformerConfig:
  preprocess:
    inputs:
      - variables:
          - name: customer_level
            literal:
              intValue: 10
      - feast:
          - tableName: customer_feature_table
            project: default
            entities:
              - name: customer_level
                valueType: INT32
                expression: customer_level
            features:
              - name: customer_feature_1
                valueType: INT64
                defaultValue: '0'
...
```
Currently, the above spec will cause validation error mentioning that `customer_level` expression is not defined even though it's declared in the variable block.
The issue is due to validator not reusing symbol registry which should contains the declared variables.

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/transformer/pipeline/compiler_test.go
+++ b/api/pkg/transformer/pipeline/compiler_test.go
@@ -311,6 +311,7 @@ func TestCompiler_Compile(t *testing.T) {
 				expressions: []string{
 					"customer_level",
 					"customer_id",
+					"Now().Hour()",
 				},
 				jsonPaths: []string{
 					"$.customer.id",
--- a/api/pkg/transformer/pipeline/compiler_test.go
+++ b/api/pkg/transformer/pipeline/compiler_test.go
@@ -259,6 +259,77 @@ func TestCompiler_Compile(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			name: "feast preprocess",
+			fields: fields{
+				sr:          symbol.NewRegistry(),
+				feastClient: &mocks.Client{},
+				feastOptions: &feast.Options{
+					CacheEnabled: true,
+				},
+				cacheOptions: &cache.Options{
+					SizeInMB: 100,
+				},
+				logger: logger,
+			},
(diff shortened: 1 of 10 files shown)


api/pkg/transformer/pipeline/compiler_test.go
```diff
@@ -258,6 +258,77 @@ func TestCompiler_Compile(t *testing.T) {
...
+					CacheEnabled: true,
+				},
+				cacheOptions: &cache.Options{
+					SizeInMB: 100,
+				},
+				logger: logger,
+			},
+			specYamlFilePath: "./testdata/valid_feast_expression.yaml",
+			want: want{
+				expressions: []string{
```
ariefrahmansyah, line 311: can you add test for UDF expresion, such as Now(), or Geohash()?

api/pkg/transformer/pipeline/testdata/valid_feast_expression.yaml
```diff
@@ -0,0 +1,79 @@
...
+                defaultValue: '0'
+          - tableName: customer_feature_table
+            project: default
+            entities:
+              - name: customer_id
+                valueType: STRING
+                expression: customer_id
+              - name: customer_level
+                valueType: INT32
+                expression: customer_level
```
ariefrahmansyah, line 38: Can you also add UDF expression here? Such as Now() or Geohash()

pradithya: Added

api/pkg/transformer/pipeline/compiler_test.go
pradithya: Added

ariefrahmansyah (approved):
lgtm

PR #157, opened 2021-07-06
Add config for jaeger tracing and add more tracing span

tiopramayudi:

Pass jaeger configuration to standard transformer spec so jaeger can be enabled. Adding more tracing in standard transformer in:
* Create table op
* Create variable op
* Table transformation op
* Table join op
* Json output op

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/config/config.go
```diff
@@ -162,6 +162,15 @@ type StandardTransformerConfig struct {
...
 	EnableAuth            bool   `envconfig:"FEAST_AUTH_ENABLED" default:"false"`
+	Jaeger                JaegerConfig
+}
+
+type JaegerConfig struct {
+	AgentHost    string `envconfig:"JAEGER_AGENT_HOST"`
+	AgentPort    string `envconfig:"JAEGER_AGENT_PORT"`
+	SamplerType  string `envconfig:"JAEGER_SAMPLER_TYPE" default:"const"`
+	SamplerParam string `envconfig:"JAEGER_SAMPLER_PARAM" default:"1"`
+	Disabled     string `envconfig:"JAEGER_DISABLED" default:"false"`
```
ariefrahmansyah, line 173: If the tracing enabled by default, we should change the jaeger sampler to `probabilistic` with param value to be `0.1` (or less). Having `const` as jaeger sampler type would use a lot of space and also affect the transformer performance in production.

api/pkg/transformer/pipeline/create_table_op.go
```diff
@@ -17,7 +18,10 @@ func NewCreateTableOp(tableSpecs []*spec.Table) Op {
 	return &CreateTableOp{tableSpecs: tableSpecs}
 }
 
-func (c CreateTableOp) Execute(_ context.Context, env *Environment) error {
+func (c CreateTableOp) Execute(ctx context.Context, env *Environment) error {
+	span, _ := opentracing.StartSpanFromContext(ctx, "pipeline.CreateTableOp")
```
ariefrahmansyah, line 22: We should still use the returned / updated `ctx` here.

api/pkg/transformer/pipeline/table_transform_op.go
```diff
@@ -18,6 +19,9 @@ func NewTableTransformOp(tableTransformSpec *spec.TableTransformation) Op {
 }
 
 func (t TableTransformOp) Execute(context context.Context, env *Environment) error {
+	span, _ := opentracing.StartSpanFromContext(context, "pipeline.TableTransformOp")
+	defer span.Finish()
+
 	inputTableName := t.tableTransformSpec.InputTable
 	outputTableName := t.tableTransformSpec.OutputTable
```
ariefrahmansyah, line 26: We can add more data to span by using SetTag:

```
span.SetTag("table.input", inputTableName)
span.SetTag("table.output", outputTableName)
```

Having this can help us identify the operation on an individual level, especially when there are multiple similar operations.

api/pkg/transformer/pipeline/table_join_op.go
```diff
@@ -19,6 +20,9 @@ func NewTableJoinOp(tableJoinSpec *spec.TableJoin) Op {
 }
 
 func (t TableJoinOp) Execute(context context.Context, environment *Environment) error {
+	span, _ := opentracing.StartSpanFromContext(context, "pipeline.TableJoin")
```
ariefrahmansyah, line 23: We can add more data to span by using SetTag:

```
span.SetTag("table.left", t.tableJoinSpec.LeftTable)
span.SetTag("table.right", t.tableJoinSpec.RightTable)
span.SetTag("table.output", t.tableJoinSpec.OutputTable)
```

Having this can help us identify the operation on an individual level, especially when there are multiple similar operations.

api/pkg/transformer/pipeline/feast_op.go
```diff
@@ -34,6 +35,9 @@ func NewFeastOp(feastClient feastSdk.Client, feastOptions *feast.Options, cache
 }
 
 func (op *FeastOp) Execute(context context.Context, env *Environment) error {
+	span, context := opentracing.StartSpanFromContext(context, "pipeline.FeastOp")
```
ariefrahmansyah, line 38: nit: rename context to ctx

ariefrahmansyah (changes requested):

api/pkg/transformer/pipeline/create_table_op.go
tiopramayudi: why? it is not passed to other function

ariefrahmansyah: ah, it's okay to omit it then.

ariefrahmansyah (approved):

api/config/config.go
pradithya: I prefer to have 0.01 as default.

PR #161, opened 2021-07-21
Adding toogle to use heimdall hystrix integration

tiopramayudi:

We face issue with additional latency by using heimdall library (hystrix integration), after did benchmark without using heimdall the latency is back to normal. Hence we add toggle to use heimdall or not. The downside by disable heimdall is there won't be any retry capability

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/hystrix/client.go
+++ b/api/pkg/hystrix/client.go
@@ -1,7 +1,7 @@
 package hystrix
 
 import (
-	"errors"
+	"fmt"
 	"net/http"
 
 	"github.com/afex/hystrix-go/hystrix"
@@ -39,7 +39,7 @@ func (cl *Client) Do(request *http.Request) (*http.Response, error) {
 			return err
 		}
 		if response.StatusCode >= http.StatusInternalServerError {
-			return errors.New("got 5xx response code")
+			return fmt.Errorf("got 5xx response code: %d", response.StatusCode)
 		}
 		return nil
 	}, nil)
--- a/api/pkg/hystrix/client.go
+++ b/api/pkg/hystrix/client.go
@@ -1,0 +1,47 @@
+package hystrix
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/afex/hystrix-go/hystrix"
+)
+
+type Doer interface {
+	Do(request *http.Request) (*http.Response, error)
+}
+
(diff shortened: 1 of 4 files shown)


api/pkg/transformer/server/server.go
```diff
@@ -67,25 +69,49 @@ type Server struct {
...
+		SleepWindow:            o.ModelHystrixSleepWindowMs,
+		ErrorPercentThreshold:  o.ModelHystrixErrorPercentageThreshold,
+	}
+	cl := &http.Client{
+		Timeout: o.ModelTimeout,
+	}
+	return hystrixpkg.NewClient(cl, &hystrixConfig, hystrixCommandName)
+}
+
+func newHeimdallHystrixClient(commandName string, o *Options) *hystrix.Client {
```
pradithya, line 114: It's better to name it `newHeimdallClient`

api/pkg/hystrix/client.go
```diff
@@ -0,0 +1,47 @@
...
+	var response *http.Response
+	var err error
+
+	err = hystrix.Do(cl.commandName, func() error {
+		response, err = cl.client.Do(request)
+		if err != nil {
+			return err
+		}
+		if response.StatusCode >= http.StatusInternalServerError {
+			return errors.New("got 5xx response code")
```
pradithya, line 42: Can you add response code the the error message

pradithya: Also is there a way if we can also read the response payload and add error logging here.

api/pkg/transformer/server/server.go
```diff
@@ -67,25 +69,49 @@ type Server struct {
 	PostprocessHandler func(ctx context.Context, response []byte, responseHeaders map[string]string) ([]byte, error)
 }
 
+type doer interface {
```
pradithya, line 72: Maybe call it `httpClient`

api/pkg/hystrix/client.go
tiopramayudi: I think for response payload can be read by the caller instead of in the client itself

pradithya (approved):

ariefrahmansyah (approved):

PR #163, opened 2021-07-23
Limit Message Saved for VersionEndpoints to 2048

tiopramayudi:

Previously we faced issue where the deployment stuck in `pending` state where in theory `pending` state only transient status. We found the root cause is, the deployment failed when building image, the error of this failure is really long that exceed the maximum allowed in version_endpoints table. Hence save the version endpoints was failing, to mitigate this we trim the message with maximum length 2048 according to max char allowed in db

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/storage/version_endpoint_storage.go
```diff
@@ -50,9 +52,18 @@ func (v *versionEndpointStorage) Get(uuid uuid.UUID) (*models.VersionEndpoint, e
 }
 
 func (v *versionEndpointStorage) Save(endpoint *models.VersionEndpoint) error {
+	checkDBContraint(endpoint)
 	return v.db.Save(&endpoint).Error
 }
 
+func checkDBContraint(endpoint *models.VersionEndpoint) {
```
ariefrahmansyah, line 59: nit: since this function not only check the endpoint but also alter the endpoint, I prefer to rename it to something like `sanitizeEndpoint`

ariefrahmansyah (approved):
LGTM with a nit comment.

PR #164, opened 2021-07-26
Improve Model Endpoint and Batch Job's Logging UX

ariefrahmansyah:

This PR improves the Logging UX by aggregating logs from several pods and containers to one stream. It allows the user to stream the component's logs (model, transformer, batch job driver/executor, image builder) instead of one container log at a time.

In this PR we also fixes following issues:
- Fixes Log UI component scrolling
- Fixes Strackdriver URL

```release-note
- Update Logging UI
```

**Screenshots**

1. Model and transformer
image

2. Pyfunc and it's image builder
image

3. Batch job
image

4. No model
image

5. No batch job
image

**Checklist**

- [x] Tested locally
- [ ] ~Regenerated Golang and Python client if the PR introduce API changes~ -- To be raised in separate PR.


api/api/log_api.go
```diff
@@ -42,38 +42,34 @@ func (l *LogController) ReadLog(w http.ResponseWriter, r *http.Request) {
...
-	buff := bufio.NewReader(res)
-	for {
-		line, readErr := buff.ReadString('\n')
-		_, writeErr := w.Write([]byte(line))
-		if writeErr != nil {
-			// connection from caller is closed
-			return
-		}
+	go func() {
+		for podLog := range podLogs {
```
pradithya, line 53: I think it's cleaner if logservice already merge the pod logs into a single stream.

api/service/log_service.go
```diff
@@ -15,51 +15,239 @@
...
+							TextPayload:   logLine[timestampIndex+1:],
+						}
+
+						podLogs = append(podLogs, log)
+					}
+
+					allPodLogs = append(allPodLogs, podLogs...)
+				}
+
+				for _, container := range pod.Spec.Containers {
```
tiopramayudi, line 149: the logic below is the same with the above, can you extract that logic into a function

api/service/log_service.go
```diff
@@ -15,51 +15,239 @@
...
+						// Error is handled here by logging it rather than returned because the caller usually does not know how to
+						// handle it. Example of what can trigger ListPodLogs error: while the container is being created/terminated
+						// Kubernetes API server will return error when logs are requested. In such case, it is better to return
+						// empty logs and let the caller retry after the container becomes ready eventually.
+						log.Warnf("Failed to ListPodLogs: %s", err.Error())
+						return
+					}
+
+					scanner := bufio.NewScanner(stream)
+					podLogs := make([]*PodLog, 0)
```
tiopramayudi, line 107: any reason using array of pointer?

api/service/log_service.go
ariefrahmansyah: Yup. Done that.

api/service/log_service.go
ariefrahmansyah: No particular reason. I change it to a normal struct.

api/api/log_api.go
```diff
@@ -42,38 +48,33 @@ func (l *LogController) ReadLog(w http.ResponseWriter, r *http.Request) {
...
-	defer res.Close()
-	buff := bufio.NewReader(res)
-	for {
-		line, readErr := buff.ReadString('\n')
-		_, writeErr := w.Write([]byte(line))
-		if writeErr != nil {
-			// connection from caller is closed
-			return
-		}
+	go func() {
```
tiopramayudi, line 59: what happens if user open the log UI and after several seconds open another page? Should we need to listen to context change, and close all the necessary channel? So there won't be unfinished goroutine

api/service/log_service.go
```diff
@@ -15,53 +15,290 @@
...
+			text = append(text, p(l.PodName)+" "+p(l.ContainerName))
+		} else if options.Prefix == prefixAddPod {
+			text = append(text, p(l.PodName))
+		} else if options.Prefix == prefixAddContainer {
+			text = append(text, p(l.ContainerName))
+		}
+	}
+
+	text = append(text, l.Timestamp.Format(time.RFC3339), l.TextPayload)
+	return strings.Join(text, " ") + "\n"
```
tiopramayudi, line 81: I think you need to benchmark this function and for comparison maybe using bytes.Buffer to concatenate sting to reduce allocation. For some language time formatting is expensive, I'm not sure with golang better to check also

Makefile
```diff
@@ -24,7 +25,7 @@ init-dep: init-dep-ui init-dep-api
 .PHONY: init-dep-ui
 init-dep-ui:
 	@echo "> Initializing UI dependencies ..."
-	@cd ${UI_PATH} && yarn
+	@cd ${UI_PATH} && yarn install --network-concurrency 1
```
tiopramayudi, line 28: just curious why do we need this `--network-concurrency 1`

api/api/log_api.go
ariefrahmansyah: Currently, if the caller stop streaming the logs, we will close the stopCh and return as handled in this code: https://github.com/gojek/merlin/pull/164/files/55f9de8f7cd497d88114a3c4d5241178e4168917#diff-4d8a066a2fe8e1b33215ba5fa1142a25d9ea12586226a456f06d81b0b722dcebR64-R68

I don't think listening to context would help, but let me try it too.

Makefile
ariefrahmansyah: It is related to this issue: https://github.com/yarnpkg/yarn/issues/7212

We are installing react-lazylog from our forked git repository: https://github.com/gojekfarm/react-lazylog and in it's package.json, we have `prepare` script that could raise "data race" during dependencies installation.

api/api/log_api.go
ariefrahmansyah: You are correct. Listening to context solves the performance issue! See this commit: https://github.com/gojek/merlin/pull/164/commits/5d4991641bd0b264aa9d3dfe733cb1b019cb51f8

api/service/log_service.go
ariefrahmansyah: The difference between using strings.Join and bytes.Buffer is not negligible and the `generateText()`'s duration itself is insignificant compare to other functions (fetching the container logs). I'm going to keep it as is.

ui/src/utils/createStackdriverUrl.js
```diff
@@ -18,20 +18,19 @@ const querystring = require("querystring");
...
-resource.labels.cluster_name:${container.cluster}
-resource.labels.namespace_name:${container.namespace}
-resource.labels.pod_name:${container.pod_name}
-resource.labels.container_name:${container.name}`;
+const stackdriverFilter = query => {
+  return `resource.type:"k8s_query" OR "k8s_container" OR "k8s_pod"
+resource.labels.project_id:${query.gcp_project}
+resource.labels.cluster_name:${query.cluster}
+resource.labels.namespace_name:${query.namespace}
+resource.labels.pod_name:${query.pod_name}`;
```
pradithya, line 26: `resource.labels.pod_name:${query.pod_name}` --> this will restrict the query to only view one pod.  I think we should create query that is similar to the merlin logs (multiple pods)

ariefrahmansyah: The pod_name value is something like this: `pod_a OR pod_b OR pod_c`. It allows stackdriver to stream logs from multiple pods.
The UI generate this value from Get Containers API: https://github.com/gojek/merlin/blob/improve-logging-ux/ui/src/components/logs/ContainerLogsView.js#L132-L142

api/api/log_api.go
```diff
@@ -42,38 +48,36 @@ func (l *LogController) ReadLog(w http.ResponseWriter, r *http.Request) {
...
+	w.Header().Set("Content-Type", "text/event-stream")
+	w.Header().Set("Cache-Control", "no-cache")
+	w.Header().Set("Connection", "keep-alive")
+	w.Header().Set("Transfer-Encoding", "chunked")
+
+	go func() {
+		for {
+			logLine := <-logLineCh
+
+			fmt.Fprint(w, logLine)
```
pradithya, line 71: Better to also handle error when writing to the http response writer here.

api/api/log_api.go
```diff
@@ -42,38 +48,36 @@ func (l *LogController) ReadLog(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	res, err := l.LogService.ReadLog(&query)
-	if err != nil {
-		log.Errorf("Error while retrieving log %v", err)
-		InternalServerError(fmt.Sprintf("Error while retrieving log for container %s: %s", query.Name, err)).WriteTo(w)
-		return
-	}
+	logLineCh := make(chan string, 1000)
```
pradithya, line 51: Why `1000`?

api/cmd/api/main.go
```diff
@@ -105,6 +106,11 @@ func main() {
 	mount(router, "/v1", api.NewRouter(appCtx))
 	mount(router, "/metrics", promhttp.Handler())
 
+	// Debug routes
+	if cfg.EnablePprof {
```
pradithya, line 110: On a second through, let's not make it configurable and always enabled.

api/config/config.go
```diff
@@ -47,6 +47,8 @@ type Config struct {
 	UI                        UIConfig
 	StandardTransformerConfig StandardTransformerConfig
 	MlflowConfig              MlflowConfig
+
+	EnablePprof bool `envconfig:"ENABLE_PPROF" default:"false"`
```
pradithya, line 51: https://github.com/gojek/merlin/pull/164/files#r678884389 To reduce bloat in our configuration.

api/service/log_service.go
```diff
@@ -15,53 +15,290 @@
...
+	PodName string `json:"pod_name"`
+	// Container name that produces this log
+	ContainerName string `json:"container_name"`
+	// Log in text format
+	TextPayload string `json:"text_payload,omitempty"`
+
+	PrefixColor *color.Color `json:"-"`
+}
+
+func (l LogLine) GenerateText(options LogQuery) string {
```
pradithya, line 66: Can you add method description for `GenerateText`

api/service/log_service.go
```diff
@@ -15,53 +15,290 @@
...
+	namespace := options.Namespace
+	labelSelector := l.getLabelSelector(*options)
+
+	pods, err := clusterController.ListPods(namespace, labelSelector)
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
```
pradithya, line 122: I think it's more idiomatic in go to use ticker. E.g.

```
ticker := time.NewTicker(500 * time.Millisecond)
    go func() {
        for {
            select {
            case <-stopCh:
                return
            case t := <-ticker.C:
                doStuff
            }
        }
    }()
```

api/service/log_service.go
```diff
@@ -15,53 +15,290 @@
...
+				case <-stopCh:
+					return
+				case logLineCh <- logLine.GenerateText(*options):
+				}
+			}
+
+			now := time.Now()
+			options.SinceTime = &now
+
+			time.Sleep(5 * time.Second)
```
pradithya, line 180: Do we need to wait this long? It seems too long.

ariefrahmansyah: I'll reduce it to 1. Previously I increased it for debugging purposes.

ui/src/components/logs/ContainerLogsView.js
```diff
@@ -0,0 +1,211 @@
...
+
+  const [{ data: containers, isLoaded: containersLoaded }] = useMerlinApi(
+    fetchContainerURL,
+    { mock: mocks.containerOptions },
+    []
+  );
+
+  useEffect(
+    () => {
+      if (containersLoaded) {
```
pradithya, line 63: Can it be triggered periodically using timer? I think currently user need to refresh the page in order to update the list of containers. This will be a problem of discoverability.

ui/src/utils/createStackdriverUrl.js
pradithya: Ah ok got it.

ui/src/components/logs/ContainerLogsView.js
ariefrahmansyah: It's complex to achieve that. Updating the `containers` list periodically would eventually update the LazyLog's Stream URL. However, any change in LazyLog's Stream URL would re-render the whole component and re-stream the API instead of appending the new Stream API log. It would need a whole rework of our Logging logic.

pradithya: Why do we need to update the LazyLog's Stream URL? The intention is just to always provide most up to date list of list available pod. 

Screenshot 2021-07-29 at 3 42 48 PM

Otherwise, we can always shows all possible choice, but return no logs if the associated components doesn't have any container.

api/api/log_api.go
ariefrahmansyah: No longer using buffered channel here. Unbuffered channel is better suited for logging case so it doesn't need to wait until the limit of the buffer is reached to send the log.

api/service/log_service.go
ariefrahmansyah: Thanks for the suggestion. Updated!

pradithya: I think it can also further improved by breaking up the log collection from different pods.

```
ticker := time.NewTicker(1 * time.Second) 
go func() { 
for { 
   select { 
     case <-stopCh: return 
     case t := <-ticker.C: 
       allLogLines, err  :=  readPodLogs(pods) 
       	for _, logLine := range allLogLines {
		logLineCh <- logLine.generateText(*options)
	}
    } 
  } 
}()


```

api/service/log_service.go
```diff
@@ -15,53 +15,287 @@
...
+	KnativeServiceLabelKey     = "serving.knative.dev/service"
+	BatchPredictionJobLabelKey = "prediction-job-id"
+
+	prefixAddPod          = "pod"
+	prefixAddContainer    = "container"
+	prefixAddPodContainer = "pod_and_container"
+)
+
+var (
+	skippedContainers = []string{"queue-proxy"}
```
pradithya, line 47: rename to `ignoreContainers`

api/service/log_service.go
ariefrahmansyah: Done that. Please take a look again

pradithya: Great!

ui/src/components/logs/ContainerLogsView.js
ariefrahmansyah: Previously, I thought it would also update the request to log's stream URL because it was coupled. I have updated the UI logic to separately update the component list and stream the logs.

PR #165, opened 2021-08-02
Cleanup image building jobs

ariefrahmansyah:

On deploying a Pyfunc model, Merlin creates Kubernetes Job to build and push the image. Once completed, the Kubernetes Job and its pod still exist in the cluster. The increasing number of undeleted Jobs could lead to side effects such as slower Kubernetes API server, failure of new Pyfunc image building job.

```release-note
NONE
```

**Checklist**

- [x] Tested locally


ariefrahmansyah:
Please take a look at the early draft PR to automate pyfunc image building cleanup @pradithya @tiopramayudi

api/config/config.go
```diff
@@ -110,6 +111,8 @@ type ImageBuilderConfig struct {
 	BuildNamespace               string `envconfig:"IMG_BUILDER_NAMESPACE" default:"mlp"`
 	DockerRegistry               string `envconfig:"IMG_BUILDER_DOCKER_REGISTRY"`
 	BuildTimeout                 string `envconfig:"IMG_BUILDER_TIMEOUT" default:"10m"`
+	// How long to keep the image building job resource in the Kubernetes cluster. Default: A week (168 hours).
+	Retention time.Duration `envconfig:"IMG_BUILDER_RETENTION" default:"168h"`
```
pradithya, line 115: I think we should reduce the default retention to 2 days. Having 1 week as default is not useful because the docker container will already be deleted by kubelet and user won't be able to fetch the logs anyway.

api/imagebuilder/janitor.go
```diff
@@ -0,0 +1,91 @@
...
+		if now.Sub(job.Status.CompletionTime.Time) > j.cfg.Retention {
+			expiredJobs = append(expiredJobs, job)
+		}
+	}
+
+	return expiredJobs, nil
+}
+
+func (j *Janitor) deleteJobs(expiredJobs []batchv1.Job) error {
+	for _, job := range expiredJobs {
```
pradithya, line 72: Is there batch deletion api?

charts/merlin/values.yaml
```diff
@@ -63,6 +63,7 @@ merlin:
     timeout: "30m"
     contextSubPath: "python/pyfunc-server"
     predictionJobContextSubPath: "python/batch-predictor"
+    retention: "168h"
```
pradithya, line 66: I think 48 hours is better default

api/cmd/api/cronjob.go
```diff
@@ -0,0 +1,35 @@
...
+
+	imageBuilderJanitor := dependencies.imageBuilderJanitor
+
+	c, err := cronjob.New()
+	if err != nil {
+		return err
+	}
+
+	c.AddFunc("@hourly", tracker.TrackMetrics)
+	c.AddFunc("@daily", imageBuilderJanitor.CleanJobs)
```
pradithya, line 30: It's better to delete the expired job hourly too.

api/imagebuilder/janitor.go
ariefrahmansyah: Yup, #TIL k8s' client-go has DeleteCollection method: https://github.com/kubernetes/client-go/blob/v0.16.6/kubernetes/typed/batch/v1/job.go#L163-L177

Since you are proposing to reduce the interval hence increase the frequency, I think it's better to do batch deletion to reduce the number of API call to k8s API server. Is this what you suggesting?

pradithya: Yes, it's better to use batch API

api/imagebuilder/janitor.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (j *Janitor) deleteJobs() error {
+	deleteOptions := &metav1.DeleteOptions{}
+	if j.cfg.DryRun {
+		deleteOptions.DryRun = []string{"All"}
+	}
+
+	listOptions := metav1.ListOptions{LabelSelector: labelOrchestratorName + "=merlin"}
+
+	return j.cc.DeleteJobs(j.cfg.BuildNamespace, deleteOptions, listOptions)
```
pradithya, line 53: This command will also delete all new job and disregard the`Retention` configuration.

api/imagebuilder/janitor.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	"github.com/gojek/merlin/cluster"
+	"github.com/gojek/merlin/log"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Janitor cleans the finished image building jobs.
+type Janitor struct {
+	cc  cluster.Controller
```
pradithya, line 13: I think it's cleaner to just directly depend to kubernetes API here

api/imagebuilder/janitor_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"time"
+
+	"github.com/gojek/merlin/cluster/mocks"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestJanitor_CleanJobs(t *testing.T) {
```
pradithya, line 14: Can you add test that prove it will only delete Merlin's image build jobs older than the retention period?

api/imagebuilder/janitor.go
ariefrahmansyah: I think it's better to refactor our cluster.Controller too. For now, I'm going to keep it as is. I'll keep your idea in mind for future changes.

pradithya (approved):
LGTM

pradithya:
Can you update the PR description before merging?

PR #169, opened 2021-08-18
Support multiple Feast endpoints on Standard Transformer

ariefrahmansyah:

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -302,16 +332,18 @@ func (fc *feastCall) do(
...
 
+	feastEndpoint := fc.feastEndpoint
+
 	startTime := time.Now()
 	feastResponse, err := fc.feastClient.GetOnlineFeatures(ctx, &feastRequest)
 	durationMs := time.Now().Sub(startTime).Milliseconds()
 	if err != nil {
-		feastLatency.WithLabelValues("error").Observe(float64(durationMs))
-		feastError.Inc()
+		feastLatency.WithLabelValues("error", feastEndpoint).Observe(float64(durationMs))
```
pradithya, line 341: Can you create ticket to update feast related dashboard.

api/cluster/templater_test.go
```diff
@@ -90,9 +90,13 @@ var (
 	}
 
 	standardTransformerConfig = config.StandardTransformerConfig{
-		ImageName:       "merlin-standard-transformer",
-		FeastServingURL: "serving.feast.dev:8081",
-		FeastCoreURL:    "core.feast.dev:8081",
+		ImageName:                   "merlin-standard-transformer",
+		DefaultFeastServingEndpoint: "serving.feast.dev:8081",
```
pradithya, line 94: Can we be more consistent on choosing name? I prefer to use `DefaultFeastServingURLs`

api/cluster/templater_test.go
```diff
@@ -90,9 +90,13 @@ var (
...
+		FeastServingEndpoints: config.FeastServingEndpoints{
```
pradithya, line 95: Same as https://github.com/gojek/merlin/pull/169/files#r690974985

api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -290,6 +305,21 @@ func (fr *FeastRetriever) newFeastCall(
 	}
 }
 
+func (fr *FeastRetriever) getFeastClient(endpoint string) feast.Client {
+	client, ok := fr.feastClients[Endpoint(endpoint)]
+	if !ok {
```
pradithya, line 310: I think this will hide issue if users incorrectly set the feast serving URL. They will create request to the default feast serving.

pradithya: I prefer to:
1. Check if the value is empty, if so use default URL
2. Check if the value exists, if not then throw error
3. Return existing client

api/pkg/transformer/feast/feature_retriever_test.go
```diff
@@ -922,6 +922,9 @@ func TestFeatureRetriever_RetrieveFeatureOfEntityInRequest(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			mockFeast := &mocks.Client{}
+			feastClients := Clients{}
```
pradithya, line 925: You are adding test where there is non-default URL right?

protos/merlin/transformer/spec/feast.proto
```diff
@@ -9,6 +9,7 @@ message FeatureTable {
   repeated Entity entities = 2;
   repeated Feature features = 3;
   string tableName = 4;
+  string servingEndpoint = 5;
```
pradithya, line 12: servingUrl ?

pradithya:
I've done my 1st review round, there are 3 follow ups:
1. Make the config and variable name consistent. The term URL and endpoint are interchangeable so we have to pick one. In our existing implementation mostly we use URL in config/variable. 
2. Add unit test for multiple feast serving case in feature_retrieval.go.
3. End to end test case.

api/cluster/templater.go
```diff
@@ -286,7 +286,17 @@ func (t *KFServingResourceTemplater) createTransformerSpec(modelService *models.
 			}
 		}
 		transformer.Image = t.standardTransformerConfig.ImageName
-		envVars = append(envVars, models.EnvVar{Name: transformerpkg.FeastServingURLEnvName, Value: t.standardTransformerConfig.FeastServingURL})
+
+		// TODO: (arief)
```
pradithya, line 290: What todo?

charts/merlin/values.yaml
```diff
@@ -89,7 +89,19 @@ merlin:
 
   transformer:
     feast:
-      servingURL: serving.feast.dev
+      defaultServingEndpoint: online-serving-redis.feast.dev
```
pradithya, line 92: Can you change to URL too?

pradithya:
I left small feedback, overall LGTM

api/cluster/templater.go
```diff
@@ -296,7 +300,7 @@ func (t *KFServingResourceTemplater) createTransformerSpec(modelService *models.
 			{Name: transformerpkg.JaegerSamplerType, Value: jaegerCfg.SamplerType},
 			{Name: transformerpkg.JaegerDisabled, Value: jaegerCfg.Disabled},
 		}
-		envVars = append(jaegerEnvVars, envVars...)
+		envVars = append(envVars, jaegerEnvVars...)
```
tiopramayudi, line 303: why you change the ordering? initially the purpose of put `jaegerEnvVars` first to let user overwrite the value of jaeger envs

ariefrahmansyah: No particular reason. At first, I got confused why my test failed, turns out the envVars order is different with the order of the code written. I also forgot about these ordering but I think moving forward we should use filtering for those allowed-to-be-modified env vars.

tiopramayudi: what kind of filtering? with this change user can't enable the jaeger tracing

ariefrahmansyah: Filtering on UI: https://github.com/gojek/merlin/blob/HEAD/ui/src/components/EnvVarsConfigTable.js#L41

Why can user enable the jaeger tracing ya if the order changed?

tiopramayudi: currently the default env variable JAEGER_DISABLED is set to true, and user wants to set JAEGER_DISABLED to false, user will add JAEGER_DISABLED value in env variables to false when about to deploy. So  initial `envVars` has `JAEGER_DISABLED` false because it is set by user. And later this `jaegerEnvVars` value is appended to `envVars`. Let say `jaegerEnvVars` only has [`JAEGER_DISABLED`: true] after appended the `envVars` will have value [`JAEGER_DISABLED`: false,  `JAEGER_DISABLED`: true] . The value that win will be the last one right, which is the default value?

ariefrahmansyah: Understood. Thanks for detailed explanation. I've update my PR to revert the order so Jaeger env vars is on top of the slice.

On another note, I think we should patch the `JAEGER_DISABLED` with user-provided value (if exist) instead of having two env var with the same name. It's out of scope of this PR and we should refactor it in the future.

api/cmd/transformer/.env.sample
```diff
@@ -2,12 +2,13 @@ LOG_LEVEL=ERROR
...
-MERLIN_TRANSFORMER_MODEL_NAME=echo-2
-MERLIN_TRANSFORMER_MODEL_PREDICT_URL=localhost:8081
-MERLIN_TRANSFORMER_PORT=8080
+MERLIN_TRANSFORMER_MODEL_NAME=echo
+MERLIN_TRANSFORMER_MODEL_PREDICT_URL=localhost:8082
+MERLIN_TRANSFORMER_PORT=8081
 MODEL_HYSTRIX_RETRY_COUNT=0
 
-FEAST_SERVING_URL=localhost:6566
+DEFAULT_FEAST_SERVING_ENDPOINT=localhost:6566
```
tiopramayudi, line 10: is this still being used?

api/cmd/transformer/.env.sample
```diff
@@ -2,12 +2,13 @@ LOG_LEVEL=ERROR
...
+FEAST_SERVING_ENDPOINTS=localhost:6566,localhost:6567
```
tiopramayudi, line 11: is this still being used?

tiopramayudi:
I left two little comments, beside that it LGTM

api/cmd/transformer/.env.sample
ariefrahmansyah: It is. I've updated the environment name to match the code.

api/cmd/transformer/.env.sample
ariefrahmansyah: It is. I've updated the environment name to match the code.

PR #171, opened 2021-08-22
Safely delete only succeeded image building jobs

### LINKED ISSUE #170: Panic on janitor - image building jobs deletion
Given a list of image building jobs on the cluster with one failed job:

```sh
$ kubectl get job
NAME                                                 COMPLETIONS   DURATION   AGE
merlin-e2e-ariefrahmansyah-std-transformer-4         0/1           6h31m      6h31m
merlin-e2e-ariefrahmansyah-std-transformer-5         1/1           3m43s      5h38m

$ kubectl get job merlin-e2e-ariefrahmansyah-std-transformer-4 -o json
...
    "status": {
        "conditions": [
            {
                "lastProbeTime": "2021-08-21T19:17:51Z",
                "lastTransitionTime": "2021-08-21T19:17:51Z",
                "message": "Job has reached the specified backoff limit",
                "reason": "BackoffLimitExceeded",
                "status": "True",
                "type": "Failed"
            }
        ],
        "failed": 4,
        "startTime": "2021-08-21T19:08:40Z"
    }
...
```

The Image Builder Janitor Cron panic with following error:

```go
2021/08/21 22:00:00 cron: panic running job: runtime error: invalid memory address or nil pointer dereference
goroutine 3629 [running]:
github.com/robfig/cron.(*Cron).runWithRecovery.func1(0xc000511180)
	/go/pkg/mod/github.com/robfig/cron@v1.2.0/cron.go:161 +0x9e
panic(0x203aa20, 0x3aef850)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/gojek/merlin/imagebuilder.(*Janitor).getExpiredJobs(0xc00118cf90, 0xc000582f00, 0x23f0f3a, 0x2d, 0x0, 0x0)
	/src/api/imagebuilder/janitor.go:81 +0x184
github.com/gojek/merlin/imagebuilder.(*Janitor).CleanJobs(0xc00118cf90)
	/src/api/imagebuilder/janitor.go:56 +0x7c
github.com/robfig/cron.FuncJob.Run(0xc001094b70)
	/go/pkg/mod/github.com/robfig/cron@v1.2.0/cron.go:92 +0x25
github.com/robfig/cron.(*Cron).runWithRecovery(0xc000511180, 0x2829060, 0xc001094b70)
	/go/pkg/mod/github.com/robfig/cron@v1.2.0/cron.go:165 +0x59
created by github.com/robfig/cron.(*Cron).run
	/go/pkg/mod/github.com/robfig/cron@v1.2.0/cron.go:199 +0x747
```

ariefrahmansyah:

In the previous implementation, the Image Building Janitor simply depends on the `job.Status.CompletionTime` field which could be `nil` if the job is still active or failed. This PR avoids panic by checking the `job.Status.Succeeded` value first and only proceed with the deletion of jobs if its value is positive. I also add Prometheus metrics (`image_builder_failed_jobs_count`) to track failed jobs. For now, Janitor will do nothing upon these failed jobs and manual intervention is needed from Merlin's administrator to debug them.

Fixes #170 

```release-note
NONE
```

**Checklist**

- [ ] Tested locally

### DIFF
--- a/api/imagebuilder/janitor.go
+++ b/api/imagebuilder/janitor.go
@@ -14,6 +14,12 @@ import (
 )
 
 var (
+	failedJobs = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: "merlin_api",
+		Name:      "image_builder_failed_jobs_count",
+		Help:      "The total number of failed jobs. These jobs won't be deleted by Janitor and must be manually troubleshooted by Merlin administrators.",
+	})
+
 	deleteJobsErrors = promauto.NewCounter(prometheus.CounterOpts{
 		Namespace: "merlin_api",
 		Name:      "image_builder_delete_jobs_error_count",
@@ -78,8 +84,15 @@ func (j *Janitor) getExpiredJobs() ([]batchv1.Job, error) {
 
 	now := time.Now()
 	for _, job := range jobs.Items {
-		if now.Sub(job.Status.CompletionTime.Time) > j.cfg.Retention {
-			expiredJobs = append(expiredJobs, job)
+		if job.Status.Failed > 0 {
+			failedJobs.Inc()
+			continue
+		}
+
+		if job.Status.Succeeded > 0 {
+			if now.Sub(job.Status.CompletionTime.Time) > j.cfg.Retention {
+				expiredJobs = append(expiredJobs, job)
+			}
 		}
 	}
 
--- a/api/imagebuilder/janitor.go
+++ b/api/imagebuilder/janitor.go
@@ -14,12 +14,6 @@ import (
 )
 
 var (
-	failedJobs = promauto.NewCounter(prometheus.CounterOpts{
-		Namespace: "merlin_api",
-		Name:      "image_builder_failed_jobs_count",
-		Help:      "The total number of failed jobs. These jobs won't be deleted by Janitor and must be manually troubleshooted by Merlin administrators.",
-	})
-
 	deleteJobsErrors = promauto.NewCounter(prometheus.CounterOpts{
 		Namespace: "merlin_api",
 		Name:      "image_builder_delete_jobs_error_count",
@@ -31,7 +25,7 @@ var (
 		Name:      "image_builder_delete_jobs_duration_ms",
 		Help:      "Image builder delete jobs histogram",
 		Buckets:   prometheus.ExponentialBuckets(64, 2, 5), // 64, 128, 256, 512, 1024, +Inf
-	}, []string{"status"})
+	}, []string{"result", "job_status"})
 )
 
 // Janitor cleans the finished image building jobs.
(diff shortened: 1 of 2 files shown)


pradithya:
What if you depend on creation time instead? Please avoid the necessity of manual intervention.

ariefrahmansyah:
> What if you depend on creation time instead? Please avoid the necessity of manual intervention. I am refering to this https://github.com/kubernetes/apimachinery/blob/19d98351410f7b9c6b2910a7c8713b064995f180/pkg/apis/meta/v1/types.go#L202

We can use CreationTime or job.Status.StartTime (https://github.com/kubernetes/api/blob/master/batch/v1/types.go#L221) to be more precise. But are we going to ignore the status of the job? In another word, are we going to delete active and failed jobs if their duration is more than retention time (`now - start time > image builder's retention`)?

Yes, we can automatically delete failed jobs but we may lose visibility about these jobs leading to harder troubleshooting. I prefer to surface those errors so we can act on them asap. I want to avoid undetected and unnoticeable bugs on our image building job. The Prometheus metrics should have us detect the error (we must create dashboard and alerts first).

pradithya:
> > What if you depend on creation time instead? Please avoid the necessity of manual intervention. I am refering to this https://github.com/kubernetes/apimachinery/blob/19d98351410f7b9c6b2910a7c8713b064995f180/pkg/apis/meta/v1/types.go#L202
> 
> We can use CreationTime or job.Status.StartTime (https://github.com/kubernetes/api/blob/master/batch/v1/types.go#L221) to be more precise. But are we going to ignore the status of the job? In another word, are we going to delete active and failed jobs if their duration is more than retention time (`now - start time > image builder's retention`)?
> 
> Yes, we can automatically delete failed jobs but we may lose visibility about these jobs leading to harder troubleshooting. I prefer to surface those errors so we can act on them asap. I want to avoid undetected and unnoticeable bugs on our image building job. The Prometheus metrics should have us detect the error (we must create dashboard and alerts first).

Yes, we should delete all jobs if it has gone beyond retention. We should still emit the job status metrics though so we know if there is problem. Those 2 things can be independent.

The failed jobs doesnt give us much information anyway as the docker log already dissapear and we can just use stackdriver.

ariefrahmansyah:
> Yes, we should delete all jobs if it has gone beyond retention. We should still emit the job status metrics though so we know if there is problem. Those 2 things can be independent. But then, the metrics that's created after the job pass its retention time is probably not useful either as it's too late to be actionable. There is existing metrics for this ( ie. Image build job status metrics)
> 
> The failed jobs doesnt give us much information anyway as the docker log already dissapear and we can just use stackdriver.

Understood. I've updated the PR as you've suggested. I entirely forget that we cannot access container logs after several amount of time after the job completed / failed. During my local tests, the log is still exists.

pradithya (approved):
Thanks, LGTM!

PR #175, opened 2021-08-30
Rework image builder

tiopramayudi:
Currently image builder job doesn't run in dedicated nodepools, which sometimes it cause disk pressure when a node running image building jobs. Hence in this PR, image building job will be run in the dedicated nodepools. Also previously job configuration was hardcoded, in this PR we make it configurable

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/cmd/api/setup.go
```diff
@@ -165,6 +168,10 @@ func initImageBuilder(cfg *config.Config, vaultClient vault.Client) (webserviceB
 		DockerRegistry:       cfg.ImageBuilderConfig.DockerRegistry,
 		ContextSubPath:       cfg.ImageBuilderConfig.PredictionJobContextSubPath,
 		BuildTimeoutDuration: timeout,
+		KanikoImage:          cfg.ImageBuilderConfig.KanikoImage,
+		MaximumRetry:         cfg.ImageBuilderConfig.MaximumRetry,
+		CpuRequest:           cfg.ImageBuilderConfig.CpuRequest,
+		NodePoolName:         cfg.ImageBuilderConfig.NodePoolName,
```
pradithya, line 174: Can we accept yaml instead of node name so that merlin's administrator can have better flexibility? Also make the config optional.

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -358,6 +360,17 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
 							},
 						},
 					},
+					Tolerations: []v1.Toleration{
```
pradithya, line 363: We shouldn't hardcode it. Instead it should be configurarable the same way node selector does.

api/cmd/api/setup.go
pradithya: The configuration should be compatible with the `NodeSelector` field in kubernetes.

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -72,18 +71,15 @@ const (
 	labelEnvironment      = "gojek.com/environment"
 	labelOrchestratorName = "gojek.com/orchestrator"
 	labelUsersHeading     = "gojek.com/user-labels/%s"
+
+	podTolerationKey   = "image-build-job"
+	podTolerationValue = "true"
+	podNodeSelectorKey = "cloud.google.com/gke-nodepool"
```
pradithya, line 77: We shouldn't hardcode this as well. It will only work in GKE.

pradithya:
The node selector and node toleration configuration should be flexible and compatible with its kubernetes fields. 
I.e. 

node selector config should accept `map[string]string` :
```
nodeSelectors:
   key_a: value_a
   key_b: value_b
```

and node toleration config should accept structure compatible with `[]v1.Toleration`:
```
tolerations:
- key: "key1"
  operator: "Equal"
  value: "value1"
  effect: "NoSchedule"
- key: "key1"
  operator: "Exists"
  effect: "NoSchedule"
```
But since we are using env var for config, those value should be represented as json instead of yaml.

api/cmd/api/setup.go
ariefrahmansyah: At first, I thought we are migrating to a file-based configuration. Thinking out loud here, I've been thinking to move from environment variables-based configuration to file-based configuration since the number of our configurations already grew relatively long. But since this topic is out of scope for this PR, we can revisit it later.

For this PR scope, I agree with Aria to have this NodePoolName compliance with Kubernetes' NodeSelector,

charts/merlin/values.yaml
```diff
@@ -68,6 +68,12 @@ merlin:
     contextSubPath: "python/pyfunc-server"
     predictionJobContextSubPath: "python/batch-predictor"
     retention: "48h"
+    kanikoImage: "gcr.io/kaniko-project/executor:v1.1.0"
```
ariefrahmansyah, line 71: Should we use the newer version of Kaniko image as default?

pradithya:
LGTM, can you rebase first before merging?

pradithya (approved):

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -285,18 +280,14 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
...
-	var labels = map[string]string{
+	labels := map[string]string{
 		labelTeamName:         project.Team,
 		labelStreamName:       project.Stream,
 		labelAppName:          model.Name,
 		labelEnvironment:      c.config.Environment,
 		labelOrchestratorName: "merlin",
 	}
 
-	for _, label := range project.Labels {
```
ariefrahmansyah, line 296: Please help me refresh my memory, why we remove this?

ariefrahmansyah (approved):
LGTM

api/pkg/imagebuilder/imagebuilder.go
tiopramayudi: because previously it is breaking the label name constraints that used in k8s, since it is not used for cost labelling i prefer to remove it

PR #176, opened 2021-08-30
UI improvements on model list, version list, and deployment pages

ariefrahmansyah:

**1. Fix columns width**

Model list page, before:

image

Model list page, after:

image

Model version list page, before:

image

Model version list page, after:

image

**2. Fix changing transformer type to None**

Before, even though we revert back to None, the deployment form wizard still displays steps from previous selected transformer type:

image

After:

image

**Checklist**

- [x] Tested locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,185 @@
-.idea
-bin/
-**/*.ipynb_checkpoints/
+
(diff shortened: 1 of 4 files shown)


tiopramayudi:
LGTM

PR #177, opened 2021-09-02
Add retry on check existence of PyFunc image

ariefrahmansyah:

To make check image existence prior to Pyfunc model deployment more reliable.

```release-note
NONE
```

**Checklist**

- [x] Tested locally

### DIFF
--- a/api/pkg/imagebuilder/imagebuilder.go
+++ b/api/pkg/imagebuilder/imagebuilder.go
@@ -218,9 +218,10 @@ func (c *imageBuilder) imageRefExists(imageName, imageTag string) (bool, error) 
 			if terr.StatusCode == http.StatusNotFound {
 				return false, nil
 			}
-		} else {
-			return false, errors.Wrapf(err, "error getting image tags for %s", repo)
 		}
+
+		// If it's not transport error or 404, raise error
+		return false, errors.Wrapf(err, "error getting image tags for %s", repo)
 	}
 
 	for _, tag := range tags {
--- a/api/pkg/imagebuilder/imagebuilder.go
+++ b/api/pkg/imagebuilder/imagebuilder.go
@@ -36,7 +36,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/google"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
-	"github.com/pkg/errors"
 )
 
 var maxCheckImageRetry int = 2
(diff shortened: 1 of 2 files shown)


api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -214,13 +216,24 @@ func (c *imageBuilder) imageRefExists(imageName, imageTag string) (bool, error)
...
 			// If image not found, it's not exist yet
 			if terr.StatusCode == http.StatusNotFound {
 				return false, nil
 			}
-		} else {
-			return false, errors.Wrapf(err, "error getting image tags for %s", repo)
+
+			// If unauthorized, it's more likely due to expired token. So, let's try again
+			if terr.StatusCode == http.StatusUnauthorized {
+				if retryCounter <= maxCheckImageRetry {
```
tiopramayudi, line 228: should it be < instead of <=

ariefrahmansyah: I put `<=` because I want to retry twice. Otherwise it's only retry once.

tiopramayudi: it will retry thrice if you use <= . 
let simulate
 imageRefExist(retry=0)
     failed -> imageRefExist(retry=1)
                     failed -> imageRefExist(retry=2)
                                       failed -> imageRefExist(retry=3)
                                                         failed -> no retry

ariefrahmansyah: Woop. You're right. Updated as suggested

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -214,13 +216,24 @@ func (c *imageBuilder) imageRefExists(imageName, imageTag string) (bool, error)
...
 				return false, nil
 			}
-		} else {
-			return false, errors.Wrapf(err, "error getting image tags for %s", repo)
+
+			// If unauthorized, it's more likely due to expired token. So, let's try again
+			if terr.StatusCode == http.StatusUnauthorized {
+				if retryCounter < maxCheckImageRetry {
+					log.Infof("%d retry listing tags for %s", retryCounter+1, imageName)
+					return c.imageRefExists(imageName, imageTag, retryCounter+1)
```
tiopramayudi, line 230: can we add delay between retry, let say 1s?

ariefrahmansyah: Added. I was thinking about it before, but forget to implement it. Thanks for the reminder!

api/pkg/imagebuilder/imagebuilder_test.go
```diff
@@ -1052,11 +1052,13 @@ func Test_kanikoBuilder_imageRefExists(t *testing.T) {
 	type args struct {
 		imageName string
 		imageTag  string
+		retry     int
```
tiopramayudi, line 1055: is there a way to create test to simulate retry cases? if not possible add case where `retry` value greater than 0 should be sufficient

tiopramayudi:
i left one more comment, but the rest is LGTM

PR #179, opened 2021-09-09
Fix swagger deployment 

tiopramayudi:
…values

Swagger container needs to fetch swagger spec from merlin repository. Current approach is not working because swagger spec could not fetched from merlin service, hence we need to fetch this from github directly

Fixes #

Fix `args` command when creating swagger container and adding new value in swagger helm values to specifies specUrl that needs to be used by swagger container

```release-note
NONE
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/charts/merlin/values.yaml
+++ b/charts/merlin/values.yaml
@@ -337,7 +337,6 @@ swagger:
     tag: v3.23.5
   apiHost: "merlin.dev"
   basePath: "/api/merlin/v1"
-  specUrl: "https://raw.githubusercontent.com/gojek/merlin/main/swagger.yaml"
   service:
     internalPort: 8080
     externalPort: 8080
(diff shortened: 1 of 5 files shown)


charts/merlin/values.yaml
```diff
@@ -337,6 +337,7 @@ swagger:
     tag: v3.23.5
   apiHost: "merlin.dev"
   basePath: "/api/merlin/v1"
+  specUrl: "https://raw.githubusercontent.com/gojek/merlin/main/swagger.yaml"
```
ariefrahmansyah, line 340: There might be a case when the swagger spec differs from the Merlin server. For example, we run Merlin 0.15 but swagger spec is still using the `main` branch.

I prefer to keep chart template and update the API to serve swagger.yaml

ariefrahmansyah (approved):
LGTM

PR #180, opened 2021-09-09
Add recoverer on transformer

ariefrahmansyah:

Everytime there's panic on standard transformer, the app will just stop serving traffic. This PR adds recovered to make standard transformer more reliable.

```release-note
NONE
```

**Checklist**

- [x] Tested locally

### DIFF
--- a/api/pkg/transformer/server/server.go
+++ b/api/pkg/transformer/server/server.go
@@ -305,7 +305,6 @@ func recoveryHandler(next http.Handler) http.Handler {
 		defer func() {
 			if err := recover(); err != nil {
 				debug.PrintStack()
-				w.WriteHeader(http.StatusInternalServerError)
 				response.NewError(http.StatusInternalServerError, fmt.Errorf("panic: %v", err)).Write(w)
 			}
 		}()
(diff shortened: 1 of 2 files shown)


api/pkg/transformer/server/server.go
```diff
@@ -295,6 +300,19 @@ func (s *Server) Run() {
 	}
 }
 
+func recoveryHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
+			if err := recover(); err != nil {
+				debug.PrintStack()
+				w.WriteHeader(http.StatusInternalServerError)
```
pradithya, line 308: Can we return error message to the caller?

```
response.NewError(http.StatusInternalServerError, errors.Wrapf(err, "panic")).Write(w)
```

pradithya:
Can you add test?

api/pkg/transformer/server/server.go
pradithya: I think `w.WriteHeader(http.StatusInternalServerError)` should be removed as `response.NewError(http.StatusInternalServerError, fmt.Errorf("panic: %v", err)).Write(w)` already write the status code.

pradithya:
Other than https://github.com/gojek/merlin/pull/180#discussion_r705880658, it looks good to me

PR #182, opened 2021-09-22
Feature scaling and encoders

tiopramayudi:

Feature scaling and encoding are common preprocessing step in data science, and these are missing in current standard transformer. Hence we introduce this feature in this PR. The changes are:
* New input operation to create encoder specification
* Adding scale column, and encode colum steps in table transformation

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


protos/merlin/transformer/spec/encoder.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package merlin.transformer;
+
+option go_package = "github.com/gojek/merlin/pkg/transformer/spec";
+
+message Encoder {
+  string name = 1;
+  oneof encoderConfig {
```
ariefrahmansyah, line 9: I think Encoder message still need an `encoderType` field here to store the encoderConfig type value. There are three reasoning behind this:
1. If user accidentally (or intentionally) specify two encoderConfigs, we can refer to `encoderType` to decide which one to use.
2. It helps UI component to preserve the other config. For example, I created and configured ordinal encoder, then I switch to other encoder type. But when I want to use ordinal encoder again, the config is still there.
3. In the API implementation, we can refer to `encoderType` to do casting instead of checking if the config is nil or not.

protos/merlin/transformer/spec/table.proto
```diff
@@ -75,3 +78,16 @@ enum JoinMethod {
   CROSS = 5;
   CONCAT = 6;
 }
+
+message ScaleColumn {
+  string column = 1;
+  oneof scalerConfig {
```
ariefrahmansyah, line 84: Similar to `encoderType` on Encoder message, I think ScaleColumn also needs `scalerType`

protos/merlin/transformer/spec/encoder.proto
tiopramayudi: I prefer not to add this because there is not much value by adding `encoderType` cause it is will be redundant with the config, and also prone to make mistake if the config specified is mismatch with the type (if user use sdk)

* If user accidentally specify two encoderConfigs, it will choose the latest one specified based on the proto `oneOf` behaviour, and `encoderType` can't decide which one to use.
* I think UI can still achieve that behaviour without encoderType, or with encoderType which the type won't be processed in the backend since it only need the config 
* there won't be issue with config is nil IMO, because it will failed in deployment (because invalid proto)

Makefile
```diff
@@ -176,5 +176,5 @@ gen-proto:
 	@echo "> Generating specification configuration from Proto file..."
 	@cd protos/merlin && \
 		protoc -I=. --go_out=../../api --go_opt=module=github.com/gojek/merlin \
-		--go-json_out=../../api --go-json_opt=module=github.com/gojek/merlin \
```
pradithya, line 179: Why was it changed?

tiopramayudi: it is not working previously

ariefrahmansyah (approved):
LGTM with nit comments

PR #184, opened 2021-09-29
Image builder: More retry if check image fail

ariefrahmansyah:
Refactor checking the previous image of image builder jobs using https://github.com/cenkalti/backoff.

Now we will always retry (using a backoff mechanism) regarding the error status.

### DIFF
--- a/api/pkg/imagebuilder/imagebuilder.go
+++ b/api/pkg/imagebuilder/imagebuilder.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	backoff "github.com/cenkalti/backoff/v4"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
@@ -38,7 +39,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 )
 
-var maxCheckImageRetry int = 2
+var maxCheckImageRetry uint64 = 2
 
 type ImageBuilder interface {
 	// BuildImage build docker image for the given model version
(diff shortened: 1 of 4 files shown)


api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -222,13 +222,10 @@ func (c *imageBuilder) imageRefExists(imageName, imageTag string, retryCounter i
...
 
-			// If unauthorized, it's more likely due to expired token. So, let's try again
-			if terr.StatusCode == http.StatusUnauthorized {
-				if retryCounter < maxCheckImageRetry {
-					time.Sleep(1 * time.Second)
-					log.Infof("retry (%d) listing tags for %s", retryCounter+1, imageName)
-					return c.imageRefExists(imageName, imageTag, retryCounter+1)
-				}
+			if retryCounter < maxCheckImageRetry {
+				time.Sleep(1 * time.Second)
```
pradithya, line 226: Can we have exponential backoff retry? You can implement it yourself or using this library https://github.com/cenkalti/backoff

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -222,13 +222,10 @@ func (c *imageBuilder) imageRefExists(imageName, imageTag string, retryCounter i
...
+				log.Infof("retry (%d) listing tags for %s", retryCounter+1, imageName)
+				return c.imageRefExists(imageName, imageTag, retryCounter+1)
```
pradithya, line 228: I just realised this is recursive call. Can we refactor this function so that the implementation is cleaner. I prefer to have loop, instead of recursive call.

pradithya: Actually, it will be cleaner if you use the library (https://github.com/cenkalti/backoff)

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -96,10 +97,22 @@ func newImageBuilder(kubeClient kubernetes.Interface, config Config, nameGenerat
 func (c *imageBuilder) BuildImage(project mlp.Project, model *models.Model, version *models.Version) (string, error) {
 	// check for existing image
 	imageName := c.nameGenerator.generateDockerImageName(project, model)
-	imageExists, err := c.imageRefExists(imageName, version.ID.String(), 0)
+
+	var imageExists bool = false
```
pradithya, line 101: nit: 
```
imageExists := false
```

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -194,14 +207,37 @@ func (c *imageBuilder) imageRef(project mlp.Project, model *models.Model, versio
 	return fmt.Sprintf("%s:%s", c.nameGenerator.generateDockerImageName(project, model), version.ID)
 }
 
+func (c *imageBuilder) imageExists(imageName, imageTag string) bool {
+	var imageExists bool = false
```
pradithya, line 211: nit:
```
imageExists := false
```

pradithya (approved):
LGTM

PR #187, opened 2021-10-04
Fix cache key collision when 2 tables have same list of entity

pradithya:

Currently cache key in standard transformer is constructed by entity values and project name. However two or more feature retrieval table can have same for both values which cause cache key collision.
This PR also do some refactoring in feature_retriever.go

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally


api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -95,6 +92,7 @@ type Options struct {
 	BatchSize               int           `envconfig:"FEAST_BATCH_SIZE" default:"50"`
 	CacheEnabled            bool          `envconfig:"FEAST_CACHE_ENABLED" default:"true"`
 	CacheTTL                time.Duration `envconfig:"FEAST_CACHE_TTL" default:"60s"`
+	CacheSizeInMB           int           `envconfig:"CACHE_SIZE_IN_MB" default:"100"`
```
ariefrahmansyah, line 95: I prefer `FEAST_CACHE_SIZE_IN_MB` as env name

api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -529,90 +340,18 @@ func getColumnNames(config *spec.FeatureTable) []string {
...
-			results = append(results, base64.StdEncoding.EncodeToString(bytes))
-		}
-		return results, types.ValueType_STRING_LIST, nil
-	default:
-		return nil, types.ValueType_INVALID, fmt.Errorf("unknown feature cacheValue type: %T", val.Val)
-	}
+	return entitySet
 }
 
 func mergeColumnTypes(dst []types.ValueType_Enum, src []types.ValueType_Enum) []types.ValueType_Enum {
```
ariefrahmansyah, line 357: nit: move this function to `types.go`

ariefrahmansyah (approved):
LGTM with nit comments.

api/pkg/transformer/feast/feature_retriever.go
pradithya: It will be a breaking change if we rename the env var.

PR #188, opened 2021-10-06
Update pyfuncserver to use new mlflow version

tiopramayudi:
With the recent release of mlflow sdk, there is breaking changes in our pyfunc server because we rely on internal implementation of mlflow. This PR will remove usage of internal implementation and use the API that specified by the SDK

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/pyfunc-server/pyfuncserver/test_model.py
+++ b/python/pyfunc-server/pyfuncserver/test_model.py
@@ -26,25 +26,34 @@ import tornado.web
 from prometheus_client import Counter, Gauge
 
 from pyfuncserver import PyFuncModel
+from merlin.model import PYFUNC_EXTRA_ARGS_KEY, PYFUNC_MODEL_INPUT_KEY
 
-
-class ModelImpl(mlflow.pyfunc.PythonModel):
+class NewModelImpl(mlflow.pyfunc.PythonModel):
     def load_context(self, context):
         self.initialize(context.artifacts)
         self._use_kwargs_infer = True
 
-    def predict(self, model_inputs, **kwargs):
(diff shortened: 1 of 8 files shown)


python/sdk/merlin/model.py
```diff
@@ -1453,11 +1455,20 @@ def _wait_build_complete(self, logs):
...
+    
     def load_context(self, context):
         self.initialize(context.artifacts)
         self._use_kwargs_infer = True
 
-    def predict(self, model_input, **kwargs):
+    def predict(self, context,  model_input):
+        extra_args = model_input.get(PYFUNC_EXTRA_ARGS_KEY, {})
+        input = model_input.get(PYFUNC_MODEL_INPUT_KEY, {})
+        if extra_args is not None and (type(extra_args) is dict) :
```
pradithya, line 1466: Why do we need to add additional condition of ` and (type(extra_args) is dict)` ?

python/pyfunc-server/pyfuncserver/test_model.py
```diff
@@ -26,25 +26,34 @@
 from prometheus_client import Counter, Gauge
 
 from pyfuncserver import PyFuncModel
+from merlin.model import PYFUNC_EXTRA_ARGS_KEY, PYFUNC_MODEL_INPUT_KEY
 
-
-class ModelImpl(mlflow.pyfunc.PythonModel):
+class NewModelImpl(mlflow.pyfunc.PythonModel):
```
pradithya, line 31: Can we add test for new model separately? and ensure the old model still satisfy this.

PR #193, opened 2021-10-29
Publish charts to turing-ml Helm repository.

ashwinath:

Turing requires a publicly hosted Helm chart to deploy Turing correctly. As of now, Turing depends on Merlin for getting the environments available in the cluster.

I have fixed the migration of stable repository from `@stable` to `https://charts.helm.sh/stable` as well as it will cause problems while packaging

This PR publishes the helm charts into https://turing-ml.github.io/charts/. The user would be able to install Merlin via `helm install turing/merlin`. Eventually, I would suspect that this will be moved to another repository once we have figured out where all the charts should be hosted eventually. For now this is the only repository that has been set up and due to time constraints, this should suffice.

One concern that needs to be addressed in a separate PR is that we should be publishing a public release for the Merlin container. The tagged versions are based on branch names and not built on tags.

Fixes None

```release-note
release public helm chart at https://turing-ml.github.io/charts/
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -365,3 +365,20 @@ jobs:
 #        run: ./merlin/scripts/e2e/deploy-merlin.sh "merlin/charts/merlin"
 #      - name: Run E2E test
 #        run: ./merlin/scripts/e2e/run-e2e.sh
+
+  release-charts:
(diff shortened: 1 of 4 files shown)


ashwinath:
Deciding to publish charts manually for 0.9.0 instead, this won't be required.

PR #197, opened 2021-11-23
Add time related built in function and cumulative value function

tiopramayudi:

In this PR we add several built in functions that can be used as expression in standard transformer spec. Below are the functions:
* `IsWeekend(timestamp interface{}, timezone string) interface{}`. Return whether a timestamp fall in weekend or not
* `DayOfWeek(timestamp interface{}, timezone string)`. Return int day indicate day of the week.
* `ParseTimeStampsWithFormat(timestamp interface{}, timezone, format string) interface{}`. Parsing timestamp into specific format
* `CumulativeValue(values interface{}) []float64`. Accumulate values based on the index, e.g values `[1,2,3] => [1, 1+2, 1+2+3] => [1, 3, 6]`

Fixes #

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


api/pkg/transformer/pipeline/testdata/valid_table_join_multiple_columns.yaml
```diff
@@ -34,17 +34,29 @@ transformerConfig:
...
+          inputTable: result_table
+          outputTable: result_table
+          steps:
+            - updateColumns:
+              - column: is_weekend
+                expression: IsWeekend(driver_table.Col('eta'), 'Asia/Jakarta')
+              - column: day_of_week
+                expression: DayOfWeek(driver_table.Col('eta'), 'Asia/Jakarta')
+              - column: pretty_date
+                expression: ParseTimeStampsWithFormat(driver_table.Col('eta'), 'Asia/Jakarta', '2006-01-02')
```
ariefrahmansyah, line 57: I think `ParseTimeStampsWithFormat` as function name is quite long. Wdyt about `FormatTimestamp` for function name?

ariefrahmansyah (approved):
lgtm with nit comment

api/pkg/transformer/pipeline/testdata/valid_table_join_multiple_columns.yaml
pradithya: Aggree, I think `FormatTimestamp` is better name

pradithya:
What is the usecase of `CumulativeValue`?

api/pkg/transformer/symbol/function/time.go
```diff
@@ -0,0 +1,23 @@
+package function
+
+import "time"
+
+func IsWeekend(timestamp int64, location *time.Location) int {
```
pradithya, line 5: Can you add documentation?

api/pkg/transformer/symbol/function/time.go
```diff
@@ -0,0 +1,23 @@
...
+	timeFromTs := time.Unix(timestamp, 0).In(location)
+	day := timeFromTs.Weekday()
+	// if it is saturday or sunday return 1 else 0
+	if day == time.Saturday || day == time.Sunday {
+		return 1
```
pradithya, line 10: Why don't it returns boolean?

api/pkg/transformer/symbol/registry.go
```diff
@@ -195,7 +195,57 @@ func (sr Registry) PolarAngle(latitude1 interface{}, longitude1 interface{}, lat
...
+	return sr.processTimestampFunction(timestamp, timezone, timeFn)
+}
+
+// FormatTimestamp will convert timestamp into formatted date string
+// timestamp can be:
+// - Json path string
+// - Slice / gota.Series
+// - int64 value
+func (sr Registry) FormatTimestamp(timestamp interface{}, timezone, format string) interface{} {
+	timeFn := func(ts int64, tz *time.Location) interface{} {
```
pradithya, line 223: Wouldn't it be slow if we recreate a new function every time it's being called?

api/pkg/transformer/symbol/time.go
```diff
@@ -0,0 +1,146 @@
...
+//
+// Examples:
+// Request JSON: {"booking_time": "2021-12-01 11:30:00"}
+// Expression: ParseDateTime($.booking_time, "2006-01-02 15:04:05")
+// Output: 2021-12-01 11:30:00 UTC
+//
+// Request JSON: {"booking_time": "2021-12-01 11:30:00", "timezone": "Asia/Jakarta"}
+// Expression: ParseDateTime($.booking_time, "2006-01-02 15:04:05", $.timezone)
+// Output: 2021-12-01 11:30:00 +07:00
+func (sr Registry) ParseDateTime(datetime interface{}, format string, timezone ...string) interface{} {
```
pradithya, line 106: Why is the `timezone` variadic?

pradithya: It seems that ParseDateTime can follow same implementation using `processTimestampFunction`

api/pkg/transformer/symbol/statistics.go
```diff
@@ -0,0 +1,37 @@
...
+			}
+			prevCumulativeVal := float64(0)
+			if idx > 0 {
+				prevCumulativeVal = values[idx-1]
+			}
+			values = append(values, prevCumulativeVal+floatVal)
+		}
+		return values
+	default:
+		panic("the values should in array format")
```
pradithya, line 35: `the values should be in array format`

api/pkg/transformer/symbol/time.go
ariefrahmansyah: > Why is the timezone variadic?

As shown in the example in the comment, it allows users to ParseDateTime without considering any timezone. I prefer the user to ignore the timezone field instead of passing an empty string ("").

> It seems that ParseDateTime can follow same implementation using processTimestampFunction

No, it cannot. `processTimestampFunction` assumes timestamp which is an integer. Date time is a string.

api/pkg/transformer/symbol/registry.go
ariefrahmansyah: No, it's not slower. I put the benchmark result in time_bench_test.go file.

api/pkg/transformer/symbol/time_test.go
```diff
@@ -0,0 +1,658 @@
...
+				sr.SetModelResponseJSON(responseJSONObj)
+			}
+
+			got := sr.ParseTimestamp(tC.timestamp)
+			assert.Equal(t, tC.expectedValue, got)
+		})
+	}
+}
+
+func TestSymbolRegistry_IsWeekend(t *testing.T) {
```
pradithya, line 96: Can you add test where input is a `Series`?

ariefrahmansyah: Tio already added that: https://github.com/gojek/merlin/blob/std-trf-built-in-function/api/pkg/transformer/symbol/time_test.go#L117-L172

api/pkg/transformer/symbol/time.go
pradithya: > As shown in the example in the comment, it allows users to ParseDateTime without considering any timezone. I prefer the user to ignore the timezone field instead of passing an empty string ("").


Can we make it consistent with other expression then?

ariefrahmansyah: Which expression ya?
If we want to align with others, it's better to not have location here as variadic

pradithya: Other time-related expression that @tiopramayudi created. Yes, either the others follow this or the other way around. But please be consistent.

ariefrahmansyah:
Ready for review @pradithya @tiopramayudi

api/pkg/transformer/symbol/time.go
```diff
@@ -0,0 +1,241 @@
...
+			values = append(values, timeTransformerFn(tsInt64, location))
+		}
+		return values
+	default:
+		tsInt64, err := converter.ToInt64(ts)
+		if err != nil {
+			panic(err)
+		}
+
+		timeLocation, err := LoadLocationCached(fmt.Sprintf("%s", tz))
```
tiopramayudi, line 126: what happen if `tz` is:
* slice with single value
* slice with multiple values

ariefrahmansyah: it will return error, hence panic because the location is not found

ariefrahmansyah: do you think we need to validate the timezone data type here? we expect timezone to be single value to match timestamp

api/pkg/transformer/symbol/time.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+
+		return dateTime
+	}
+}
+
+type LocationCache map[string]*time.Location
+
+var (
+	locationCache     = LocationCache{}
```
pradithya, line 221: Shall we use sync.Map instead? Based on the benchmark (write less, read more) the performance is the best. https://www.fatalerrors.org/a/map-concurrency-performance-and-principle-analysis-of-golang.html

api/pkg/transformer/symbol/time.go
pradithya: I think it's better to be defensive and check the type as well

docs/user-guide/transformer.md
```diff
@@ -85,76 +85,14 @@ transformerConfig:
       features: # list of features to be retrieved
         - name: merchant_t1_discovery:t1_estimate # feature name. make sure to include feature set
           valueType: DOUBLE # feature value type
-          defaultValue: '0' # default value if feature value not exist
+          defaultValue: "0" # default value if feature value not exist
 ```
 
 > To learn more about JSON Path syntax, please go to [this link](https://goessner.net/articles/JsonPath/), [this](https://restfulapi.net/json-jsonpath/) or [this](https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html).
 
 ### List of Supported UDFs
```
pradithya, line 93: It's not UDF. UDF is User-Defined Function. We should replace the mention of UDF in our document with "expression"

docs/user-guide/transformer_udf.md
```diff
@@ -0,0 +1,376 @@
+# Standard Transformer UDFs
```
pradithya, line 1: Built-in expression

docs/user-guide/transformer_udf.md
```diff
@@ -0,0 +1,376 @@
...
+| Geospatial | [S2ID](#s2id)                           |
+| Geospatial | [HaversineDistance](#haversinedistance) |
+| Geospatial | [PolarAngle](#polarangle)               |
+| JSON       | [JsonExtract](#jsonextract)             |
+| Time       | [Now](#now)                             |
+| Time       | [DayOfWeek](#dayofweek)                 |
+| Time       | [IsWeekend](#isweekend)                 |
+| Time       | [FormatTimestamp](#formattimestamp)     |
+| Time       | [ParseTimestamp](#parsetimestamp)       |
+| Time       | [ParseDateTime](#parsedatetime)         |
```
pradithya, line 17: We also have "statistics"

pradithya:
I love that we started to make better documentation about standard transformer1

api/pkg/transformer/symbol/time.go
```diff
@@ -0,0 +1,240 @@
...
+	timezoneVals := reflect.ValueOf(tz)
+
+	switch timestampVals.Kind() {
+	case reflect.Slice:
+		if timestampVals.Len() == 0 {
+			panic("empty array of timestamp provided")
+		}
+
+		location := &time.Location{}
+		if timezoneVals.Kind() == reflect.Slice {
```
tiopramayudi, line 85: can we move out the validation of timezonevals into separate function, i notice it also be used in `ParseDateTime` function ?

tiopramayudi:
LGTM

pradithya (approved):
LGTM! 🥳

PR #201, opened 2021-12-10
Load location using evaluated args value

ariefrahmansyah:
We should pass the evaluated args value to loadLocation function instead of the interface.


tiopramayudi:
LGTM

tiopramayudi (approved):

PR #202, opened 2021-12-10
Add Go's zoneinfo to standard transformer docker image

ariefrahmansyah:
TIL that Golang needs the host machine to provide zone info. In our standard transformer Docker image, we are using Alpine and it only has a UTC timezone. Hence we need to supply zone info to our image.


pradithya (approved):
LGTM

tiopramayudi (approved):

tiopramayudi:
LGTM

PR #203, opened 2021-12-16
Add default jsonpath capability

tiopramayudi:

Add capability to users to they can specify default value when trying to get `jsonpath` from JSON. This new capability, will affect:
* variable declaration
* entity extraction config
* create table column

Fixes #

Untitled_ Dec 16, 2021 8_40 AM (1)

Screen Shot 2021-12-16 at 11 42 01

Screen Shot 2021-12-16 at 11 42 14

```release-note
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


pradithya:
@tiopramayudi Can you upload video with higher resolution? I am not able to see it in details.

protos/merlin/transformer/spec/common.proto
```diff
@@ -19,4 +19,13 @@ message FromTable {
 message FromJson {
   string jsonPath = 1;
   bool addRowNumber = 2;
+  string defaultValue = 3;
+  ValueType valueType = 4;
+}
+
+enum ValueType {
```
pradithya, line 26: I think we should support array too, right?

api/pkg/transformer/jsonpath/jsonpath_test.go
```diff
@@ -182,6 +187,98 @@ func TestCompiled_Lookup(t *testing.T) {
...
+			},
+			sourceJSONs: jsonContainer,
+			want:        4.4,
+			wantErr:     false,
+			err:         nil,
+		},
+		{
+			desc: "null array with default value",
+			opt: JsonPathOption{
+				JsonPath:     "$.null_array",
```
pradithya, line 244: Do we know if it's an array? It's just empty json field right?

api/pkg/transformer/jsonpath/jsonpath_test.go
```diff
@@ -182,6 +187,98 @@ func TestCompiled_Lookup(t *testing.T) {
...
+			},
+			sourceJSONs: jsonContainer,
+			want:        "default",
+			wantErr:     false,
+			err:         nil,
+		},
+		{
+			desc: "empty array with default value",
+			opt: JsonPathOption{
+				JsonPath:     "$.empty_array",
```
pradithya, line 256: I think the handling is in correct. If it's empty array we should probably return empty slice with the given target type.

api/pkg/transformer/jsonpath/jsonpath_test.go
```diff
@@ -20,6 +21,12 @@ const (
 		  "instances": [
 			{"sepal_length":2.8, "sepal_width":1.0, "petal_length":6.8, "petal_width":0.4},
 			{"sepal_length":0.1, "sepal_width":0.5, "petal_length":1.8, "petal_width":2.4}
+		  ],
+		  "empty_array": [],
```
pradithya, line 25: Can you add test where the array has some null field
```
"instances_with_null": [
			{"sepal_length":2.8, "sepal_width":1.0, "petal_length":6.8, "petal_width":0.4},
			{"sepal_length":0.1, "sepal_width":0.5, "petal_length":1.8, "petal_width":null},
                         {"sepal_length":0.1, "sepal_width":0.5, "petal_length":1.8, "petal_width":0.5}
		  ],
```

and  json path `$.instances_with_null[*].petal_width`

pradithya:
@tiopramayudi Can you add documentation describing its complete behavior? We should make it clear to user how do we handle both nominal and edge-cases.

api/pkg/transformer/jsonpath/jsonpath_test.go
tiopramayudi: oh yeah, it is not necessary an array. This is the case that we face when user specified their array value `null` instead of `[]`, also it can be applied to arbitary json field. I'll update the desc

api/pkg/transformer/jsonpath/jsonpath_test.go
tiopramayudi: can you give the example?

pradithya:
LGTM!

pradithya (approved):

docs/user-guide/standard_transformer.md
```diff
@@ -0,0 +1,789 @@
...
+* Result of jsonpath is array, where some of its value is null
+  ```
+    fromJson:
+        jsonPath: $.instances_with_null[*].petal_width
+        defaultValue: -1
+        valueType: INT
+  ```
+  the result of above Jsonpath is `[0.4,-1,0.5]`, because the original jsonpath result `[0.4,null,0.5]` containing null value, the default value is used to replace `null` value
+
+## Expression
```
ariefrahmansyah, line 131: Please add link to expression documentation.
```
For full list of standard transformer built-in function, please check [Transformer Expressions](./transformer_expressions).```

ariefrahmansyah (approved):
LGTM! Awesome works, Tio! Thanks for standard transformer documentation too!

(I have nit comments)

PR #200, opened 2021-12-01
Direct storage client for bigtable

khorshuheng:
**What this PR does / why we need it**:
This adds the implementation of a Feast bigtable client, which will need to be integrated to the standard transformer later.

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Checklist**

- [ ] Added unit test, integration, and/or e2e tests
- [ ] Tested locally
- [ ] Updated documentation
- [ ] Update Swagger spec if the PR introduce API changes
- [ ] Regenerated Golang and Python client if the PR introduce API changes

PR #204, opened 2022-01-04
Standard Transformer serving directly from Bigtable

tiopramayudi:

This adds the implementation of a Feast bigtable client, and also the integration with merlin API. This PR is continuation of this [PR](https://github.com/gojek/merlin/pull/200)

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/pkg/transformer/feast/redis/client.go
```diff
@@ -0,0 +1,106 @@
...
+		MinIdleConns:       int(option.MinIdleConnections),
+	})
+
+	return RedisClient{
+		encoder:   newRedisEncoder(featureTablesMetadata),
+		pipeliner: redisClient.Pipeline(),
+	}, nil
+}
+
+func (r RedisClient) GetOnlineFeatures(ctx context.Context, req *feast.OnlineFeaturesRequest) (*feast.OnlineFeaturesResponse, error) {
```
pradithya, line 70: Can you add test which involve GetOnlineFeatures? It seems that `client.go` is never tested.

api/pkg/transformer/feast/bigtablestore/client.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, err
+	}
+	return client, nil
+}
+
+// GetOnlineFeatures will first convert the request to RowQuery, which contains all
+// information needed to perform a query to BigTable to retrieve the Feast features,
+// stored in Avro format. The retrieved BigTable rows are then decoded back into
+// the Feast feature values again.
+func (b BigTableClient) GetOnlineFeatures(ctx context.Context, req *feast.OnlineFeaturesRequest) (*feast.OnlineFeaturesResponse, error) {
```
pradithya, line 81: The `client.go` is not tested. Can you add test for it? particularly for `GetOnlineFeatures`

pradithya:
Overall looks good to me, there are several follow up: 
1. Add test for `client.go` (there are 3 client.go that doesn't have test)
2. Documentation of the configuration for Bigtable and Redis. I think we can add it in the /docs folder

pradithya (approved):
LGTM

api/config/feast_storage_config.go
```diff
@@ -43,20 +43,37 @@ func (d *Duration) UnmarshalJSON(b []byte) error {
...
-	MaxRetries           int32     `json:"max_retries"`
-	MinRetryBackoff      *Duration `json:"min_retry_backoff"`
-	DialTimeout          *Duration `json:"dial_timeout"`
-	ReadTimeout          *Duration `json:"read_timeout"`
-	WriteTimeout         *Duration `json:"write_timeout"`
-	MaxConnAge           *Duration `json:"max_conn_age"`
-	PoolTimeout          *Duration `json:"pool_timeout"`
-	IdleTimeout          *Duration `json:"idle_timeout"`
-	IdleCheckFrequency   *Duration `json:"idle_check_frequency"`
+	// Flag to indicate whether redis configuration is for single redis or redis cluster
```
ariefrahmansyah, line 46: Thanks for this comment description!

api/pkg/transformer/feast/bigtablestore/encoding.go
```diff
@@ -0,0 +1,462 @@
...
+func entityKeysToBigTable(project string, entityKeys []*spec.Entity) string {
+	keyNames := make([]string, len(entityKeys))
+	for i, e := range entityKeys {
+		keyNames[i] = e.Name
+	}
+	return project + "__" + strings.Join(keyNames, "__")
+}
+
+func feastValueToStringRepr(val *types.Value) (string, error) {
+	switch v := val.Val.(type) {
```
ariefrahmansyah, line 51: Why is there no float / double value type?

api/pkg/transformer/feast/bigtablestore/encoding.go
```diff
@@ -0,0 +1,462 @@
...
+	return &CachedCodecRegistry{
+		codecs: make(map[string]*goavro.Codec),
+		tables: tables,
+	}
+}
+
+// GetCodec returns an Avro codec, either from the in-memory cache or built
+// from the Avro schema specification.
+func (r *CachedCodecRegistry) GetCodec(ctx context.Context, schemaRef []byte, project string, entityKeys []*spec.Entity) (*goavro.Codec, error) {
+	r.RLock()
```
ariefrahmansyah, line 442: Why do you use RLock/RUnlock here but Lock/Unlock below?

api/pkg/transformer/feast/bigtablestore/storage_test.go
```diff
@@ -0,0 +1,59 @@
+package bigtablestore
```
ariefrahmansyah, line 1: How did you generate this file? Is it possible to add that information in this file too? Similar to what we've done with mockery.

tiopramayudi: it is basically generated by mockery but I moved to the same package, because the interface is not exposed publicly

charts/merlin/templates/merlin-deployment.yaml
```diff
@@ -165,6 +165,8 @@ spec:
           - name: FEAST_BIG_TABLE_CONFIG
             value: {{ .Values.merlin.transformer.feast.bigtableStorage | toJson | quote }}
           {{- end }}
+          - name: FEAST_BIGTABLE_CREDENTIAL
+            value: {{ .Values.merlin.transformer.feast.bigtableCredential}}
```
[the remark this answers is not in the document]
tiopramayudi: tried that before but inferenceservice could not mount volume contains that secret. Hence pass this to env var

api/pkg/transformer/feast/bigtablestore/encoding.go
tiopramayudi: nice catch, I'll add those

tiopramayudi: after discuss with @khorshuheng , it doesn't make sense to have double/float for entity. Also current implementation of feast serving also like this

api/pkg/transformer/feast/bigtablestore/encoding.go
tiopramayudi: using `RLock` enable multiple go routine to read (not write), that's why we use it on the read part. For write part we only enable one goroutine by using `Lock`

charts/merlin/templates/merlin-deployment.yaml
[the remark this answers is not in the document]
ariefrahmansyah: I think it's possible to do that, but instead of mounting the secret, we need to mount the service account that is linked to that secret. Something like this: https://kserve.github.io/website/0.7/modelserving/storage/s3/s3/

Anyway, your implementation is lgtm.

api/pkg/transformer/feast/bigtablestore/encoding.go
ariefrahmansyah: TIL

ariefrahmansyah (approved):

charts/merlin/templates/merlin-deployment.yaml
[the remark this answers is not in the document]
tiopramayudi: I've tried something similar like this https://github.com/kserve/kserve/issues/766, but the pod still use compute service account instead of the one that specified in inferenceservice manifest

PR #205, opened 2022-01-05
Fix yaml specification on feast standard transformer page

tiopramayudi:

Fix issue on how ui render standard transformer yaml spec when user use `jsonPath` instead of `jsonPathConfig`. The issue is happening on :
* variable declaration. Currently it showing `jsonPath` field instead of `jsonPathConfig`
* feast entity. Currently showing both `jsonPath` and `jsonPathConfig` where the expected is only `jsonPathConfig`

Fixes #

Before:
Screen Shot 2022-01-05 at 15 32 06

After:
Screen Shot 2022-01-05 at 16 38 27

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


ariefrahmansyah (approved):
lgtm

PR #208, opened 2022-01-11
Cyclical Encoder + Column Ordering fix

karzuo:
- Cyclical encoder implementation
- UI Support for cyclical encoder
- Fix to column ordering after data modifications (such as encoding)

image

image

Fixes #

```release-note
1. Support for cyclical encoder
2. Fix column ordering issue: Previously columns modified by transformation steps get moved to the front even if the column name remains unchanged. Now, columns ordering or columns modified inplace remain unchanged and new columns created will be appended to the back.
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,312 @@
...
+	8:  UnitDaysInSec31,
+	9:  UnitDaysInSec30,
+	10: UnitDaysInSec31,
+	11: UnitDaysInSec30,
+	12: UnitDaysInSec31,
+	13: UnitDaysInSec29, //Leap year Feb
+}
+
+type CyclicalEncoder struct {
+	Period interface{}
```
pradithya, line 75: Why is `Period` an `interface{}`?

karzuo: So that I can accept a "nil" when we use min max instead... hmm... ok, maybe I can set that to 0

karzuo: oh wait. should be enum.

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// by epoch time
+	byEpochTime := config.GetByEpochTime()
+	var min, max float64 = 0, 0
+	var period spec.PeriodType
+
+	if byEpochTime != nil {
+		switch byEpochTime.Period {
+		case spec.PeriodType_HOUR:
+			period = spec.PeriodType_UNDEFINED
```
tiopramayudi, line 103: why period `UNDEFINED` for HOUR, DAY, WEEK ?

karzuo: They are treated as having fixed period, so we can set the period using min and max range.

api/pkg/transformer/types/encoder/cyclical_encoder_test.go
```diff
@@ -0,0 +1,972 @@
...
+			if tC.expectedErr != nil {
+				assert.EqualError(t, err, tC.expectedErr.Error())
+			} else {
+				assert.Equal(t, tC.expectedEncoder, gotEncoder)
+			}
+		})
+	}
+}
+
+func TestCyclicalEncode(t *testing.T) {
```
tiopramayudi, line 257: can you add test where each input in `reqValues` have at least other input that fall in the same cycle.

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+//Computes the angle in radians represented by per unit second of a pre-defined period
+//This is derived from the formula for calculating phase:
+//phase = time passed / period * 2pi
+//By rearranging the formula (for optimizing computation) into this:
+//phase = time pass * 2pi / period
+//we define unit angle as (2pi / period)
+//The motivation is that we can pre-compute this value once and use it repeatedly.
+func getUnitAngle(periodType interface{}, t time.Time) (float64, error) {
```
tiopramayudi, line 263: why is periodType interface{} ?

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,312 @@
...
+	_, pType := periodType.(spec.PeriodType)
+	if !pType {
+		return 0, fmt.Errorf("invalid type for periodType")
+	}
+
+	switch periodType {
+	case spec.PeriodType_MONTH:
+		dayElapsed := t.Day() - 1
+		hr, min, sec := t.Clock()
+		elapsed := float64(dayElapsed*DayInSec + hr*HourInSec + min*MinInSec + sec)
```
tiopramayudi, line 206: can you make new function for `elapsed` ? it looks like it share the same logic with the one below. Maybe `elapsed(time time.Time, dayElapsed int64, periodElapsed int64)` ?

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,312 @@
...
+		return elapsed, nil
+	case spec.PeriodType_QUARTER:
+		dayElapsed := t.YearDay() - 1
+		hr, min, sec := t.Clock()
+		elapsed := float64(dayElapsed*DayInSec + hr*HourInSec + min*MinInSec + sec)
+
+		if t.Month() <= 3 {
+			return elapsed, nil
+		} else if t.Month() <= 6 {
+			if isLeapYear(t.Year()) {
```
tiopramayudi, line 216: to reduce the branching maybe check the appropriate value for leapYear? e.g
```
        q1Insec := Q1InSec
	h1Insec := H1InSec

	if (isLeapYear(t.Year()) {
		q1InSec = Q1LeapInSec
		h1Insec = H1Insec
	}
```

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	switch periodType {
+	case spec.PeriodType_MONTH:
+		if t.Month() == 2 && isLeapYear(t.Year()) {
+			return MonthInSec[13], nil
+		}
+		return MonthInSec[int(t.Month())], nil
+	case spec.PeriodType_QUARTER:
+		if t.Month() <= 3 {
```
tiopramayudi, line 276: better to store value 3, 6, 9 into variable with proper naming so other understand the meaning of that value

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,312 @@
...
+			}
+			return UnitQ1InSec, nil
+		} else if t.Month() <= 6 {
+			return UnitQ2InSec, nil
+		} else if t.Month() <= 9 {
+			return UnitQ3InSec, nil
+		}
+		return UnitQ4InSec, nil
+	case spec.PeriodType_HALF:
+		if t.Month() <= 6 {
```
tiopramayudi, line 288: same here need to store this to const variable to capture the intention

ui/src/services/transformer/TransformerConfig.js
```diff
@@ -341,6 +341,24 @@ export class Pipeline {
               });
           });
 
+        input.encoders &&
+          input.encoders.forEach(encoder => {
+            if (encoder.cyclicalEncoderConfig) {
```
tiopramayudi, line 346: can it be simplified ? too many nested if

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,316 @@
...
+	UnitH1InSec     = completeAngle / H1InSec
+	UnitH1LeapInSec = completeAngle / H1LeapInSec
+	UnitH2InSec     = completeAngle / H2InSec
+
+	UnitYearInSec     = completeAngle / YearInSec
+	UnitLeapYearInSec = completeAngle / LeapYearInSec
+)
+
+// Unit angles for variable periods for each month
+var MonthInSec = map[int]float64{
```
tiopramayudi, line 65: can you make it private?

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,316 @@
...
+import (
+	"fmt"
+	"github.com/gojek/merlin/pkg/transformer/spec"
+	"github.com/gojek/merlin/pkg/transformer/types/converter"
+	"math"
+	"time"
+)
+
+const (
+	FloatZero   = 0.0000000001
```
tiopramayudi, line 12: if it is not mean to be used by other package, it is better to make it private

tiopramayudi:
can you update existing standard transformer integration test to use this encoder?

tiopramayudi:
I add several comments, but the overall the PR already looking good @karzuo

docs/user-guide/standard_transformer.md
```diff
@@ -386,6 +387,64 @@ See below for a complete example on how to declare an ordinal encoder
...
+
+*By range:*
+```
+-encoders: 
+  -name: wind_dir 
+   cyclicalEncoderConfig:
+      byRange:
+        min: 0
+        max: 360
+```
```
ariefrahmansyah, line 447: Thanks for updating the doc!

Anyway, can you also include the input & output example of the encoder, i.e. for epoch time encoder given a list of date & time what is the epoch output.

ariefrahmansyah:
Currently, there's no validation for the encoder name. Can you add encoderInputSchema validation here: https://github.com/gojek/merlin/blob/main/ui/src/pages/version/components/forms/validation/schema.js#L123

api/pkg/transformer/types/encoder/cyclical_encoder.go
```diff
@@ -0,0 +1,316 @@
...
+	8:  UnitDaysInSec31,
+	9:  UnitDaysInSec30,
+	10: UnitDaysInSec31,
+	11: UnitDaysInSec30,
+	12: UnitDaysInSec31,
+	13: UnitDaysInSec29, // Leap year Feb
+}
+
+type CyclicalEncoder struct {
+	Period spec.PeriodType
```
ariefrahmansyah, line 82: I prefer to name this field `PeriodType`

api/pkg/transformer/types/table/table.go
```diff
@@ -218,26 +218,19 @@ func (t *Table) Sort(sortRules []*spec.SortColumnRule) error {
 
 // UpdateColumnsRaw add or update existing column with values specified in columnValues map
 func (t *Table) UpdateColumnsRaw(columnValues map[string]interface{}) error {
-	origColumns := t.Columns()
+	df := t.DataFrame()
```
ariefrahmansyah, line 221: Can you explain what happens here and why do you change and use t.DataFrame()? Is there any performance gain and benchmark that support it?

karzuo: the behaviour we intended is already available in the function Mutate, which works on the dataframe directly.
It will be redundant and complex to re-implement the same logic ourselves, and from what I see our own logic may be less efficient.

ariefrahmansyah: Can you share benchmarking for this? Something like this: https://github.com/gojek/merlin/blob/main/api/pkg/transformer/symbol/time_bench_test.go#L68-L132

protos/merlin/transformer/spec/encoder.proto
```diff
@@ -10,11 +10,39 @@ message Encoder {
...
+
+message CyclicalEncoderConfig {
+  oneof encodeBy {
+    ByEpochTime byEpochTime = 1;
+    ByRange byRange = 2;
+  }
+}
+
+message ByEpochTime {
+  PeriodType period = 1;
```
ariefrahmansyah, line 31: Can you also change this to `PeriodType periodType`

python/sdk/test/transformer/standard_transformer_no_feast.yaml
```diff
@@ -19,6 +19,10 @@ transformerConfig:
               suv: "1"
               sedan: "2"
               mpv: "3"
+        - name: "daily_cycle"
+          cyclicalEncoderConfig:
+            byEpochTime:
+              period: "DAY"
```
ariefrahmansyah, line 25: Please update it to periodType

docs/user-guide/standard_transformer.md
```diff
@@ -386,6 +387,90 @@ See below for a complete example on how to declare an ordinal encoder
...
+
+See below for complete examples on how to declare a cyclical encoder:
+
+*By epoch time:*
+```
+-encoders: 
+  -name: payday_trend
+   cyclicalEncoderConfig:
+     byEpochTime:
+       period: MONTH
```
ariefrahmansyah, line 436: Please update it to periodType

ariefrahmansyah (approved):
LGTM!

PR #213, opened 2022-02-08
Use pod as label for cAdvisor container metrics

ariefrahmansyah:


pradithya:
I believe these template should be better as a configuration, can you add it to our backlog?[](url)

pradithya (approved):
LGTM

tiopramayudi (approved):

PR #216, opened 2022-02-10
Fixed column update Algo and tests

karzuo:
1. Change update column algorithm in standard transformer to be more efficient fixed problem where columns may not have the correct dimensions.
2. Fixed tests to reflect change to requirements (column ordering after standard transformation)

**Checklist**

- [ x] Added unit test, integration, and/or e2e tests
- [ x] Tested locally
- [ ] Updated documentation


api/pkg/transformer/types/table/table.go
```diff
@@ -218,18 +218,60 @@ func (t *Table) Sort(sortRules []*spec.SortColumnRule) error {
 
 // UpdateColumnsRaw add or update existing column with values specified in columnValues map
 func (t *Table) UpdateColumnsRaw(columnValues map[string]interface{}) error {
-	df := t.DataFrame()
+	origColumns := t.Columns()
+	updateCol := map[string]bool{} //name of col that are updates
+	combinedColumns := make([]*series.Series, 0)
+
+	columnValues = broadcastScalar(columnValues, t.NRow())
```
pradithya, line 225: Why perform `broadcastScalar` ?

karzuo: It is to make all columns have the same number of rows, like how it is done in createSeries function below. Otherwise may have dimensions different error.

api/pkg/transformer/types/table/table.go
```diff
@@ -218,18 +218,60 @@ func (t *Table) Sort(sortRules []*spec.SortColumnRule) error {
...
+
+	//Check through all original columns for updates
+	for _, origColumn := range origColumns {
```
pradithya, line 228: nit: I think this part can be simplified a bit

```
	// Check through all original columns for updates
	for _, origColumn := range origColumns {
		origColumnName := origColumn.Series().Name
		val, update := columnValues[origColumnName]

		updatedColumn := origColumn
		if update {
			// Update original column
			colSeries, err := series.NewInferType(val, origColumnName)
			if err != nil {
				return err
			}
			updatedColumn = colSeries

			// Record down columns that are updates
			updateCol[origColumnName] = true
		}

		// Add column, either the origColumn or the updated one
		combinedColumns = append(combinedColumns, updatedColumn)
	}
```

In general, try to have less `else` statement.

pradithya (approved):
LGTM

PR #218, opened 2022-02-11
Fix int test

karzuo:
Fix input data of integration test (conversion error due to timezone)

**Checklist**

- [ x] Added unit test, integration, and/or e2e tests
- [x ] Tested locally
- [ ] Updated documentation


python/sdk/test/integration_test.py
```diff
@@ -706,9 +706,9 @@ def test_standard_transformer_without_feast(
     request_json = {
         "drivers": [
             # 1 Feb 2022, 18:00:00 
-            {"id": 1, "name": "driver-1", "vehicle": "motorcycle", "previous_vehicle": "suv","rating": 4, "ep_time": 1643709600}, 
+            {"id": 1, "name": "driver-1", "vehicle": "motorcycle", "previous_vehicle": "suv","rating": 4, "ep_time": 1643738400},
```
pradithya, line 709: Why do you need to change this?

pradithya (approved):

PR #222, opened 2022-03-01
Use golang 1.17

pradithya:

Some package in https://github.com/gojek/merlin/pull/221 requires newer version of golang (i.e. 1.15) . This PR is to update golang version in CI, docker, and development environment to 1.17

None

None

**Checklist**

- [x] Tested locally
- [x] Updated documentation

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -1,15 +1,11 @@
 module github.com/gojek/merlin
 
-go 1.13
+go 1.17
 
 require (
 	cloud.google.com/go/bigtable v1.11.0
-	cloud.google.com/go/container v1.0.0 // indirect
-	cloud.google.com/go/monitoring v1.1.0 // indirect
-	cloud.google.com/go/trace v1.0.0 // indirect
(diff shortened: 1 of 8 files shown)


api/go.mod
```diff
@@ -81,6 +73,121 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
+require (
```
ariefrahmansyah, line 76: Why is there two require blocks?

ariefrahmansyah: You might need to run `go mod tidy`

pradithya: It's because in 1.17 they implements https://go-review.googlesource.com/c/go/+/325922/

ariefrahmansyah: I see. #TIL

ariefrahmansyah:
LGTM.

However, you need to update Makefile for remaining test recipes: https://github.com/gojek/merlin/blob/main/Makefile#L71-L81
You need to add `-gcflags=all=-d=checkptr=0` flag.

ariefrahmansyah (approved):

PR #223, opened 2022-03-01
Fix mapping of entity names to bigtable

khorshuheng:
Signed-off-by: Khor Shu Heng <khor.heng@gojek.com>

Currently, standard transformer uses the entities provided by the users to infer the bigtable storage location. If the order of entities is different from the order of entities in the feature spec, the storage location would be wrong.

This PR fixed this by using the entities retrieve from the feature spec instead.

Fixes #

```release-note
None
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


pradithya (approved):
LGTM

api/pkg/transformer/feast/bigtablestore/encoding_test.go
```diff
@@ -1015,7 +1015,7 @@ func TestEntityKeysToBigTable(t *testing.T) {
 	}
 	for _, tC := range testCases {
 		t.Run(tC.desc, func(t *testing.T) {
-			got := entityKeysToBigTable(tC.project, tC.entityKeys)
+			got := entityKeysToBigTable(tC.project, entitiesToEntityNames(tC.entityKeys))
```
tiopramayudi, line 1018: can you add test where the entities supplied by user doesn't have same order with the one from feature table response?

tiopramayudi:
LGTM

PR #225, opened 2022-03-03
Initialize inference service's annotations

ariefrahmansyah:

Previously, in #224, when there's no queueResourcePercentage in environment's configuration, we won't add `"queue.sidecar.serving.knative.dev/resourcePercentage"` in inference service's annotation. On doing this, we are not initializing annotation field on inference service object, leading to panic when another value expect this annotation field to be exist.

Fixes deploying pyfunc mode without queue resource percentage.

**Checklist**

- [x] Tested locally

### DIFF
--- a/api/cluster/controller_test.go
+++ b/api/cluster/controller_test.go
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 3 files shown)


pradithya (approved):
LGTM

PR #226, opened 2022-03-03
Support convertion from []interface to list type.

ariefrahmansyah:

Given a first successful standard transformer request that has feast feature retrieval with list types:

```
➜ curl -X POST http://localhost:8081/v1/models/echo:predict -d '{"driver_id":1}'
{"merlin":[{"merlin_test_driver_features:test_double":3.14,"merlin_test_driver_features:test_double_list":[1.23,2.34,3.45],"merlin_test_driver_features:test_int64":null,"merlin_test_driver_features:test_int64_list":null,"merlin_test_driver_features:test_string":null,"merlin_test_driver_features:test_string_list":["AAABBBCCC"],"merlin_test_driver_id":"1"}]}
```

And feast cache is enabled, and we send the same request, it will return error:

```
➜ curl -X POST http://localhost:8081/v1/models/echo:predict -d '{"driver_id":1}'
{"code":500,"message":"preprocessing error: error executing preprocessing operation: *pipeline.FeastOp: unsupported conversion from []interface {} to []float64"}
```

This is because, on the second request, standard transformer uses cached feature value which stored as []interface instead of the correct Feast's list type.

**Checklist**

- [x] Tested locally

### DIFF
--- a/api/pkg/transformer/feast/feature_cache.go
+++ b/api/pkg/transformer/feast/feature_cache.go
@@ -145,7 +145,7 @@ func (fc *featureCache) insertFeaturesOfEntity(entity feast.Row, columnNames []s
 
 func castValueRow(row types.ValueRow, columnTypes []feastTypes.ValueType_Enum) (types.ValueRow, error) {
 	for idx, val := range row {
-		castedVal, err := castIntValue(val, columnTypes[idx])
+		castedVal, err := castValue(val, columnTypes[idx])
 		if err != nil {
 			return row, err
 		}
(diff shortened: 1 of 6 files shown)


pradithya:
Can you add test that reproduce the issue, those new unit tests seems didn't cover the bug.

pradithya:
Can you also handle it in feature cache (https://github.com/gojek/merlin/blob/a382020bfa5f660467103a21d68b12b78c8f47ef/api/pkg/transformer/feast/feature_cache.go#L159) so that anything return by cache is already correct.

pradithya:
Awesome! LGTM!

tiopramayudi:
LGTM

tiopramayudi (approved):

PR #228, opened 2022-03-10
Column wise operator, filter and slice row

tiopramayudi:

In this PR, we add new capabilities in standard transformer:
* Conditional Update
* Column wise operator
* Filter row
* Slice row

Fixes #

https://user-images.githubusercontent.com/2369255/161458901-ce538b34-7e28-49bb-b46a-bc82e064f5a4.mov

```release-note

```

**Checklist**

- [X] Tested locally
- [X] Updated documentation


pradithya:
Can you rebase?

pradithya:
Can you check why the CI is failing? What's the video resolution? It seems very blurry, I am not sure if github perform compression but can you set the resolution to at least 720p?

api/pkg/transformer/pipeline/testdata/valid_table_transform_conditional_filtering.yaml
```diff
@@ -0,0 +1,46 @@
...
+          inputTable: driver_table
+          outputTable: transformed_driver_table
+          steps:
+            - updateColumns:
+              - column: customer_id
+                expression: customer_id
+              - column: driver_performa
+                conditions:
+                  - if: driver_table.Col("rating") * 2 <= 7
+                    expression: driver_table.Col("rating") * 1
```
pradithya, line 26: So this expression will only apply to rows that satisfy `driver_table.Col("rating") * 2 <= 7` condition, right? I now feel calling it `if` feel strange. Maybe `toRowsIf` sounds better?

pradithya: or `rowSelector` ?

tiopramayudi: `rowSelector` sounds better

api/pkg/transformer/pipeline/util.go
```diff
@@ -16,6 +18,51 @@ func evalJSONPath(env *Environment, jsonPath string) (interface{}, error) {
...
 func evalExpression(env *Environment, expression string) (interface{}, error) {
+	val, err := getVal(env, expression)
+	if err != nil {
+		return nil, err
+	}
+	switch exprVal := val.(type) {
+	case *operation.OperationNode:
+		val, err = exprVal.Execute()
+	case operation.OperationNode:
+		val, err = exprVal.Execute()
```
pradithya, line 29: nit: Can it do it like this?

```go
case  *operation.OperationNode, operation.OperationNode:
    val, err = exprVal.Execute()
```

api/pkg/transformer/pipeline/util.go
```diff
@@ -16,6 +18,51 @@ func evalJSONPath(env *Environment, jsonPath string) (interface{}, error) {
...
+	}
```
pradithya, line 30: What happen with the `default` case?

api/pkg/transformer/pipeline/util.go
```diff
@@ -16,6 +18,51 @@ func evalJSONPath(env *Environment, jsonPath string) (interface{}, error) {
 }
 
 func evalExpression(env *Environment, expression string) (interface{}, error) {
+	val, err := getVal(env, expression)
+	if err != nil {
+		return nil, err
+	}
+	switch exprVal := val.(type) {
+	case *operation.OperationNode:
+		val, err = exprVal.Execute()
```
pradithya, line 27: Is there any compilation penalty with this?

api/pkg/transformer/symbol/operator.go
```diff
@@ -0,0 +1,57 @@
+package symbol
+
+import (
+	"github.com/gojek/merlin/pkg/transformer/types/operation"
+)
+
+func (sr Registry) GreaterOp(left, right interface{}) interface{} {
```
pradithya, line 7: Can you add doc for each method?

api/pkg/transformer/symbol/registry.go
```diff
@@ -81,6 +82,18 @@ func (sr Registry) evalArg(arg interface{}) (interface{}, error) {
 		return cplJsonPath.LookupFromContainer(sr.jsonObjectContainer())
 	case *series.Series:
 		return val.GetRecords(), nil
+	case *operation.OperationNode:
```
pradithya, line 85: nit: 
```go
case **operation.OperationNode, operation.OperationNode:
```

api/pkg/transformer/symbol/operator.go
pradithya, line 7: Do we have test for methods in operator.go?

api/pkg/transformer/types/operation/operation.go
```diff
@@ -0,0 +1,194 @@
+package operation
+
+import (
+	"fmt"
+
+	"github.com/gojek/merlin/pkg/transformer/types/series"
+)
+
```
pradithya, line 8: Can you add documentation too all exported symbols?

api/pkg/transformer/types/table/table_test.go
```diff
@@ -771,3 +770,299 @@ func TestTable_NewRaw(t *testing.T) {
...
+			wantErr:    true,
+			errMessage: "failed slice col: A due to: slice index out of bounds",
+		},
+		{
+			name: "start < 0",
+			inputTable: New(
+				series.New([]int{1, 2, 3, 4, 5}, series.Int, "A"),
+				series.New([]string{"a", "b", "c", "d", "e"}, series.String, "B"),
+			),
+			startIdx:   -1,
```
pradithya, line 838: Can we allow negative index? It'll be useful for selecting series entry without having to know the series length. E.g.
```python
>>> a = [1,2,3,4,5,6,7,8,9,10]
# return last item
>>> a[-1]
10

# return 2nd last item
>>> a[-2]
9

# return last 2 items
>>> a[-2:]
[9, 10]
```

docs/user-guide/standard_transformer.md
```diff
@@ -577,6 +577,85 @@ tableTransformation:
...
+        - column: "col2"
+          conditions:
+          - if: table1.Col('col1') * 2 > 10
+            expression: table1.Col('col1')
+          - default:
+              expression: -1
+```
+
+There are two ways to update columns:
+* Update column to the whole rows. Users need to only specify `column` and `expression`. Evaluation from that `expression` must contains of one value or array / series that has the same length with the rest of the column. Following the example:
```
pradithya, line 589: Rephares

```text
Update all rows in the column. You need to specify `column`and `expression`. `column` determines which column to be updated and `expression` determines the value that will be used to update the column.
Values produced by the `expression` must be a scalar or a series that has the same length as the other columns. Following the example:
```

docs/user-guide/standard_transformer.md
```diff
@@ -577,6 +577,85 @@ tableTransformation:
...
+  ```
+    - updateColumns:
+      - column: "customer_id"
+        expression: "cust_1" # the value is scalar and will be broadcasted to all the row
+      - column: "s2id"
+        expression: S2ID(table1.Col('lat'), table2.Col('lon'), 12) # the value is array or series that the length should be the same with the rest of the columns in a table 
+  ```
+* Update column to the subset of rows based on conditions. For this users can set multiple `if` with `expression` and also default value if none of conditions are match. For example users have following table
```
pradithya, line 597: Rephrase

```text
Update subset of rows in the columns given some row selector condition. 
```

docs/user-guide/standard_transformer.md
```diff
@@ -577,6 +577,85 @@ tableTransformation:
...
+tableTransformation:
+     inputTable: myTable
+     outputTable: myTransformedTable
+     steps:
+     - filterRow:
+        condition: myTable.Col('total_booking_1w') < 5
+```
+
+### Slice Row
+Slice row is an operation to slice a table based on start(lower bound) and end index(upper bound) that given by the user. The result will including start index but not end index. Below is the example of this operation
```
pradithya, line 650: The result includes starting index but excluding end index.

ui/src/pages/version/components/forms/components/transformer/components/table_operations/UpdateColumnPanel.js
```diff
@@ -0,0 +1,147 @@
...
+              description="Add column to updated or created"
+              onClick={() => onAddColumn()}
+            />
+          </EuiFlexItem>
+        </EuiFlexGroup>
+      </EuiDragDropContext>
+    </EuiPanel>
+  );
+
+  //   const tableColumns = [
```
pradithya, line 91: Can we remove?

api/pkg/transformer/types/table/table_test.go
tiopramayudi: can, I'll update this

api/pkg/transformer/pipeline/testdata/valid_table_transform_conditional_filtering.yaml
karzuo: sounds good. But `conditions` a bit not clear. Maybe call it `selectiveTransform` ?

api/pkg/transformer/pipeline/util.go
tiopramayudi: with the above `exprVal` type will be `interface` and can't call `Execute`

api/pkg/transformer/types/table/table_test.go
tiopramayudi: # return 2nd last item
>>> a[-2]
what is the end index in this case ya?
>>> a[-2:]
for this case the end index is nil

api/pkg/transformer/pipeline/util.go
tiopramayudi: what do you mean by `compilation penalty`?

pradithya: Nevermind, I thought `exprVal.Execute()` is calling expr `Execute`

ui/src/pages/version/components/forms/components/transformer/components/table_operations/SelectUpdateColumnStrategy.js
```diff
@@ -0,0 +1,56 @@
...
+    strategy !== "" ? option.value === strategy : option.value === ""
+  );
+
+  return (
+    <EuiForm>
+      <EuiFormRow
+        fullWidth
+        label={
+          <FormLabelWithToolTip
+            label="Choose Strategy *"
```
pradithya, line 37: I think we can drop `Choose`.

api/pkg/transformer/types/table/table_test.go
pradithya: I think you should mimic Python behavior

pradithya (approved):
LGTM!

PR #231, opened 2022-03-18
Refactor how to record feast metrics.

ariefrahmansyah:

Fixes not able to log NULL feature status if user didn't specify default value.

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/pkg/transformer/feast/call.go
```diff
@@ -82,6 +82,11 @@ func (fc *call) processResponse(feastResponse *feast.OnlineFeaturesResponse) (*i
 			var rawValue *types.Value
 
 			featureStatus := responseStatus[rowIdx][column]
+			// put behind feature toggle since it will generate high cardinality metrics
+			if fc.statusMonitoringEnabled {
```
pradithya, line 86: I don't understand why putting it here helps?

ariefrahmansyah: These lines will skip recordMetrics() function: https://github.com/gojek/merlin/blob/5d786310fc9062781ed56710f8570f1e67dfd248/api/pkg/transformer/feast/call.go#L100-L103

So currently, if feature X is NULL, NOT FOUND, or etc, standard transformer tries to find the default value. But if default value is not exist, we will skip it entirely and return it as is without recording the metrics.

pradithya: Can we refactor that part instead? So it shouldn't exit the loop. `processResponse` is already a complex method, I was trying to reduce the noise by extracting metrics recording to `fc.recordMetrics(val, column, featureStatus)`

pradithya:
@ariefrahmansyah Are you going to continue this PR?

PR #234, opened 2022-03-31
Create tables from file & UI bugs fix

karzuo:
Feature:
- Create table from CSV / Parquet files stored in local (with model) or gcs bucket
- UI for creating table from file
image
image

Bug fixes:
- Create table from JSON missing "add columns"
- Column type for create table has unnecessary type, it should only have json and expression
- switching between json and expression goes haywire in create table, feast table and variable
image

```release-note
- Allow user to create table from CSV/parquet file
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


protos/merlin/transformer/spec/common.proto
```diff
@@ -23,6 +23,27 @@ message FromJson {
   ValueType valueType = 4;
 }
 
+message FromFile {
```
pradithya, line 26: Can we put `uri` at the top?
```
message FromFile {
  enum FileType {
    CSV = 0;
    PARQUET = 1;
  }
  string uri = 1;
  FileType format = 2;
  
  repeated Schema schema = 3;
}

```

protos/merlin/transformer/spec/common.proto
```diff
@@ -23,6 +23,27 @@ message FromJson {
...
+  enum FileType {
+    CSV = 0;
+    PARQUET = 1;
+  }
+  FileType format = 1;
+  string uri = 2;
```
pradithya, line 32: Also since we use `uri` term here it should follow correct [convention](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#:~:text=The%20URI%20comprises%3A,%2C%20or%20hyphen%20(%20%2D%20).).
That means we can't use relative path (which we need for use case of accessing csv in model artifact).

As such, I think we should differentiate the field. So:

```
message FromFile {
  enum FileType {
    CSV = 0;
    PARQUET = 1;
  }

  // either uri or artifactPath must be populated
  string uri = 1; // e.g. gs://my-bucket/my-file.csv or file:///tmp/absolute/path/my-file.csv
  string artifactPath = 2; // e.g. path/my-file.csv --> standard transformer will translate it to file:///mnt/models/artifact/path/my-file.csv

  FileType format = 3;
  
  repeated Schema schema = 4;
}
```

api/pkg/transformer/pipeline/compiler.go
```diff
@@ -17,6 +17,11 @@ import (
 	"github.com/gojek/merlin/pkg/transformer/types/table"
 )
 
+const (
+	envPredictorStorageURI = "STORAGE_URI"
+	artifactsFolder        = "artifacts"
```
pradithya, line 22: `artifactsFolder` should be `/mnt/models/artifacts/`

api/pkg/transformer/pipeline/compiler.go
```diff
@@ -17,6 +17,11 @@ import (
 	"github.com/gojek/merlin/pkg/transformer/types/table"
 )
 
+const (
+	envPredictorStorageURI = "STORAGE_URI"
```
pradithya, line 21: you don't need this `envPredictorStorageURI`

api/pkg/transformer/pipeline/compiler.go
```diff
@@ -212,9 +237,30 @@ func (c *Compiler) parseTablesSpec(tableSpecs []*spec.Table, compiledJsonPaths *
...
 				}
 				compiledJsonPaths.Set(bt.FromJson.JsonPath, compiledJsonPath)
+			case *spec.BaseTable_FromFile:
+				var records [][]string
+				var err error
+
+				if tableSpec.BaseTable.GetFromFile().GetFormat() == spec.FromFile_CSV {
+					records, err = table.RecordsFromCsv(tableSpec.BaseTable.GetFromFile().GetUri())
+				} else if tableSpec.BaseTable.GetFromFile().GetFormat() == spec.FromFile_PARQUET {
+					prefix := os.Getenv(envPredictorStorageURI)
```
pradithya, line 250: The `prefix` is `artifactsFolder`

api/pkg/transformer/pipeline/compiler.go
```diff
@@ -212,9 +237,30 @@ func (c *Compiler) parseTablesSpec(tableSpecs []*spec.Table, compiledJsonPaths *
...
 				if err != nil {
-					return nil, err
+					return nil, nil, err
 				}
 				compiledJsonPaths.Set(bt.FromJson.JsonPath, compiledJsonPath)
+			case *spec.BaseTable_FromFile:
+				var records [][]string
+				var err error
+
+				if tableSpec.BaseTable.GetFromFile().GetFormat() == spec.FromFile_CSV {
```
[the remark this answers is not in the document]
karzuo: overlooked. sorry

api/cluster/templater.go
```diff
@@ -275,6 +275,7 @@ func (t *KFServingResourceTemplater) createTransformerSpec(modelService *models.
 	envVars = append(envVars, models.EnvVar{Name: envTransformerPort, Value: defaultTransformerPort})
 	envVars = append(envVars, models.EnvVar{Name: envTransformerModelName, Value: modelService.Name})
 	envVars = append(envVars, models.EnvVar{Name: envTransformerPredictURL, Value: createPredictURL(modelService)})
+	envVars = append(envVars, models.EnvVar{Name: envPredictorStorageURI, Value: utils.CreateModelLocation(modelService.ArtifactURI)})
```
pradithya, line 278: @tiopramayudi This should be enough to mount model artifact in transformer right?

api/pkg/transformer/types/table/output.csv
```diff
@@ -0,0 +1,6 @@
+First Name,Last Name,Age,Weight,Is VIP
```
pradithya, line 1: Do we need this file here?

api/pkg/transformer/types/table/table.go
```diff
@@ -40,6 +52,162 @@ func NewRaw(columnValues map[string]interface{}) (*Table, error) {
...
+
+	return records, nil
+}
+
+// NewFromRecords create a new table from records in the form [][]string
+// The first row of the records array shall contain all the column names.
+// This function will check that there are at least 2 rows in records (header and data).
+// It will also check that the number of columns and the column names matches the schema specified,
+// before creating a new table to be returned
+func NewFromRecords(records [][]string, schema []*spec.Schema) (*Table, error) {
```
pradithya, line 164: very neat!!

api/pkg/transformer/types/table/table.go
```diff
@@ -40,6 +52,162 @@ func NewRaw(columnValues map[string]interface{}) (*Table, error) {
...
+	if len(records) <= 1 {
+		return nil, fmt.Errorf("no data found")
+	}
+	// check no. of columns same as defined in schema
+	header := make(map[string]bool)
+	for _, colName := range records[0] {
+		header[colName] = true
+	}
+
+	if len(header) != len(schema) {
```
pradithya, line 176: I think we shouldn't enforce this. We should make `schema` as optional input and it's only used to override our inferred type.

pradithya:
Can you post the UI screenshot and recording

api/pkg/transformer/types/table/output.csv
karzuo: no need.

protos/merlin/transformer/spec/common.proto
karzuo: I prefer to stick to just 1 variable, say `filePath`. it is not that difficult to distinguish between a full and a relative path. However, if you give too many variables to the user, they will get confused and end up using wrongly.

pradithya: It's general purpose uri parsing https://go.dev/play/p/5PY9lpmEIkU

api/cluster/templater.go
tiopramayudi: transformer container name also must be changed to `kfserving-container`

pradithya: This will break logging isn't it? @tiopramayudi

api/pkg/transformer/pipeline/create_table_op.go
```diff
@@ -48,7 +48,7 @@ func (c CreateTableOp) Execute(ctx context.Context, env *Environment) error {
 	return nil
 }
 
-func createBaseTable(env *Environment, baseTableSpec *spec.BaseTable) (*table.Table, error) {
+func createBaseTable(env *Environment, baseTableSpec *spec.BaseTable, tableName string) (*table.Table, error) {
```
tiopramayudi, line 51: this `tableName` is not used

api/pkg/transformer/pipeline/environment.go
```diff
@@ -28,6 +28,11 @@ func NewEnvironment(compiledPipeline *CompiledPipeline, logger *zap.Logger) *Env
 		logger:           logger,
 	}
 
+	// attach pre-loaded tables to environment
+	for k, v := range compiledPipeline.preloadedTables {
```
tiopramayudi, line 32: there is a chance that the preloadtables has the same name with the other tables, how do we handle this?

karzuo: It should throw the same error as the normal way to create tables?

api/pkg/transformer/pipeline/testdata/valid_input_only.yaml
```diff
@@ -38,6 +38,22 @@ transformerConfig:
               - name: column_2
                 fromJson:
                   jsonPath: $.entity_2
+          - name: preload_1
+            baseTable:
+              fromFile:
+                format: PARQUET
+                uri: ../types/table/testdata/normal.parquet
+                schema:
```
pradithya, line 46: Can you add test that doesn't have `schema`?

pradithya, line 46: Can you add test that doesn't have `schema`?

api/pkg/transformer/types/table/table.go
```diff
@@ -40,6 +51,159 @@ func NewRaw(columnValues map[string]interface{}) (*Table, error) {
...
+		defer file.Close()
+	}
+
+	fr, err := goparquet.NewFileReader(r)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get header
+	schema := fr.GetSchemaDefinition()
```
pradithya, line 86: I think you should also return the column type from schema and make use of it when creating the table. We should use this information for inferring column type instead of gota type inference which might be incorrect in some cases.

pradithya, line 86: I think you should also return the column type from schema and make use of it when creating the table. We should use this information for inferring column type instead of gota type inference which might be incorrect in some cases.

api/pkg/transformer/pipeline/environment.go
pradithya: Shall we create warning logs if there is duplicated name?

pradithya: Shall we create warning logs if there is duplicated name?

pradithya:
Can you create separate PR for the bug fix? We should not merge multiple changes into same PR
```
Bug fixes:

Create table from JSON missing "add columns"
Column type for create table has unnecessary type, it should only have json and expression
switching between json and expression goes haywire in create table, feast table and variable
```

pradithya:
Can you create separate PR for the bug fix? We should not merge multiple changes into same PR
```
Bug fixes:

Create table from JSON missing "add columns"
Column type for create table has unnecessary type, it should only have json and expression
switching between json and expression goes haywire in create table, feast table and variable
```

karzuo:
probably too late to separate out. Let's not create more bugs for now. Will take note for next round

api/pkg/transformer/pipeline/environment.go
tiopramayudi: i thought `preloadedTables` is table that created only from file, not the table that use `fromFile` `baseTable`. With current approach if there are two table creation that use same file as baseFile it will download and parse multiple time ?

pradithya (approved):
LGTM! Can you fix the conflict before merging?

PR #236, opened 2022-04-06
Fix insert cache with NaN value

tiopramayudi:

There is issue when insert cache that contains `NaN` value. To solve this, we change all `NaN` to nil because it will be converted to nil eventually

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/transformer/types/converter/converter.go
+++ b/api/pkg/transformer/types/converter/converter.go
@@ -1008,7 +1008,6 @@ func extractArrayFromString(str interface{}, delimiter string) (interface{}, err
 }
 
 func ExtractFeastValue(val *feastType.Value) (interface{}, feastType.ValueType_Enum, error) {
-	fmt.Printf("val: %+v and type %T\n", val, val)
 	switch val.Val.(type) {
 	case *feastType.Value_StringVal:
 		return val.GetStringVal(), feastType.ValueType_STRING, nil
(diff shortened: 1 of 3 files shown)


pradithya:
Can we add test that involve caching?

api/pkg/transformer/types/converter/converter.go
```diff
@@ -500,14 +501,24 @@ func ToFloat32List(val interface{}) ([]float32, error) {
 		}
 		return out, nil
 	case float32, float64:
-		val := reflect.ValueOf(v)
-		return []float32{float32(val.Float())}, nil
+		d, err := ToFloat32(val)
```
pradithya, line 504: I feel it's incorrect. NaN and nil is 2 different values. In some cases user/client library might have different handling for those.  Ideally, instead of changing how we convert NaN, we should change our cache mechanism to support NaN.

tiopramayudi: eventually all `NaN` value will converted to `null` in our case (gota handle `NaN` as nil), that's why I modify it to `null`.  I already did 2 approaches:
* using gob instead of `json` encoding, it is surprisingly slower
* modify `NaN` (float64) value in feast row values to string before inserting to cache. It is work well but there is small latency increase.

tiopramayudi: so the response from API is still the same using `NaN` or null

pradithya: Ok got it.

pradithya (approved):
LGTM

PR #240, opened 2022-04-12
Update CI for automating package publishing

pradithya:
This PR will add automation for publishing docker image of all merlin components when certain action is performed. Below is the list of action and the package version that will be generated
1. a pull request is created. The version will be `0.0.0-{commit hash of the PR}`
2. a tag/release is created with prefix `v`. The version will be the tag value without `v`. E.g. if tag is `v1.2.3` then the version is `1.2.3`
3. a merge to `main` branch is performed. The version will use value of `git describe --tags --always --first-parent`, which will produce version string such as `0.19.0-21-g28ca755`

The PR also contains some improvement in end to end test

No package release

### DIFF
--- a/.github/workflows/merlin.yml
+++ b/.github/workflows/merlin.yml
@@ -26,9 +26,9 @@ jobs:
         name: Create version string
         run: |
           # This version will be used to identify intermediary artifact published to github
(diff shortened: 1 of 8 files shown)


ariefrahmansyah (approved):
LGTM

PR #241, opened 2022-04-14
Fix issue with built-in function

tiopramayudi:

We faced issue when dealing with expr built function, such as `filter`, `map`, `none`, `all`, and `any`. This is because we override arithmetic and comparator operation. The override operation will register the operation and evaluate later, this is due to subset operation for series, but for operation between primitive types (int, bool, string, float) we can evaluate the value immediately. Hence the fix, will evaluate the value of operation between primitive types immediately

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


pradithya:
Can you add test that will catch this issue in https://github.com/gojek/merlin/blob/d5f9a3b63fcc7e84d1bd3a43e9eda93f68b53460/api/pkg/transformer/server/server_test.go#L146 . This test cover standard transformer behavior almost end to end, so I would like to use it to minimise number of end to end test / manual test.

api/pkg/transformer/symbol/operator.go
```diff
@@ -2,95 +2,118 @@ package symbol
...
+	}
+
+	val, err := op.Execute()
+	if err != nil {
+		panic(err)
+	}
+	return val
+}
+
+func evalImmediately(v interface{}) bool {
```
pradithya, line 112: rename to `shouldEagerlyEvaluate`

api/pkg/transformer/symbol/operator.go
```diff
@@ -2,95 +2,118 @@ package symbol
 
 import (
 	"github.com/gojek/merlin/pkg/transformer/types/operation"
+	"github.com/gojek/merlin/pkg/transformer/types/series"
 )
 
 // GreaterOp is function that override default '>' operator that originally only applicable to numeric type
 // This override method enable user to do operation '>' to series
-// Returning chain of operation that needs to be executed later
+// Returning chain of operation that can be executed later or right
```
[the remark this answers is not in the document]
tiopramayudi: updated

api/pkg/transformer/symbol/operator.go
tiopramayudi: updated

tiopramayudi:
> Can you add test that will catch this issue in
> 
> https://github.com/gojek/merlin/blob/d5f9a3b63fcc7e84d1bd3a43e9eda93f68b53460/api/pkg/transformer/server/server_test.go#L146
> 
> . This test cover standard transformer behavior almost end to end, so I would like to use it to minimise number of end to end test / manual test.

done

pradithya (approved):
LGTM

PR #247, opened 2022-04-26
Add variable operation in transformations step

tiopramayudi:

Adding variable operation in transformations step. The purpose of this is to reduce the computation if somehow a user tries to use modification of outputTable from table transformation or table join

Fixes #

https://user-images.githubusercontent.com/2369255/165221140-c4373fdf-467c-4137-ba82-08365c231309.mov

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/pkg/transformer/pipeline/compiler.go
```diff
@@ -172,14 +180,14 @@ func (c *Compiler) parseVariablesSpec(variables []*spec.Variable, compiledJsonPa
 		case *spec.Variable_Expression:
 			compiledExpression, err := c.compileExpression(v.Expression)
 			if err != nil {
-				return nil, nil
+				return nil, err
```
ariefrahmansyah, line 183: Ah, was it a bug if we did not return error?

ariefrahmansyah (approved):
LGTM!

api/pkg/transformer/pipeline/compiler.go
tiopramayudi: yes, just realised this when adding new test

PR #248, opened 2022-04-26
Update google cloud dependencies on merlin sdk

tiopramayudi:

Current merlin sdk has tight contraint on their dependency version, it might cause conflict with other dependencies that are used alongside merlin-sdk. This PR relax the dependency in merlin sdk

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/examples/pyfunc/env.yaml
+++ b/examples/pyfunc/env.yaml
@@ -1,5 +1,5 @@
 dependencies:
-  - python=3.9
+  - python=3.7
   - pip:
     - joblib
     - numpy
--- a/examples/pyfunc/env.yaml
+++ b/examples/pyfunc/env.yaml
@@ -1,5 +1,5 @@
 dependencies:
-  - python=3.7
+  - python=3.9
   - pip:
     - joblib
     - numpy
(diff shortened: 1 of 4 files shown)


pradithya:
Do we see any risk of breaking existing batch model?
Do we need to regenerate the proto definition for batch predictor? I remember you need to regenerate it in https://github.com/gojek/merlin/pull/47

tiopramayudi:
> Do we see any risk of breaking existing batch model?

There is a risk if user use older version of `grpcio` e.g 1.2.2, to solve this user needs to pin older merlin-sdk version (not the latest version)

>  Do we need to regenerate the proto definition for batch predictor? I remember you need to regenerate it in #47
I think we can regenerate the proto with the same risk like the first point. I don't regenerate the proto because currently it didn't block anything

pradithya:
> There is a risk if user use older version of grpcio e.g 1.2.2, to solve this user needs to pin older merlin-sdk version (not the latest version)

Pin merlin-sdk? Isn't it embedded in the batch predictor image? Is it better if they update their dependency?

>  I don't regenerate the proto because currently it didn't block anything

I see, I think we don't need to regenerate since it's forward compatible correct?

tiopramayudi:
> Pin merlin-sdk? Isn't it embedded in the batch predictor image? Is it better if they update their dependency?

Yes, it is better for them to update the dependency. If user hesitant to update their dependency they can pin merlin-sdk version on their conda yaml

pradithya (approved):
LGTM

PR #253, opened 2022-06-02
Standard Transformer Simulation

tiopramayudi:

Add simulation for standard transformer, this is including changes of UI to reflect the new API

Fixes #

https://user-images.githubusercontent.com/2369255/171762171-95b1d464-89e2-4861-bc7a-5b776986a73f.mov

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/pkg/transformer/pipeline/operation.go
```diff
@@ -1,7 +1,118 @@
...
+func NewOperationTracing(operationSpecs interface{}, opType types.OperationType) *OperationTracing {
+	opTracing := &OperationTracing{
+		Specs:  operationSpecs,
+		OpType: opType,
+	}
+	return opTracing
+}
+
+func (ot *OperationTracing) AddInputOutput(input, output map[string]interface{}) error {
+	ot.mutex.Lock()
```
pradithya, line 42: Why is there need for mutex? Is there case that `AddInputOutput` is called concurrently?

api/config/config_test.go
```diff
@@ -52,9 +52,147 @@ func TestFeastServingURLs_URLs(t *testing.T) {
 	}
 }
 
-func TestStandardTransfomer_ToFeastStorageConfigs(t *testing.T) {
+func TestStandardTransformerConfig_ToFeastStorageConfigsForSimulation(t *testing.T) {
 	redisCfg := `{"is_redis_cluster": true,"serving_url":"online-storage.merlin.dev","redis_addresses":["10.1.1.10", "10.1.1.11"],"pool_size": 4,"max_retries": 1,"dial_timeout": "10s"}`
 	bigtableCfg := `{"serving_url":"10.1.1.3","project":"gcp-project","is_using_direct_storage":true,"instance":"instance","app_profile":"default","pool_size":3,"keep_alive_interval":"2m","keep_alive_timeout":"1m"}`
```
pradithya, line 57: You mentioned briefly that direct storage is not going to be supported first, is it still the case?

api/pkg/transformer/executor/predictor.go
```diff
@@ -0,0 +1,47 @@
+package executor
```
pradithya, line 1: Shall we call the package `simulator`? Or `executor` is more appropriate?

api/pkg/transformer/pipeline/operation.go
```diff
@@ -1,7 +1,118 @@
...
+	"fmt"
+	"reflect"
+	"sync"
+
+	"github.com/gojek/merlin/pkg/transformer/spec"
+	"github.com/gojek/merlin/pkg/transformer/types"
+	"github.com/gojek/merlin/pkg/transformer/types/table"
+)
 
 type Op interface {
```
pradithya, line 14: Can you add documentation for `Op` , `OperationIO`, `OperationTracing` ? including their field and methods. I think this is quite important part to be documented.

api/pkg/transformer/pipeline/operation.go
```diff
@@ -1,7 +1,118 @@
...
+		return err
+	}
+
+	ot.Input = append(ot.Input, input)
+	ot.Output = append(ot.Output, output)
+
+	return nil
+}
+
+func sanitizeIO(io map[string]interface{}) error {
```
pradithya, line 65: In what case `sanitizeIO` is required?

api/pkg/transformer/pipeline/variable_op.go
```diff
@@ -5,59 +5,77 @@ import (
...
+	*OperationTracing
 }
 
-func NewVariableDeclarationOp(variables []*spec.Variable) Op {
-	return &VariableDeclarationOp{
-		variableSpec: variables,
+func NewVariableDeclarationOp(variables []*spec.Variable, tracingEnabled bool) Op {
+	varOp := &VariableDeclarationOp{
+		variableSpec:     variables,
+		OperationTracing: NewOperationTracing(variables, types.VariableOpType),
```
pradithya, line 20: It seems redundant with line 23.

api/pkg/transformer/pipeline/operation.go
tiopramayudi: yeah, we do it sequentially. I removed the mutex

api/config/config_test.go
tiopramayudi: from executor POV it can support both, but from service POV it can only support feast serving because we needs to fetch feature metadata

api/pkg/transformer/executor/predictor.go
tiopramayudi: the purpose not only for simulation but also execute transformer in general, so in the future suppose turing want to call standard transformer it can orchestrate the executor from turing itself by supplying modelPredictor implementation (HTTP or mock)

api/pkg/transformer/pipeline/operation.go
tiopramayudi: it is to change `table` value to formatted value because `table` can't be marshalled without create custom marshaller

api/models/transformer.go
```diff
@@ -40,3 +44,22 @@ type Transformer struct {
 	EnvVars         EnvVars          `json:"env_vars"`
 	CreatedUpdated
 }
+
+// TransformerSimulation is payload that user supplies to simulate standard transformer config
+type TransformerSimulation struct {
+	Payload          types.JSONObject                `json:"payload"`
```
pradithya, line 50: Rename to `RequestPayload`

api/api/transformer_api.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/gojek/merlin/models"
+)
+
+// TransformerController
+type TransformerController struct {
+	*AppContext
+}
+
+// SimulateTransformer API handles simulation of standard transformer
+func (c *TransformerController) SimulateTransformer(r *http.Request, vars map[string]string, body interface{}) *Response {
```
pradithya, line 15: Can we add error logging when returning from `SimulateTransformer`

api/pkg/transformer/executor/transformer.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return &standardTransformer{
+		compiledPipeline: compiledPipeline,
+		modelPredictor:   executorConfig.modelPredictor,
+		executorConfig:   *executorConfig,
+	}, nil
+}
+
+// Predict will process all standard transformer request including preprocessing, model prediction and postprocess
+func (st *standardTransformer) Predict(ctx context.Context, requestBody types.JSONObject, requestHeaders map[string]string) (*types.PredictResponse, error) {
```
pradithya, line 66: Rename`Predict` to `Execute`

karzuo:
image
Do you want to change the background colour of "simulation" to disassociate it from pre and post processing as logical steps?

PR #256, opened 2022-06-10
Simulation review & improvement

ariefrahmansyah:


api/pkg/transformer/pipeline/variable_op.go
```diff
@@ -31,6 +31,8 @@ func (v *VariableDeclarationOp) Execute(context context.Context, env *Environmen
 
 	for _, varDef := range v.variableSpec {
 		name := varDef.Name
+		input := make(map[string]interface{})
```
pradithya, line 34: Can you undo all changes in `api/pkg/transformer/pipeline/variable_op.go` ? I think we should discuss again what's the best approach to populate operation that's currently has null `input`

pradithya:
The rest LGTM

PR #267, opened 2022-06-28
Add Liveness Probe to model

karzuo:
To add liveness probe to each model pod created via kserve. This probe will check liveness every 10s.

None

```release-note
Added Liveness probe to model pods in kserve & transformer pod
Added option to disable liveness probe of models by setting MERLIN_DISABLE_LIVENESS_PROBE=true in model
Added option to disable liveness probe of transformer by setting MERLIN_DISABLE_LIVENESS_PROBE=true in transformer
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


api/cluster/templater_test.go
```diff
@@ -92,6 +93,20 @@ var (
 		},
 	}
 
+	tfLivenessProbe = &corev1.Probe{
```
ariefrahmansyah, line 96: Why is this variable name prefixed with `tf`?

ariefrahmansyah (commented):
Can you document (comment in the code is enough) on how you come up with those liveness probe configs for each model type?

api/cluster/templater_test.go
karzuo: This one only used for tensorflow

karzuo:
The values are invented. There's no statistics background behind it. We just check every 10 seconds. Are you suggesting some tests to be done? More info?

api/cluster/templater.go
```diff
@@ -160,6 +161,19 @@ func createPredictorSpec(modelService *models.Service, config *config.Deployment
 					Container: corev1.Container{
 						Name:      kserveconstant.InferenceServiceContainerName,
 						Resources: resources,
+						LivenessProbe: &corev1.Probe{
+							Handler: corev1.Handler{
+								TCPSocket: &corev1.TCPSocketAction{
+									Port: intstr.IntOrString{
```
pradithya, line 167: Why do wr use port 9000 for tensorflow? This is grpc port and not guaranteed to be open. Can we make use endpoiny in port 8080?

api/cluster/templater.go
```diff
@@ -208,6 +252,21 @@ func createPredictorSpec(modelService *models.Service, config *config.Deployment
...
+							Handler: corev1.Handler{
+								HTTPGet: &corev1.HTTPGetAction{
+									Path:   fmt.Sprintf("/v1/models/%s", modelService.Name),
+									Scheme: "HTTP",
+									Port: intstr.IntOrString{
+										IntVal: 8080,
+									},
+								},
+							},
+							InitialDelaySeconds: 10,
```
pradithya, line 265: How do ypu drcide these values?

api/cluster/templater.go
karzuo: I didnt know that this value can be chosen by me. Initially I thought it is server configured, so I copied this value from someone else online. Was still experimenting in the last commit. Willl update this as well.

api/cluster/templater.go
karzuo: As discussed with @ariefrahmansyah I don't observe any delay in model starting. It is quite instant. Hence a 10s initial delay.

pradithya: Got, do you know what is the impact to slow model that are not able to answer after 10s delay?

api/cluster/templater.go
pradithya: If you are experimenting, I suggest to do it in the dev cluster by creating the inference service manifest manually so that you can observe if it's correct/not right away and iterate faster. If you use CI to experiment it will be very slow and wasted your time.

api/cluster/templater.go
karzuo: Do you suggest adding a readiness probe then?

pradithya: I think adding readiness probe and make it same as liveness probe is good idea, wdyt @ariefrahmansyah . Another thing is that can we make these values configurable at merlin level? Additionally we probably should provide configuration to disable it in case there is issue with specific user's deployment.

ariefrahmansyah: I think liveness probe should be less sensitive than readiness probe. So if the model is not ready yet (e.g. downloading huge size of model), we still give it a chance to finish its job and be ready. If the model 

+1 on making it configurable.

api/cluster/templater.go
```diff
@@ -160,6 +161,21 @@ func createPredictorSpec(modelService *models.Service, config *config.Deployment
...
```
pradithya, line 174: Can you add `failureThreshold` ? And also can you extract all these values into a constant to avoid repetition?

api/cluster/templater.go
```diff
@@ -36,23 +37,34 @@ import (
...
+	defaultPredictorDisableLiveness   = "false"
+	defaultTransformerDisableLiveness = "false"
 
 	annotationPrometheusScrapeFlag = "prometheus.io/scrape"
 	annotationPrometheusScrapePort = "prometheus.io/port"
 
 	prometheusPort = "8080"
+
+	liveProbePort             = 8080
+	liveProbeinitialDelaySec  = 60
```
pradithya, line 63: Can we make it 30 seconds?

pradithya (approved):
LGTM!

PR #268, opened 2022-07-05
Solve issue of duplicated env variables in model and transformer due to append

karzuo:
Previously, when creating containers for model or transformer, default values for certain environment variables are appended to the list. If the same variables has already been defined, this leads to error due to duplicates.

Solution:
Instead of appending the env var directly, we will accumulate a list of these variables first.
a. if these variables are not supposed to be modified by users, this list will merge to the existing list and overwrite any existing values with defaults
b. if these variables are supposed to be modified by users, the original list will merge to this list, overwriting the default values.

```release-note
Fix duplicated environment variables if user define with the same name
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cluster/templater.go
+++ b/api/cluster/templater.go
@@ -48,11 +48,9 @@ const (
 	envPredictorStorageURI       = "STORAGE_URI"
 	envPredictorDisableLiveness  = "MERLIN_DISABLE_LIVENESS_PROBE"
 
-	defaultPredictorArtifactLocation  = "/mnt/models"
-	defaultPredictorPort              = "8080"
-	defaultTransformerPort            = "8080"
-	defaultPredictorDisableLiveness   = "false"
-	defaultTransformerDisableLiveness = "false"
+	defaultPredictorArtifactLocation = "/mnt/models"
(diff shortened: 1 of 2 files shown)


api/cluster/templater.go
```diff
@@ -48,11 +48,9 @@ const (
 	envPredictorStorageURI       = "STORAGE_URI"
 	envPredictorDisableLiveness  = "MERLIN_DISABLE_LIVENESS_PROBE"
 
-	defaultPredictorArtifactLocation  = "/mnt/models"
-	defaultPredictorPort              = "8080"
-	defaultTransformerPort            = "8080"
-	defaultPredictorDisableLiveness   = "false"
-	defaultTransformerDisableLiveness = "false"
```
ariefrahmansyah, line 55: Why did these lines got deleted? Is it github rendering issue or actual deletion?

karzuo: Not used anymore, forgot to delete in last PR

ariefrahmansyah (approved):
LGTM! +100 on reusing models.MergeEnvVars()!

PR #270, opened 2022-07-19
Add Workflow to Trigger Remote Sync of Docs to caraml-dev/docs via Github Action

karzuo:
In order to sync our docs with caraml gitbook, we add a workflow that will be triggered with PR commits, Merge or Tag in Merlin. This workflow will request caraml docs repo to pull any changes in the doc folder of Merlin.

No

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/trigger-caraml-doc-sync.yml
+++ b/.github/workflows/trigger-caraml-doc-sync.yml
@@ -27,7 +27,7 @@ jobs:
     #       })
       - uses: actions/checkout@v3
       - id: trigger-caraml
-        uses: actions/trigger-doc-pull
+        uses: caraml-dev/docs/.github/actions/trigger-doc-pull@sync
         with:
           who-to-greet: 'Mona the Octocat'
       - run: echo random-number ${{ steps.trigger-caraml.outputs.random-number }}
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-caraml-doc-sync.yml
```diff
@@ -0,0 +1,24 @@
...
+  trigger-sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger Remote Doc Sync Workflow
+        uses: caraml-dev/docs/.github/actions/trigger-remote-docs-sync@main
+        with:
+          module: 'model'
+          git_https: 'https://github.com/gojek/merlin.git'
+          doc_folder: 'docs'
+          branch: 'main'
```
ariefrahmansyah, line 23: May I know why you only use `main` branch here instead of the job can be triggered by `on.push.branches` and `on.push.tags`?

I think we should only trigger the docs sync when a new tag/release is released because development can still happen in the main branch which means the number of changes introduced will be more compared to using only the tag release. If you are agree, we can rename `ref` as the field key and use `github.ref` as the value as suggested in the documentation: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context. Then you might need to update the actions in caraml-dev/docs as well.

karzuo: I have set it to tag only.

ariefrahmansyah (approved):

PR #272, opened 2022-07-28
Fix column type where all the value in a column is null

tiopramayudi:

This PR address several issues that triggered when user try impute `NaN` value from a table by using conditional update. Following are the issues:
* Series/Column that generated by feast input is in string type where the type defined in `featureTableSpec` is double, this is happening because currently we try to infer the type by the value, for this case all the values are `NaN` or `null` so the type can't be inferred and default to string
* Join two table where one of the columns type is string but the value all `NaN` or `null`, internally each column row has `nan` flag to indicate whether the row is `NaN`. For this case affected column in result table doesn't carry `nan` flag, so the filtering or row selection for `NaN` or `null` value won't work as expected since the operation relies on `nan` flag.

Fixes #

* For case 2 the fix is on the gota library [PR](https://github.com/gojekfarm/gota/pull/3), this PR to update the library version
* For case  1 the fix is to specify column type for features based on the type in given featurespec

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/transformer/feast/call.go
+++ b/api/pkg/transformer/feast/call.go
@@ -32,6 +32,33 @@ type call struct {
 	valueMonitoringEnabled  bool
 }
 
+func newCall(
+	fr *FeastRetriever,
+	featureTableSpec *spec.FeatureTable,
+	columns []string,
+	entitySet map[string]bool,
+) (*call, error) {
+	feastClient, err := fr.getFeastClient(featureTableSpec.Source)
+	if err != nil {
+		return nil, err
+	}
+	return &call{
+		featureTableSpec:  featureTableSpec,
+		defaultValues:     fr.defaultValues,
+		columns:           columns,
+		entitySet:         entitySet,
+		columnTypeMapping: getFeatureTypeMapping(featureTableSpec),
+
+		feastClient:   feastClient,
+		servingSource: featureTableSpec.Source,
+
+		logger: fr.logger,
+
+		statusMonitoringEnabled: fr.options.StatusMonitoringEnabled,
+		valueMonitoringEnabled:  fr.options.ValueMonitoringEnabled,
+	}, nil
+}
+
 // do create request to feast and return the result as table
 func (fc *call) do(ctx context.Context, entityList []feast.Row, features []string) callResult {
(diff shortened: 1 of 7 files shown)


api/pkg/transformer/feast/call.go
```diff
@@ -65,13 +65,23 @@ func (fc *call) do(ctx context.Context, entityList []feast.Row, features []strin
...
+}
+
 // processResponse process response from feast serving and create an internal feature table representation of it
 func (fc *call) processResponse(feastResponse *feast.OnlineFeaturesResponse) (*internalFeatureTable, error) {
 	responseStatus := feastResponse.Statuses()
 	responseRows := feastResponse.Rows()
 	entities := make([]feast.Row, len(responseRows))
 	valueRows := make([]transTypes.ValueRow, len(responseRows))
 	columnTypes := make([]types.ValueType_Enum, len(fc.columns))
+	columnTypeMapping := getFeatureTypeMapping(fc.featureTableSpec)
```
pradithya, line 84: Shall we cache it as a struct field so that it wont be recomputed for each call?

tiopramayudi: updated

api/pkg/transformer/feast/feature_retriever.go
```diff
@@ -215,10 +215,11 @@ func (fr *FeastRetriever) newCall(featureTableSpec *spec.FeatureTable, columns [
 	}
 
 	return &call{
```
pradithya, line 217: Rather than creating `call` struct directly, I think it's better to have a constructor `NewCall` and abstract away `getFeatureTypeMapping` behind it. Additionally, `defaultValues`, `logger`, `statusMonitoringEnabled`, `valueMonitoringEnabled` can all be derived from `fr`

ariefrahmansyah (approved):

api/pkg/transformer/feast/feature_retriever.go
tiopramayudi: updated

pradithya (approved):
LGTM

PR #275, opened 2022-08-04
Add support for more granular configuration of MLP Core API

romanwozniak:
### Context:

Merlin API depends on MLP Core API for retrieving the list of projects/secrets. Merlin UI also depends on MLP Core API for the same purpose. 

Currently, MLP API URL can only be configured to the same value (e.g http://mlp.example.com/api/v1) for both Merlin API and UI. However, moving forward, we would like to decouple these two values so API and UI can be configured separately. In this case, Merlin UI can use a relative MLP API URL (i.e. `/api/v1`), while Merlin API can call a cluster-local API endpoint (i.e. `http://mlp.mlp.svc.cluster.local:8080/v1`) to bypass authentication policy of MLP API.

### DIFF
--- a/api/cmd/api/ui_handler.go
+++ b/api/cmd/api/ui_handler.go
@@ -11,19 +11,10 @@ import (
 )
 
 type uiEnvHandler struct {
-	OauthClientID     string                `json:"REACT_APP_OAUTH_CLIENT_ID,omitempty"`
-	Environment       string                `json:"REACT_APP_ENVIRONMENT,omitempty"`
-	SentryDSN         string                `json:"REACT_APP_SENTRY_DSN,omitempty"`
-	DocURL            config.Documentations `json:"REACT_APP_MERLIN_DOCS_URL,omitempty"`
-	HomePage          string                `json:"REACT_APP_HOMEPAGE,omitempty"`
-	MerlinURL         string                `json:"REACT_APP_MERLIN_API,omitempty"`
-	MlpURL            string                `json:"REACT_APP_MLP_API,omitempty"`
-	DockerRegistries  string                `json:"REACT_APP_DOCKER_REGISTRIES,omitempty"`
-	MaxAllowedReplica int                   `json:"REACT_APP_MAX_ALLOWED_REPLICA,omitempty"`
+	*config.ReactAppConfig
 
 	DefaultFeastServingSource string                  `json:"REACT_APP_DEFAULT_FEAST_SOURCE,omitempty"`
 	FeastServingURLs          config.FeastServingURLs `json:"REACT_APP_FEAST_SERVING_URLS,omitempty"`
-	FeastCoreURL              string                  `json:"REACT_APP_FEAST_CORE_API,omitempty"`
 
 	MonitoringEnabled              bool   `json:"REACT_APP_MONITORING_DASHBOARD_ENABLED"`
 	MonitoringPredictionJobBaseURL string `json:"REACT_APP_MONITORING_DASHBOARD_JOB_BASE_URL"`
--- a/charts/merlin/templates/mlflow-deployment.yaml
+++ b/charts/merlin/templates/mlflow-deployment.yaml
@@ -16,9 +16,6 @@ spec:
     matchLabels:
       app: {{ template "mlflow.fullname" . }}
       release: {{ .Release.Name }}
-      {{- if .Values.mlflow.podLabels}}
-        {{- toYaml .Values.mlflow.podLabels | nindent 8 }}
-      {{- end}}
   strategy:
     type: RollingUpdate
     rollingUpdate:
--- a/charts/merlin/templates/mlflow-deployment.yaml
+++ b/charts/merlin/templates/mlflow-deployment.yaml
@@ -29,9 +29,9 @@ spec:
       labels:
         app: {{ template "mlflow.fullname" . }}
         release: {{ .Release.Name }}
-      {{- if .Values.mlflow.podLabels}}
-        {{ toYaml .Values.mlflow.podLabels | nindent 8 }}
-      {{- end}}
+        {{- if .Values.mlflow.podLabels }}
+          {{- toYaml .Values.mlflow.podLabels | nindent 8 }}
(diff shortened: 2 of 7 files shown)


api/cmd/api/ui_handler.go
```diff
@@ -11,19 +11,10 @@ import (
...
-	OauthClientID     string                `json:"REACT_APP_OAUTH_CLIENT_ID,omitempty"`
-	Environment       string                `json:"REACT_APP_ENVIRONMENT,omitempty"`
-	SentryDSN         string                `json:"REACT_APP_SENTRY_DSN,omitempty"`
-	DocURL            config.Documentations `json:"REACT_APP_MERLIN_DOCS_URL,omitempty"`
-	HomePage          string                `json:"REACT_APP_HOMEPAGE,omitempty"`
-	MerlinURL         string                `json:"REACT_APP_MERLIN_API,omitempty"`
-	MlpURL            string                `json:"REACT_APP_MLP_API,omitempty"`
-	DockerRegistries  string                `json:"REACT_APP_DOCKER_REGISTRIES,omitempty"`
-	MaxAllowedReplica int                   `json:"REACT_APP_MAX_ALLOWED_REPLICA,omitempty"`
+	*config.ReactAppConfig
```
pradithya, line 14: What's the criteria for a config to be within `ReactAppConfig` or `uiEnvHandler` ?

pradithya:
Out of curiosity and not a problem since it'll be squashed, is there reason the commit message is prefixed with `-` ?

api/cmd/api/ui_handler.go
romanwozniak: `ReactAppConfig` holds config values, that are populated directly via `REACT_APP_*` env vars configured on the `merlin` container. 

`uiEnvHandler` – leverages config from `ReactAppConfig` and adds on top of it cross-component config values (i.e. feature flags): https://github.com/gojek/merlin/blob/41b648d61c1dd151dc02d88a317e7b90a39af3ee/api/cmd/api/main.go#L112

charts/merlin/values.yaml
```diff
@@ -219,6 +202,28 @@ merlin:
     # This should be the actual DNS registered
     trackingURL: "http://www.example.com"
 
+  ui:
```
pradithya, line 205: Can you also update https://github.com/gojek/merlin/blob/main/charts/merlin/values-e2e.yaml?

charts/merlin/templates/mlflow-deployment.yaml
```diff
@@ -29,9 +26,9 @@ spec:
       labels:
         app: {{ template "mlflow.fullname" . }}
         release: {{ .Release.Name }}
-      {{- if .Values.mlflow.podLabels}}
-        {{ toYaml .Values.mlflow.podLabels | nindent 8 }}
-      {{- end}}
+        {{- if .Values.mlflow.podLabels }}
```
pradithya, line 29: This can be verified by adding `mlflow.podLabels` in https://github.com/gojek/merlin/blob/main/charts/merlin/values-e2e.yaml

romanwozniak:
> Out of curiosity and not a problem since it'll be squashed, is there reason the commit message is prefixed with `-` ?

This is to make the commit message bullet-list like, and to separate two or more lines of changes from each other
```yaml
- fix chart template with pod labels
- bump up the chart version 
```
The same can be achieved by a) making commits even more granular b) just use `;` two separate the lines

So I guess, it's just a habit.

api/cmd/api/ui_handler.go
pradithya: I think I get it. So config in `uiEnvHandler` has `REACT_APP_*` json representation for writing, not for reading?
That's confusing if true @ariefrahmansyah

charts/merlin/values.yaml
romanwozniak: `values-e2e.yaml` has no UI config keys

api/cmd/api/ui_handler.go
romanwozniak: > So config in uiEnvHandler has `REACT_APP_*` json representation for writing, not for reading?

This is accurate. Generally, I'd like us to spend more time addressing the maintainability of our existing codebases because they expanded substantially over time and now new changes added to the repo are addressing a local problem (i.e. how to make this specific function work correctly) without taking into account the big picture of the entire repo. 

Example: 
These two environment variables have no effect on the deployment at all:
https://github.com/gojek/merlin/blob/28d764b4fc84009c78aadddda0962c415e6f4d5b/charts/merlin/templates/merlin-deployment.yaml#L97-L100

charts/merlin/values.yaml
pradithya: Ah yeah

charts/merlin/templates/mlflow-deployment.yaml
romanwozniak: Yeah, that makes sense. Added 👍

romanwozniak:
@pradithya since you are on it, can you approve it?

api/cmd/api/ui_handler.go
ariefrahmansyah: > So config in `uiEnvHandler` has `REACT_APP_*` json representation for writing, not for reading?
Do you mean `uiEnvHandler` writes `REACT_APP_*` and additional config values as the response when the user visits `/env.js`? If it's true, then I get you, otherwise, can you elaborate? @pradithya 

> I'd like us to spend more time addressing the maintainability of our existing codebases
Agree. One of the steps that I would like to take is to refactor the deployment configuration so it can have a better locality and still improve reusability. We are planning to achieve this in this semester 🤞🏻

pradithya (approved):
LGTM!

api/cmd/api/ui_handler.go
romanwozniak: > I would like to take is to refactor the deployment configuration

As part of this, it would be great to move most of the configuration from env vars to a config file, and only keep the sensitive overrides in env vars.

PR #279, opened 2022-09-13
Update UPI Version 

pradithya:
**What this PR does / why we need it**:

Update caraml-upi-protos version to >= 0.3.1 . This version includes the protocol update mainly adding the concept of Table in upi payload.

### Changes 

1. Pin caraml-upi-protos version in merlin-sdk to >= 0.3.1
2. Update pyfunc-server tests to use the latest protocol
3. Update merlin-sdk pyfunc upi integration test to use the latest protocol

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Checklist**

- [X] Added unit test, integration, and/or e2e tests
- [X] Tested locally
- [N.A.] Updated documentation
- [N.A.] Update Swagger spec if the PR introduce API changes
- [N.A.] Regenerated Golang and Python client if the PR introduce API changes

PR #280, opened 2022-09-14
Pyfunc upi multiprocessing

pradithya:

Add ability to run multiprocessed UPI/gRPC server. The implementation is largely similar to https://github.com/grpc/grpc/tree/master/examples/python/multiprocessing which leverage `SO_REUSEPORT` . The idea is to spawn several process of UPI server and bind to same {interface,port} which is allowed when `SO_REUSEPORT` is set to 1. In this case, load balancing is handled by kernel of the underlying OS. There is caveat though, as load balancing become OS-dependent. For example, load balancing is not working in MacOS but it's working fine using the pyfunc docker image. 

To control the number of workers, user can set env var `WORKERS` (similar to http server), in addition to that, user can also specify number of thread per worker using `GRPC_CONCURRENCY`.

## Load test result

I did loadtests to check how multiprocess improves service throughput. The model has 1 replica with 2 CPU and it only echo-ing back request to the client. 
Command to invoke the load test is. 
```
./ghz pyfunc-multiproc.integration-test:8080 --insecure --call=caraml.upi.v1.UniversalPredictionService/PredictValues -d '{}' --duration=30s  --connections=8
```
Following is the result for 4 workers vs 1 worker.  In summary, having more worker improves the throughput considerably (almost 2x in this case). 

With 4 workers:
```
Summary:
  Count:        125347
  Total:        30.00 s
  Slowest:      66.77 ms
  Fastest:      0.99 ms
  Average:      11.76 ms
  Requests/sec: 4178.24
```

With 1 worker:
```
Summary:
  Count:        54979
  Total:        30.00 s
  Slowest:      78.13 ms
  Fastest:      1.80 ms
  Average:      27.11 ms
  Requests/sec: 1832.60
```

Note: this PR assume https://github.com/gojek/merlin/pull/279 is merged 

None

```release-note
Allow users to run multi-processed UPI server
```

**Checklist**

- [X] Tested locally
- [X] Updated documentation


python/pyfunc-server/pyfuncserver/protocol/upi/server.py
```diff
@@ -25,7 +26,28 @@ def __init__(self, model: PyFuncModel, config: Config):
...
+
+        self._run_server()
+
+    def _run_server(self):
+        """
+            Start a server in a subprocess.
+
+        """
+        options = self._config.grpc_options
+        options.append(('grpc.so_reuseport', 1))
```
tiopramayudi, line 47: what 1 represent for?

tiopramayudi:
The rest is LGTM

python/pyfunc-server/pyfuncserver/protocol/upi/server.py
pradithya: It's enable flag

tiopramayudi (approved):

PR #283, opened 2022-09-15
Fix manual url prefixing

pradithya:

Fix manual URL prefixing which cause badly formatted url when the scheme is not http. This implementation will fallback to use `http://` scheme when it's not provided. 

**Checklist**

- [X] Tested locally

### DIFF
--- a/api/pkg/transformer/server/rest/server.go
+++ b/api/pkg/transformer/server/rest/server.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/pprof"
-	"net/url"
 	"os"
 	"os/signal"
 	"runtime/debug"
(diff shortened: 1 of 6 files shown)


pradithya:
Looks like it breaks some e2e test. Let me fix it first.

api/pkg/transformer/server/rest/server.go
```diff
@@ -358,3 +358,13 @@ func getHeaders(headers http.Header) map[string]string {
 	}
 	return resultHeaders
 }
+
+// getUrl return url or add default http scheme if scheme is not specified
+func getUrl(rawUrl string) (string, error) {
```
tiopramayudi, line 363: why returning err if there is no chance it will throw error?

pradithya: Fixed, I missed removing it after changing the implementation of `getUrl`

api/pkg/transformer/server/rest/server.go
```diff
@@ -67,16 +67,13 @@ type hystrixHttpClient interface {
...
 	return NewWithHandler(o, nil, logger)
 }
 
 // New initializes a new Server with pipeline handler
-func NewWithHandler(o *config.Options, handler *pipeline.Handler, logger *zap.Logger) *HTTPServer {
-	predictURL := fmt.Sprintf("%s/v1/models/%s:predict", o.ModelPredictURL, o.ModelFullName)
-	if !strings.Contains(predictURL, "http://") {
-		predictURL = "http://" + predictURL
-	}
+func NewWithHandler(o *config.Options, handler *pipeline.Handler, logger *zap.Logger) (*HTTPServer, error) {
```
tiopramayudi, line 75: this function also won't returning error

tiopramayudi:
the rest is LGTM

ariefrahmansyah (approved):

PR #284, opened 2022-09-16
Use golanci-lint for performing static check analysis in merlin golang codebase

pradithya:

This PR replace static code check from golint to golangci-lint, since golint is deprecated. At the same time the PR enable golangci-lint as part of the CI as well as fixes existing errors.
The golanci-lint is configured within `api/.golangci.yml` in which we enable all default linter and following optional linters (see https://golangci-lint.run/usage/linters/ for details):
1. errorlint 
2. goimports
3. gofmt
4. bodyclose

None

**Checklist**

- [N.A.] Added unit test, integration, and/or e2e tests
- [X] Tested locally


api/pkg/transformer/symbol/function/geospatial.go
```diff
@@ -55,45 +55,45 @@ func (l *LatLong) toS2ID(level int) interface{} {
...
 	if err != nil {
 		panic(err)
 	}
 
-	switch latLong.(type) {
+	switch latLong := ll.(type) {
 	case []*LatLong:
 		var value []interface{}
-		for _, ll := range latLong.([]*LatLong) {
+		for _, ll := range latLong {
```
karzuo, line 66: the alternating usage from ll to latlong and back to ll here is pretty confusing and may lead to potential bugs next time. Can we have more meaningful name? Or at least a different name for the inner ll?

karzuo (approved):
LGTM

api/pkg/transformer/symbol/function/geospatial.go
pradithya: Good point. Updated it.

PR #288, opened 2022-09-27
Fix autoscaling target value when memory usage is selected as the target metrics

pradithya:

Knative expect the value of `autoscaling.knative.dev/target` annotation to be in `MiB` units when `autoscaling.knative.dev/metric` is set to memory (i.e. memory-bound autoscaling). 
Existing implementation expect users to provide memory usage target % and pass it directly to knative without conversion. It leads to incorrect behavior, in which the autoscaling target will only vary between 0-100MiB.  
This PR address this bugs by converting the target memory utilization % provided by users to absolute memory requested by using user's requested memory resource. 

```release-note
Fix incorrect memory-bound autoscaling behavior
```

**Checklist**

- [X] Tested locally

### DIFF
--- a/api/cluster/resource/utils.go
+++ b/api/cluster/resource/utils.go
@@ -20,7 +20,8 @@ func toKNativeAutoscalerMetrics(metricsType autoscaling.MetricsType, metricsValu
 	case autoscaling.MemoryUtilization:
 		// Calculate memory target value based on memory request * metricsValue
 		// ref: https://github.com/krithika369/turing/blob/6cc3b5f673ccf66e9c1351b2d91382315e86f8ce/api/turing/cluster/knative_service.go#L166
-		memoryTarget := computeResource(resourceReq.MemoryRequest, metricsValue/100)
+		memoryTarget := scaleQuantity(resourceReq.MemoryRequest, metricsValue/100)
+		// Divide value by (1024^2) to convert to Mi
 		return knautoscaling.Memory, fmt.Sprintf("%.0f", float64(memoryTarget.Value())/math.Pow(1024, 2)), nil
 	case autoscaling.RPS:
 		targetValue := fmt.Sprintf("%.0f", metricsValue)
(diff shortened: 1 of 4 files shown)


api/cluster/resource/utils.go
```diff
@@ -2,25 +2,34 @@ package resource
...
 	case autoscaling.CPUUtilization:
-		return knautoscaling.CPU, nil
+		targetValue := fmt.Sprintf("%.0f", metricsValue)
+		return knautoscaling.CPU, targetValue, nil
 	case autoscaling.MemoryUtilization:
-		return knautoscaling.Memory, nil
+		// Calculate memory target value based on memory request * metricsValue
+		// ref: https://github.com/krithika369/turing/blob/6cc3b5f673ccf66e9c1351b2d91382315e86f8ce/api/turing/cluster/knative_service.go#L166
+		memoryTarget := computeResource(resourceReq.MemoryRequest, metricsValue/100)
+		return knautoscaling.Memory, fmt.Sprintf("%.0f", float64(memoryTarget.Value())/math.Pow(1024, 2)), nil
```
karzuo, line 24: maybe can add same comments from Turing's code for better understanding.

api/cluster/resource/utils.go
```diff
@@ -43,3 +52,19 @@ func toKServeDeploymentMode(deploymentMode deployment.Mode) (string, error) {
...
+		scaledMilliValue = resourceQuantity.MilliValue()
+	}
+	percentageValue := float64(scaledMilliValue) * fraction
+	newValue := int64(math.MaxInt64)
+	if percentageValue < math.MaxInt64 {
+		newValue = int64(percentageValue)
+	}
+	newquantity := resource.NewMilliQuantity(newValue, resource.BinarySI)
+	return *newquantity
+}
```
karzuo, line 70: Same for this which is linked to https://github.com/knative/serving/blob/release-0.14/pkg/reconciler/revision/resources/queue.go#L115

karzuo (approved):
LGTM

ariefrahmansyah (approved):
lgtm. thanks for providing us with reference docs

PR #293, opened 2022-09-29
Make sdk compatible with python 3.7 to 3.10

karzuo:

As python 3.7 is reaching the end of life, we are upgrading merlin to support development in python 3.8, 3.9 and 3.10.
In the first part of this upgrade, we tested our sdk against the various python versions and updated the merlin examples accordingly.

**Caveat**: This PR only makes sdk installable, usable and runnable using python 3.7 to 3.10. Subsequent work will need to be carried out to make Merlin fully compatible with python 3.8 to 3.10

**Test results:** 
1. No change in current setup dependencies are required for merlin-sdk to be installed on python 3.7 to 3.10.
2. We will need to update scikit-learn to >=1.1.2 and joblib to >=1.2.0 to fully support development work on python 3.8 to 3.10. However, this will be done after we update our kserve to use the respective libraries versions.
3. In order to support python 3.11, we will need to wait for new release of scipy. On top of that, protobuf will need to be >4 and mlflow >=1.26.1. As we still largely depend on libraries using protobuf 3, this version will not support python 3.11 at this stage.

**Subsequent work:**
To make Merlin fully compatible with python 3.8 to 3.10, the following tasks need to be completed subsequently:
1. We will need to update scikit-learn to >=1.1.2 and joblib to >=1.2.0 to fully support development work on python 3.8 to 3.10. However, this will be done after we update our kserve to use the respective libraries versions.
2. Update pyfunc server and batch prediction server to build and run pyfunc images according to python version used.

Fixes #

```release-note
Updated dependencies version to use some newer libraries for setup
Updated merlin examples to be compatible with python 3.7 to 3.10
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


examples/batch/env.yaml
```diff
@@ -1,7 +1,7 @@
 dependencies:
 - pip:
-    - joblib>=1.2.0
+    - joblib>=0.13.0,<1.2.0  # >=1.2.0 upon upgrade of kserve's version
```
pradithya, line 3: Why are we allowing older version of joblib now?

python/batch-predictor/docker/app.Dockerfile
```diff
@@ -20,6 +20,6 @@ FROM ${BASE_IMAGE}
 ARG MODEL_URL
 RUN gsutil cp -r ${MODEL_URL} .
 RUN /bin/bash -c ". activate merlin-model && \
-    sed -i 's/pip$/pip=20.2.4/' /model/conda.yaml && \
+    sed -i 's/pip$/pip=22.2.2/' /model/conda.yaml && \
```
pradithya, line 23: Is there any reason to upgrade pip version too? This was forced to allow dependency conflict which sometimes happen during model building. Without this user's PyFunc and batch predictor model might not be able to be built.

python/sdk/setup.py
```diff
@@ -24,18 +24,18 @@
     "certifi>=2017.4.17",
     "python-dateutil>=2.1",
     "six>=1.10",
-    "mlflow>=1.2.0,<=1.23.0",
+    "mlflow>=1.2.0,<=1.23.0", #  for py3.11 due to proto -> "mlflow>=1.26.1",
     "google-cloud-storage>=1.19.0",
     "boto3>=1.9.84",
     "urllib3>=1.23",
     "PyPrind>=2.11.2",
-    "google-auth>=1.11.0,<2.0dev",
```
pradithya, line 32: Why is it removed?

python/sdk/setup.py
```diff
@@ -46,14 +46,14 @@
...
-    "google-cloud-bigquery-storage==0.7.0",
-    "grpcio==1.22.0",
-    "recursive-diff==1.0.0",
+    "xgboost==1.6.2",
+    "scikit-learn==1.0.1",  # >=1.1.2 upon python 3.7 deprecation
+    "joblib>=0.13.0,<1.2.0",  # >=1.2.0 upon upgrade of kserve's version
+    "mypy>=0.812",
+    "google-cloud-bigquery>=1.18.0",
+    "google-cloud-bigquery-storage>=0.7.0",
+    "grpcio>=1.31.0",
```
pradithya, line 55: I think need to also pin grpcio to <1.49.0 to avoid having same issue as pyfunc

pradithya (changes requested):
You'll need to add test in CI that run sdk test in various Python version

examples/batch/env.yaml
karzuo: mainly because we haven't update kserve to use 1.2.0. For consistency (and non-breaking change) reason, I think we should use the older version first. But I think I will update the kserve first, so this line probably gets changed before it is merged.

pradithya: >we haven't update kserve to use 1.2.0

What kind of update are you referring to?

python/sdk/setup.py
pradithya: Yeah, agree with that statement. But try to limit the amount of change to only for supporting Python >3.7. This will make debugging issue easier compared to a big bang changes.

examples/batch/env.yaml
karzuo: We are deploying kserve using built image, hence the joblib there is frozen at whatever level was latest when the image was built. If we rebuild this time, it will be joblib 1.2.0. That will be a breaking (but necessary) change for our users using pyfunc and pyfunc v2.

In addition, sklearn server is tagged at 1.0.1 we should update that to >=1.1.2 so that users of python 3.8 is not limited by our version on kserve.

pradithya: But pyfunc and pyfunc v2 (batch prediction) doesn't use kserve image. So I'm not quite sure what's the purpose here. 
Additionally, even with currently new joblib version (joblib>=1.2.0) we didn't encounter any issue, so allowing the older version seems strange to me.

karzuo: That's correct. What I mean is that we should make it consistent. Otherwise users may be confused because when they use standard models, they have to use old joblib but when they use pyfunc and v2, they have to use 1.2.0

pradithya: I got your point, but even for the standard model we already use joblib>=1.2.0 (https://github.com/gojek/merlin/blob/0ab987ab6a1629c74d3e91a82a6b6c62b86e2c02/python/sdk/setup.py#L51) I wonder from where do you get `joblib>=0.13.0`

karzuo: no, that was updated by me 9 days ago. Unfortunately it won't work for standard model, that's how I discovered it is controlled by kserve on the server side.

Got the version from here:
https://github.com/kserve/kserve/blob/master/python/sklearnserver/setup.py#L37

pradithya: Ah ok, but then how did it pass our end to end test?

karzuo: Because our test uses pre-saved models instead of saving new ones with joblib

pradithya: Make sense, thanks!

pradithya:
For e2e-test in github CI, shall we use more recent version of Python?

karzuo:
Not until we fix pyfunc and batch.

pradithya:
> Not until we fix pyfunc and batch.

Got it

pradithya (approved):
LGTM

PR #294, opened 2022-09-30
Restrict UPI deployment for unsupported model type

pradithya:

As of now, only PyFunc model is compatible with UPI and trying to deploy other model type will result in unknown behavior. The PR is to prevent unsupported model type from being deployed using UPI Protocol by returning 400 status code. 

None

None

**Checklist**

- [X] Tested locally

Signed-off-by: Pradithya Aria <pradithya.pura@go-jek.com>


ariefrahmansyah:
Should we also disable it from the UI as well?

tiopramayudi (approved):

pradithya:
> Should we also disable it from the UI as well?

I think not necessary since we'll eventually support that

PR #296, opened 2022-10-03
Move Test DB password to github secret

pradithya:
Our test in github CI rely on postgres instance which we spawn and destroy during testing. The db credential is currently exposed set as env var which might raise alarm even though the DB itself doesn't have sensitive data. This PR move the db creds to github secret.

Fixes #

None

```release-note

```


karzuo (approved):
LGTM

ariefrahmansyah (approved):

PR #300, opened 2022-10-03
Set default protocol to HTTP_JSON

pradithya:

When redeploying an existing model (that has protocol HTTP_JSON) from UI, protocol value is currently set to empty. On the other hand, standard transformer interpret empty protocol as UPI_V1 due to incorrect condition. This leads to issue of model returning 404 since standard transformer is running gRPC service. This PR fix this bug by:
1. Set HTTP_JSON as default from UI when protocol is empty
2. Set HTTP_JSON as default in the API when protocol is empty
3. Fallback to HTTP_JSON when protocol is not set in standard transformer

Issue when redeploying existing model via UI

```release-note

```

**Checklist**

- [x] Tested locally


karzuo (approved):
LGTM. But I wonder why I could redeploy without same issue via the UI.

ariefrahmansyah (approved):

pradithya:
@karzuo The redeployment is fine, the problem happens when you send test request to the deployment that has standard transformer

PR #301, opened 2022-10-04
Upgrade pyfunc & batch predictor to support python3.7 to 3.10

karzuo:
Python 3.7 will become obsolete soon. We need to upgrade Merlin to support newer python versions, namely python 3.8, 3.9 and 3.10.

As part 2 & 3 of this effort, this PR mainly does the following:
1. Upgrade pyfunc to support python 3.7 to 3.10
2. Upgrade batch predictor to support python 3.7 to 3.10

**Notable changes:**
1. Create base image for each python version for pyfunc*:  i.e. merlin-pyfunc-base-py37 etc
2. Create base images for each python version for batch*:  i.e. merlin-pyspark-base-py37 etc
3. Update Gitlab pipeline to test sdk, api, pyfunc and batch using python 3.7 to 3.10
4. Capture user's python version major and minor when user create a model version. This is be used to determine the python version of the model to be logged/deployed. Any python version settings in the env.yaml file will be replaced to the captured version. In case user is still using older merlin sdk, python 3.7.* is assumed to ensure backward compatibility.
5. Mapping of python version captured from user to respective pyfunc/batch base docker image.

**Base images Updates:**
1. For batch base Dockerfile, pyspark 2.4.5 is not compatible with >= py3.8, hence upgraded till 3.0.0 which does not officially support py3.9 and py3.10, but tested to work on our servers. We are not able to upgrade to 3.1.1 as it is not compatible with our spark operator version (completed jobs won't shutdown)
2. miniconda has been replaced with miniforge as latest miniconda available does not support py3.10 as of now.

Fixes #

```release-note
- Support for python 3.7 to 3.10 for Merlin (SDK, Pyfunc and Batch predictors)
- Backward compatibility with older sdk
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


.github/workflows/merlin.yml
```diff
@@ -318,8 +334,8 @@ jobs:
     runs-on: ubuntu-latest
     needs: 
       - build-api
-      - build-batch-predictor
-      - build-pyfunc-server
+      #- build-batch-predictor
```
pradithya, line 337: Is it intended?

db-migrations/30_versions_python_version.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Introduce Python version column, with default as version 3.7 which is the
+-- only supported major version as of the introduction of the column.
+ALTER TABLE versions ADD COLUMN python_version varchar(16) NOT NULL DEFAULT '3.7.*';
```
pradithya, line 3: I think it can be
```
ALTER TABLE versions ADD COLUMN python_version varchar(16) NOT NULL DEFAULT '3.7.*' WITH VALUES;
```

And no need to remove the default in line 5 ?

python/batch-predictor/docker/base.Dockerfile
```diff
@@ -12,44 +12,87 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/spark-operator/spark-py:v2.4.5
+ARG SPARK_VERSION=v3.0.0
```
pradithya, line 15: Are we planning to make `SPARK_VERSION` configurable?

python/batch-predictor/docker/base.Dockerfile
```diff
@@ -12,44 +12,87 @@
...
 
-COPY batch-predictor /merlin-spark-app
-COPY sdk /sdk
+# Copy batch predictor application entrypoint to protected directory
 COPY batch-predictor/merlin-entrypoint.sh /opt/merlin-entrypoint.sh
-RUN conda env create -f /merlin-spark-app/environment.yaml && \
-    rm -rf /root/.cache
+
+# Configure non-root user
+ARG username=spark
```
pradithya, line 62: Are you planning to make username, spark_uid, and spark_gid configurable?

python/batch-predictor/docker/base.Dockerfile
```diff
@@ -12,44 +12,87 @@
...
+FROM gcr.io/spark-operator/spark-py:${SPARK_VERSION}
+
+# Switch to user root so we can add additional jars and configuration files.
+USER root
+
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 ENV PATH /opt/conda/bin:$PATH
-ENV ARROW_PRE_0_15_IPC_FORMAT 1
+
+ARG SPARK_OPERATOR_VERSION=v1beta2-1.3.7-3.1.1
```
pradithya, line 24: Are you planning to make SPARK_OPERATOR_VERSION and SPARK_BQ_CONNECTOR_VERSION configurable?

python/batch-predictor/env310.yaml
```diff
@@ -0,0 +1,8 @@
+name: merlin-model
```
pradithya, line 1: Can we move all those environment yaml files (env37, env38, etc) to its own directory?

python/batch-predictor/requirements.txt
```diff
@@ -1,6 +1,7 @@
-pyspark==2.4.5
+findspark
```
pradithya, line 1: What is `findspark` used for?

python/pyfunc-server/docker/base.Dockerfile
```diff
@@ -12,14 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM continuumio/miniconda3
+FROM condaforge/miniforge3
```
pradithya, line 15: What are the difference between miniconda with miniforge? Is there any risk of regression to our existing users?

python/sdk/client/models/model_endpoint_alert.py
```diff
@@ -48,19 +43,14 @@ class ModelEndpointAlert(object):
         'alert_conditions': 'alert_conditions'
     }
 
-    def __init__(self, model_id=None, model_endpoint_id=None, environment_name=None, team_name=None, alert_conditions=None, _configuration=None):  # noqa: E501
```
pradithya, line 51: Do you know what is the `_configuration` used for?

python/sdk/test/batch/model/env.yaml
```diff
@@ -1,9 +1,9 @@
 dependencies:
   - pip:
       - mlflow==1.23.0
-      - scikit-learn==1.0.1  # >=1.1.2 upon python 3.7 deprecation
+      - scikit-learn==1.0.2  # >=1.1.2 upon python 3.7 deprecation
       - joblib>=0.13.0,<1.2.0  # >=1.2.0 upon upgrade of kserve's version
-      - numpy==1.19.5 # https://github.com/Azure/MachineLearningNotebooks/issues/1314
+      - numpy>=1.19.5 # https://github.com/Azure/MachineLearningNotebooks/issues/1314
```
pradithya, line 6: For test dependencies, it's better to pin it to the working version to avoid test regression when new version coming out.

python/batch-predictor/requirements.txt
```diff
@@ -1,6 +1,7 @@
...
+pyspark==3.0.1
 mlflow>=1.2.0,<=1.23.0
 cloudpickle==2.0.0
-pyarrow==0.14.1
-protobuf>=3.0,<4.0.0dev
+pyarrow>=0.14.1
```
pradithya, line 5: What happen if users uses older version of pyarrow?

examples/pyfunc-upi/Pyfunc.ipynb
```diff
@@ -0,0 +1,667 @@
+{
```
pradithya, line 1: You'll need to stop the habit of including unrelated change into a PR. It should be a separate one.

pradithya:
Thanks @karzuo for implementing this! What's your strategy for safely rolling this out? I think there are several notable changes that will requires extra attention as it might break existing model in production. Namely:
1. spark version
2. pyarrow version
3. miniconda to miniforge

python/batch-predictor/docker/base.Dockerfile
karzuo: nope, but this helps with upgrading efforts and helps prevent errors, since the values are used in more than 1 place.

pradithya: In that case I think you should use `ENV` instead of `ARG`

python/batch-predictor/requirements.txt
karzuo: this is for this line which seems to use it in some cases when for some reason spark cannot be found (maybe due to package upgrades etc) https://github.com/gojek/merlin/blob/87dc891183b2eb9d728e70bb569ebba053c64f89/python/batch-predictor/main.py#L34

python/pyfunc-server/docker/base.Dockerfile
karzuo: miniconda is company-driven (Anaconda) and miniforge is community-driven (mainly to resolve compatibility issues that miniconda has yet to provide). Also some licensing restrictions by miniconda according to thread below:

https://stackoverflow.com/questions/60532678/what-is-the-difference-between-miniconda-and-miniforge

Other than that, no real differences according to sources online.

python/batch-predictor/requirements.txt
karzuo: I guess it should still work depending on the python versions. Anyway 0.14.1 was released in 2019 and only supports up till python 3.7. Loosening this restriction is to support python 3.8 to 3.10. I will add a upper limit for testing purpose, as per suggested.

karzuo:
I suggest to roll out this as a beta version in staging for a period of time, say 3 months, to ensure that it is sufficiently stable before we roll out to production. Using of older sdk version should still work but we will have to tell people that if they use older sdk, they should only use python 3.7

pradithya:
> I suggest to roll out this as a beta version in staging for a period of time, say 3 months, to ensure that it is sufficiently stable before we roll out to production.

3 months is too long :)

I think to make it safer. We should make the `dockerfilePath` and `buildContextURI` configurable. The idea is that when we have issue in happen for existing users we can safely rollback to the previous release without having to rollback the entire deployment.

So instead of having baseImage configuration
```
    dockerfilePath: "./Dockerfile"
    buildContextURI: "git://github.com/gojek/merlin.git#refs/tags/v0.19.0"
    baseImage:
      "3.7.*": "image-3.7"
      "3.8.*": "image-3.8"
```

We should make
```
    baseImage:
      "3.7.*": 
         imageName: "image-3.7"
         dockerfilePath: "./Dockerfile"
         buildContextURI: "git://github.com/gojek/merlin.git#refs/tags/v0.19.0"
      "3.8.*": "image-3.8"
         imageName: "image-3.8"
         dockerfilePath: "./Dockerfile"
         buildContextURI: "git://github.com/gojek/merlin.git#refs/tags/v0.19.0"
```

This will allow as to cleanly rollback 3.7.*

karzuo:
hmm. I don't think we need that. We can just change the base image reference that the version that was previously working.

python/sdk/client/models/model_endpoint_alert.py
karzuo: Other than being used by api_client, looking through old generated codes seems to suggest that it is `_configuration ` is not used. In some cases where they are called, the function also does not exist in the default `Configuration()`. For example:
https://github.com/swagger-api/swagger-codegen/blob/621a7330ee7940623af34f935eee9fda6992a300/samples/client/petstore/python/petstore_api/models/animal.py#L82

which seems to suggest there is no real need for the `_configuration` object.

karzuo:
> > I suggest to roll out this as a beta version in staging for a period of time, say 3 months, to ensure that it is sufficiently stable before we roll out to production.
> 
> 3 months is too long :)
> 
> I think to make it safer. We should make the `dockerfilePath` and `buildContextURI` configurable. The idea is that when we have issue in happen for existing users we can safely rollback to the previous release without having to rollback the entire deployment.
> 
> So instead of having baseImage configuration
> 
> ```
>     dockerfilePath: "./Dockerfile"
>     buildContextURI: "git://github.com/gojek/merlin.git#refs/tags/v0.19.0"
>     baseImage:
>       "3.7.*": "image-3.7"
>       "3.8.*": "image-3.8"
> ```
> 
> We should make
> 
> ```
>     baseImage:
>       "3.7.*": 
>          imageName: "image-3.7"
>          dockerfilePath: "./Dockerfile"
>          buildContextURI: "git://github.com/gojek/merlin.git#refs/tags/v0.19.0"
>       "3.8.*": "image-3.8"
>          imageName: "image-3.8"
>          dockerfilePath: "./Dockerfile"
>          buildContextURI: "git://github.com/gojek/merlin.git#refs/tags/v0.19.0"
> ```
> 
> This will allow as to cleanly rollback 3.7.*

Updated based on our discussions, and successfully tested and deployed in dev. :)

python/sdk/client/models/model_endpoint_alert.py
pradithya: Got it, thanks!

api/config/environment.go
```diff
@@ -61,7 +61,7 @@ type ResourceRequestConfig struct {
 }
 
 func initEnvironmentConfigs(path string) []EnvironmentConfig {
-	cfgFile, err := ioutil.ReadFile(path)
+	cfgFile, err := os.ReadFile(path)
```
pradithya, line 64: What's the difference?

api/batch/resource.go
```diff
@@ -57,7 +57,7 @@ const (
 
 var (
 	// Hardcoded based on python/batch-predictor/docker/base.Dockerfile
-	mainApplicationPath = "local:///merlin-spark-app/main.py"
+	mainApplicationPath = "local:///home/spark/merlin-spark-app/main.py"
```
pradithya, line 60: I think this modification will not allow us to cleanly rollback. Maybe need to make it configurable as well.

pradithya:
Awesome work @karzuo ! I left last few remaining feedbacks above but overall LGTM.
What's quite important is to allow us to use previous version of "3.7.*". i.e. we should test whether this works
```
    baseImage:
      "3.7.*": 
         imageName: // previous release image name
         dockerfilePath: // previous release docker file path
         buildContextURI: // previous release context uri
         // other configuration if necessary
      "3.8.*": 
         imageName: // new release image name
         dockerfilePath: // bew release docker file path
         buildContextURI: // bew release context uri
         // other configuration if necessary
```

In fact I think this is what we should deploy , so that people who need to stay in 3.7 will have time to migrate to newer version without breaking their work at all. cc: @ariefrahmansyah

python/batch-predictor/docker/app.Dockerfile
```diff
@@ -18,9 +18,9 @@ FROM ${BASE_IMAGE}
 # Get the model
 
 ARG MODEL_URL
-RUN gsutil cp -r ${MODEL_URL} .
+RUN gsutil -m cp -r ${MODEL_URL} .
```
ariefrahmansyah, line 21: +1 on using `-m` option for parallel download

ariefrahmansyah: Anyway, do see any improvement after adding this `-m` option?

python/batch-predictor/docker/base.Dockerfile
```diff
@@ -12,44 +12,84 @@
...
+
+# Switch to Spark user
+USER ${USER}
+WORKDIR ${HOME}
+
+# Install miniconda
+ENV CONDA_DIR ${HOME}/miniconda3
+ENV PATH ${CONDA_DIR}/bin:$PATH
+
+RUN wget --quiet https://github.com/conda-forge/miniforge/releases/download/4.14.0-0/Miniforge3-4.14.0-0-Linux-x86_64.sh -O miniconda.sh && \
```
ariefrahmansyah, line 75: nit: Can we store miniforge version to ENV as well?

```
ENV MINIFORGE_VERSION=4.14.0-0
```

python/batch-predictor/docker/env310.yaml
```diff
@@ -0,0 +1,8 @@
+name: merlin-model
+dependencies:
+  - python=3.10
+  - pip=22.2.2
```
ariefrahmansyah, line 4: Why do you pin the pip version to 22.2.2?

python/pyfunc-server/docker/Dockerfile
```diff
@@ -20,7 +20,7 @@ ARG MODEL_URL
 WORKDIR /pyfunc-server
 RUN gsutil cp -r ${MODEL_URL} .
 RUN /bin/bash -c ". activate merlin-model && \
-    sed -i 's/pip$/pip=22.2.2/' model/conda.yaml && \
+    sed -i 's/pip$/pip=20.2.4/' model/conda.yaml && \
```
ariefrahmansyah, line 23: Why do we use pip 20.2.4 instead of 22.2.2 here? The rest of env*.yaml files use 22.2.2 though

python/pyfunc-server/docker/base.Dockerfile
```diff
@@ -12,14 +12,17 @@
...
+
+ARG PYTHON_VERSION
 
 RUN wget -qO- https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-367.0.0-linux-x86_64.tar.gz | tar xzf -
```
ariefrahmansyah, line 19: batch-repdictor base image uses `ARG GCLOUD_VERSION=405.0.1`. can we also update pyfunc to use 405.0.1?

api/api/versions_api.go
```diff
@@ -115,6 +117,10 @@ func (c *VersionsController) CreateVersion(r *http.Request, vars map[string]stri
 		return BadRequest("Valid label key/values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.")
 	}
 
+	if versionPost.PythonVersion == "" { // if using older sdk, this will be blank
+		versionPost.PythonVersion = DEFAULT_PYTHON_VERSION
+	}
+
```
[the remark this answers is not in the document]
karzuo: I think it will be an overkill, especually since the value in this case is not determined by users but by our code. There should be no concern about ourselves giving invalid version format. 😅

python/batch-predictor/docker/env310.yaml
karzuo: Older pips are missing wheels information for packages used by 3.10 or m1, resulting them to try to build from source which often fails. Newer pip is able to fix that. Anyway this is only used in our base image building. I have defaulted to old version for users environment

python/pyfunc-server/docker/Dockerfile
karzuo: This pip is used in user environment. We default to older version to allow dependency conflict as pointed out by @pradithya

python/batch-predictor/docker/base.Dockerfile
```diff
@@ -12,44 +12,84 @@
...
-    /opt/conda/bin/conda clean -tipsy && \
-    ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
-    echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc && \
-    echo "conda activate base" >> ~/.bashrc
 
-WORKDIR /
-RUN wget -qO- https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-265.0.0-linux-x86_64.tar.gz | tar xzf -
+# Install gcloud SDK
 ENV PATH=$PATH:/google-cloud-sdk/bin
+ARG GCLOUD_VERSION=405.0.1
```
ariefrahmansyah, line 54: nit: should use ENV GCLOUD_VERSION

ariefrahmansyah (approved):
LGTM! 🎉 

Thanks for working on this! Awesome job!

api/config/config.go
```diff
@@ -75,6 +75,29 @@ type ReactAppConfig struct {
 	UPIDocumentation  string         `envconfig:"REACT_APP_UPI_DOC_URL" json:"REACT_APP_UPI_DOC_URL,omitempty"`
 }
 
+type BaseImageConfigs map[string]BaseImageConfig
+
+type BaseImageConfig struct {
```
tiopramayudi, line 80: can you add comment to all exported name

api/config/config.go
```diff
@@ -75,6 +75,29 @@ type ReactAppConfig struct {
...
+	ImageName string `json:"imageName"`
+
+	// Dockerfile Path within the build context
+	DockerfilePath string `json:"dockerfilePath"`
+
+	// GCS URL Containing build context
+	BuildContextURI string `json:"buildContextURI"`
+}
+
+func (b *BaseImageConfigs) Decode(value string) error {
```
tiopramayudi, line 91: can you add comment for this?

api/api/versions_api.go
```diff
@@ -115,6 +117,10 @@ func (c *VersionsController) CreateVersion(r *http.Request, vars map[string]stri
 		return BadRequest("Valid label key/values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.")
 	}
 
+	if versionPost.PythonVersion == "" { // if using older sdk, this will be blank
```
tiopramayudi, line 120: I think we need to return badrequest if the given python version is not supported.

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -317,11 +331,16 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
 		labelOrchestratorName: "merlin",
 	}
 
+	baseImageTag, ok := c.config.BaseImages[version.PythonVersion]
```
tiopramayudi, line 334: how about if the user give 3.7 instead of 3.7.* for python version?

api/batch/resource.go
```diff
@@ -57,7 +58,8 @@ const (
 
 var (
 	// Hardcoded based on python/batch-predictor/docker/base.Dockerfile
-	mainApplicationPath = "local:///home/spark/merlin-spark-app/main.py"
+	homeDir             = os.Getenv("IMG_BUILDER_PRED_HOME_DIR")
```
pradithya, line 61: It has to be part of the base image configuration for the prediction job since the `homeDir` will be specific to a docker image. i.e. previous release docker image expect `/spark/` but now it's `/home/spark` thus it won't work. It will be simpler if we keep it hardcoded to `/spark/

api/config/environment.go
pradithya: Ah I see, didn't know that it's deprecated.

api/pkg/imagebuilder/imagebuilder.go
karzuo: the value is controlled by us, not by the user. User's part is managed from python sdk side already.

api/config/config.go
```diff
@@ -79,14 +79,10 @@ type BaseImageConfigs map[string]BaseImageConfig
...
 type BaseImageConfig struct {
-	// docker image name with path
-	ImageName string `json:"imageName"`
-
-	// Dockerfile Path within the build context
-	DockerfilePath string `json:"dockerfilePath"`
-
-	// GCS URL Containing build context
-	BuildContextURI string `json:"buildContextURI"`
+	ImageName       string `json:"imageName"`       // docker image name with path
```
pradithya, line 82: nit: The convention in Golang is to put comments on top of the field so that godoc can generate the documentation automatically. You can read more about writing more idiomatic golang in https://go.dev/doc/effective_go . Highly recommended if you haven't read it before.

api/models/prediction_job.go
```diff
@@ -45,6 +45,7 @@ type Config struct {
 	ServiceAccountName string                        `json:"service_account_name"`
 	JobConfig          *spec.PredictionJob           `json:"job_config"`
 	ImageRef           string                        `json:"image_ref"`
+	MainAppPath        string                        `json:"main_app_path"`
```
pradithya, line 48: Will this config be stored in DB

karzuo: Nope. It won't be. This is related to db?

pradithya: Ah it's stored in db but doesn't require new column

api/config/config.go
karzuo: oh ok will change back. I shifted it to side because it could create nice formatting.

api/api/versions_api.go
karzuo: at this stage we have no idea if the model version is supported since we don't have it hardcoded anywhere. This might also not be a batch or pyfunc model, which... may work even if we don't officially support it?

karzuo: As per discussion, the enforcement of python version is delegated to the sdk as per:
https://github.com/gojek/merlin/blob/upgrade-pyfunc/python/sdk/merlin/model.py#L430
https://github.com/gojek/merlin/blob/upgrade-pyfunc/python/sdk/setup.py#L74

For older sdk version, this value is not captured but set to 3.7.* by default as per code above.

PR #307, opened 2022-10-19
Add DB configurations

ariefrahmansyah:

Fixes broken pipe from Merlin to DB

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cmd/api/setup.go
+++ b/api/cmd/api/setup.go
@@ -63,6 +63,7 @@ func initDB(cfg config.DatabaseConfig) (*gorm.DB, func()) {
 	}
 
 	sqlDB.SetConnMaxIdleTime(cfg.ConnMaxIdleTime)
+	sqlDB.SetConnMaxLifetime(cfg.ConnMaxLifetime)
 	sqlDB.SetMaxIdleConns(cfg.MaxIdleConns)
 	sqlDB.SetMaxOpenConns(cfg.MaxOpenConns)
 
--- a/api/cmd/api/setup.go
+++ b/api/cmd/api/setup.go
@@ -63,6 +63,8 @@ func initDB(cfg config.DatabaseConfig) (*gorm.DB, func()) {
 	}
 
 	sqlDB.SetConnMaxIdleTime(cfg.ConnMaxIdleTime)
+	sqlDB.SetMaxIdleConns(cfg.MaxIdleConns)
+	sqlDB.SetMaxOpenConns(cfg.MaxOpenConns)
 
 	return db, func() { db.Close() } //nolint:errcheck
 }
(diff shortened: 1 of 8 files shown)


api/config/config.go
```diff
@@ -124,6 +124,10 @@ type DatabaseConfig struct {
 	Password      string `envconfig:"DATABASE_PASSWORD" required:"true"`
 	Database      string `envconfig:"DATABASE_NAME" default:"mlp"`
 	MigrationPath string `envconfig:"DATABASE_MIGRATIONS_PATH" default:"file://db-migrations"`
+
+	ConnMaxIdleTime time.Duration `envconfig:"DATABASE_CONN_MAX_IDLE_TIME" default:"10m"`
+	MaxIdleConns    int           `envconfig:"DATABASE_MAX_IDLE_CONNS" default:"10"`
```
pradithya, line 129: Can we make it 0 as default?

api/config/config.go
```diff
@@ -124,6 +124,10 @@ type DatabaseConfig struct {
...
+	MaxOpenConns    int           `envconfig:"DATABASE_MAX_OPEN_CONNS" default:"100"`
```
pradithya, line 130: Do we need to update helm chart to include these?

api/cmd/api/setup.go
```diff
@@ -56,6 +56,16 @@ func initDB(cfg config.DatabaseConfig) (*gorm.DB, func()) {
...
 	db.LogMode(false)
+
+	sqlDB := db.DB()
+	if sqlDB == nil {
+		panic("fail to get the underlying database connection")
+	}
+
+	sqlDB.SetConnMaxIdleTime(cfg.ConnMaxIdleTime)
+	sqlDB.SetMaxIdleConns(cfg.MaxIdleConns)
+	sqlDB.SetMaxOpenConns(cfg.MaxOpenConns)
```
pradithya, line 67: Can you add `SetConnMaxLifetime` (https://github.com/go-sql-driver/mysql/issues/529) ?

api/config/config.go
ariefrahmansyah: Updated

api/config/config.go
ariefrahmansyah: Updated

api/cmd/api/setup.go
ariefrahmansyah: Added

charts/merlin/values.yaml
```diff
@@ -245,6 +246,11 @@ merlin-postgresql:
   # The password secret will be created using this password
   postgresqlPassword: merlin
 
+  connMaxIdleTime: 0
```
pradithya, line 249: I think it should be string and non 0
```
connMaxIdleTime: 10m
connMaxLifetime: 30m
```

pradithya (approved):
LGTM!!  thanks @ariefrahmansyah

karzuo (approved):
LGTM

PR #324, opened 2022-11-24
Update version.endpoint() to reuse _version_endpoints attribute

ariefrahmansyah:

Currently, in Python SDK, Version's endpoint() method implementations call GetVersionEndpoints API instead of reusing the existing `_version_endpoints` attribute that already got the data.

By reusing the `_version_endpoints`, we can cut the call to the Merlin server which improves the latency and reduce Merlin server workload.

```
Old implementation latency: 0.4594380855560303 second
New implementation latency: 0.00037217140197753906 second
```

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/sdk/merlin/endpoint.py
+++ b/python/sdk/merlin/endpoint.py
@@ -89,7 +89,7 @@ class VersionEndpoint:
     @property
     def env_vars(self) -> Dict[str, str]:
         env_vars = {}
-        if self._env_vars is not None:
+        if self._env_vars:
             for ev in self._env_vars:
                 env_vars[ev.name] = ev.value
         return env_vars
--- a/python/sdk/merlin/endpoint.py
+++ b/python/sdk/merlin/endpoint.py
@@ -16,8 +16,9 @@ from enum import Enum
 from typing import Dict
 
 import client
-from merlin.autoscaling import AutoscalingPolicy, MetricsType, RAW_DEPLOYMENT_DEFAULT_AUTOSCALING_POLICY, \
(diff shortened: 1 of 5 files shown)


python/sdk/merlin/endpoint.py
```diff
@@ -88,8 +89,9 @@ def environment(self) -> Environment:
     @property
     def env_vars(self) -> Dict[str, str]:
         env_vars = {}
-        for ev in self._env_vars:
-            env_vars[ev.name] = ev.value
+        if self._env_vars is not None:
```
pradithya, line 92: nit: it can be shortened to
```
if self._env_vars:
```

python/sdk/merlin/model.py
```diff
@@ -664,14 +667,18 @@ def endpoint(self) -> Optional[VersionEndpoint]:
...
+        :return: VersionEndpoint or None
         """
-        endpoint_api = EndpointApi(self._api_client)
-        ep_list = endpoint_api. \
-            models_model_id_versions_version_id_endpoint_get(
-                model_id=self.model.id, version_id=self.id)
 
-        for endpoint in ep_list:
+        # For backward compatibility, we called VersionEndpoint API if _version_endpoints empty.
+        endpoints = self._version_endpoints
```
pradithya, line 674: Shall we simplify it and use list_endpoint ? e.g.
```
if self._version_endpoints is None:
    self._version_endpoints = self.list_endpoint()

for endpoint in self._version_endpoints:
    # find the default

```

pradithya (approved):
LGTM! Thanks!

PR #325, opened 2022-11-25
Run image build using kaniko credential

pradithya:

`gsutil` command within the pyfunc and batch predictor docker image build failed with GCS access error when running on workload identity node pool even though kaniko pod has been given admin privilege to GCS. Apparently, the gsutil credential is not properly initialize and run without using the kaniko's credential.  This PR properly initialize gsutil credential to use the one provided to the kaniko pod. (similar to https://github.com/caraml-dev/turing/pull/291)

None

```release-note

```

**Checklist**

- [x] Tested locally


python/batch-predictor/docker/app.Dockerfile
```diff
@@ -18,8 +18,12 @@ FROM ${BASE_IMAGE}
 # Get the model
 
 ARG MODEL_URL
+ARG GOOGLE_APPLICATION_CREDENTIALS
+
+# Run docker build using the preovided credential
+RUN gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS}
```
pradithya, line 24: @krithika369 I wonder if the service account file will already be included in the docker build context?

krithika369: That's a good question. Tested it e2e for Turing and it works. So it's available.

python/batch-predictor/docker/app.Dockerfile
```diff
@@ -18,8 +18,12 @@ FROM ${BASE_IMAGE}
 # Get the model
 
 ARG MODEL_URL
+ARG GOOGLE_APPLICATION_CREDENTIALS
+
+# Run docker build using the preovided credential
```
tiopramayudi, line 23: nit: typo here

tiopramayudi:
LGTM

tiopramayudi (approved):

ariefrahmansyah (approved):

PR #323, opened 2022-11-17
Track merlin-sdk and python version

karzuo:
**What this PR does / why we need it**:

- Send merlin sdk version and user’s python version as part of the User-Agent header when sending request to merlin API. 
- Add Prometheus middleware that tracks the user-agent
- Expose metrics of those information and include it in merlin admin dashboard.

To avoid high cardinality, the prometheus metrics only logs user-agent that contains merlin-sdk or python keyword.

The resulting `/metrics` is something like this:

```
merlin_http_duration_seconds_sum{name="ListProjects",path="/projects",user_agent=""} 0.0046886670000000005
merlin_http_duration_seconds_count{name="ListProjects",path="/projects",user_agent=""} 2
...
merlin_http_duration_seconds_sum{name="ListProjects",path="/projects",user_agent="merlin-sdk/0.25.0 python/3.7.11"} 0.000624833
merlin_http_duration_seconds_count{name="ListProjects",path="/projects",user_agent="merlin-sdk/0.25.0 python/3.7.11"} 1
```

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
Track user's merlin-sdk and python version
```

**Checklist**

- [x] Added unit test, integration, and/or e2e tests
- [x] Tested locally
- [ ] Updated documentation
- [ ] Update Swagger spec if the PR introduce API changes
- [ ] Regenerated Golang and Python client if the PR introduce API changes

PR #329, opened 2022-11-29
Fix log API

pradithya:

Existing log api implementation requires the `http.ResponseWriter` to implement `http.Flusher` interface. By default `http.ResponseWriter` implements `http.Flusher`, however, due to https://github.com/gojek/merlin/pull/323 this is broken. 
This PR address the issue by:
1. Relax the requirement of log API to only use `Flush` when the passed `http.ResponseWriter` implements `http.Flusher`, to avoid other issue in the future where any new middlerware break the requirement above. 
2. Add `Flush` method to the prometheus middleware.

Fixes #

None

**Checklist**

- [X] Tested locally

### DIFF
--- a/api/api/log_api.go
+++ b/api/api/log_api.go
@@ -35,13 +35,6 @@ type LogController struct {
 func (l *LogController) ReadLog(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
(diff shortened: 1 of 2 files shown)


ariefrahmansyah (approved):
thanks, aria!

PR #331, opened 2023-01-12
Remove vault dependency

shydefoo:
* This PR removes the dependency on vault. Clusters are accessed via configuration passed at runtime.
* Upgrades mlp version to 1.7.5
* This [PR](https://github.com/gojek/mlp/pull/74) should be merged first before merging

No

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/cmd/api/setup.go
```diff
@@ -506,17 +463,11 @@ func initVersionEndpointService(cfg *config.Config, builder imagebuilder.ImageBu
...
-		log.Panicf("unable to retrieve secret for cluster %s from vault: %v", cfg.ImageBuilderConfig.ClusterName, err)
-	}
+func initLogService(cfg *config.Config) service.LogService {
+	creds := mlpcluster.NewK8sClusterCreds(&cfg.ImageBuilderConfig.K8sConfig)
 	ctl, err := cluster.NewController(cluster.Config{
-		Host:       imgBuilderClusterSecret.Endpoint,
-		CACert:     imgBuilderClusterSecret.CaCert,
-		ClientCert: imgBuilderClusterSecret.ClientCert,
-		ClientKey:  imgBuilderClusterSecret.ClientKey,
+		Host:        cfg.ImageBuilderConfig.K8sConfig.Cluster.Server,
```
mbruner, line 469: Host and other fields should be cleaned up now

charts/merlin/templates/_helpers.tpl
```diff
@@ -57,3 +57,12 @@
...
 {{- end -}}
+
+
+{{- define "merlin.getEnvironmentSecret" -}}
+{{- if .Values.mlp.environmentConfigSecret.name }}
+secretName: {{ .Values.mlp.environmentConfigSecret.name }}
+{{- else }}
+secretName: {{ template "merlin.fullname" .}}-environments
+{{- end -}}
+{{- end -}}
```
ariefrahmansyah, line 68: I think it's better to just return the value without the `secretName: ` field. You can use it as a variable and no need to configure the indentation.

charts/merlin/templates/merlin-deployment.yaml
```diff
@@ -311,11 +311,10 @@ spec:
             echo "Running Swagger UI..."
             /usr/share/nginx/run.sh
       {{- end }}
-
       volumes:
       - name: config
-        configMap:
-          name: {{ template "merlin.fullname" .}}-environments
+        secret:
+        {{- include "merlin.getEnvironmentSecret" . | indent 10 }}
```
ariefrahmansyah, line 317: My [above comment](https://github.com/gojek/merlin/pull/331/files#r1073284723) can avoid you using indent here.

charts/merlin/templates/_helpers.tpl
```diff
@@ -57,3 +57,12 @@
 {{- printf .Values.mlflow.backendStoreUri -}}
 {{- end -}}
 {{- end -}}
+
+
+{{- define "merlin.getEnvironmentSecret" -}}
```
ariefrahmansyah, line 62: I prefer to rename it to `merlin.environmentsSecretName` because it only returns a string for the secret name instead of the whole environments object.

charts/merlin/templates/merlin-config-secret.yaml
```diff
@@ -1,5 +1,7 @@
+# Secret is created only if mlp envconfig secret not set
```
ariefrahmansyah, line 1: Can we rename this file to `merlin-environments-secret.yaml`? In retrospect, `merlin-config-secret.yaml` is confusing for me. Is it a configmap or secret?

api/cmd/api/setup.go
shydefoo: Done

charts/merlin/templates/merlin-config-secret.yaml
shydefoo: Sure, yep it's a secret. Let's call it `merlin-environments-secret.yaml`

api/cmd/api/setup.go
shydefoo: Removed.

pradithya (approved):
LGTM! Thanks especially for updating the local development experience @shydefoo !

PR #337, opened 2023-01-19
Add gke-gcloud-auth-plugin plugin to dockerfile

shydefoo:
This PR adds the gke-gcloud-auth-plugin binary to the Dockerfile, in preparation to remove dependency on vault and static k8s credentials. See https://github.com/gojek/merlin/pull/331. This PR can be merged first without affecting existing merlin code in main branch.

Reason:
* GKE clusters now support [authentication via plugin](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke)
* `gke-gcloud-auth-plugin` is a go binary. Source code can be found [here](https://github.com/kubernetes/cloud-provider-gcp/tree/master/cmd/gke-gcloud-auth-plugin)
* There are 3 ways to install the plugin:
    1) via gcloud cli: `gcloud components install gke-gcloud-auth-plugin --quiet`
    2) via `go install <package>`
        * Howver, `go install <plugin>` returns this error
        ```sh
        ➜ go install github.com/kubernetes/cloud-provider-gcp/cmd/gke-gcloud-auth-plugin@929b1354f904a55d42389ae2a75021
        042e62e40f      
        go: downloading github.com/kubernetes/cloud-provider-gcp v0.0.0-20221223184326-929b1354f904
        go: github.com/kubernetes/cloud-provider-gcp/cmd/gke-gcloud-auth-plugin@929b1354f904a55d42389ae2a75021042e62e40
        f (in github.com/kubernetes/cloud-provider-gcp@v0.0.0-20221223184326-929b1354f904):
                The go.mod file for the module providing named packages contains one or
                more replace directives. It must not contain directives that would cause
                it to be interpreted differently than if it were the main module.
        ```
    3) via `apt-get install google-cloud-sdk-gke-gcloud-auth-plugin`
        * The go binary is built using go-alpine image, so this approach is not suitable as apt-get is only applicable to debian images

* Final approach uses `gcloud-sdk-cli:alpine` image as a build stage to download the `gke-gcloud-auth-plugin` binary, and copy the binary into the final image.
The only disadvantage of installing `gke-gcloud-auth-plugin` this way is that gcloud cli is not available. `gke-gcloud-auth-plugin` defaults to using gcloud to retrieve gcp access token. 
To get around this, the argument `--use_appliaction_default_credentials` needs to be supplied when providing model cluster / image builder cluster K8sConfig from 
https://github.com/gojek/merlin/pull/331
```
gke-gcloud-auth-plugin --use_application_default_credentials
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,13 @@ RUN go mod tidy
 RUN go build -o bin/merlin_api ./cmd/api
 
 # ============================================================
-# Build stage 2: Copy binary and config file
+# Build stage 2: Download gke auth plugin
+# ============================================================
+FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine as gke-plugin-builder
+RUN gcloud components install gke-gcloud-auth-plugin --quiet
+
+# ============================================================
+# Build stage 3: Copy binary and config file
 # ============================================================
 FROM alpine:3.12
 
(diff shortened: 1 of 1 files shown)


shydefoo:
Hey guys any comments on this? @ariefrahmansyah @tiopramayudi

pradithya (approved):
LGTM, thanks @shydefoo !

PR #343, opened 2023-01-27
Add graceful termination for pyfunc http server

pradithya:

Currently, pyfunc server implementation will immediately quit after receiving termination signal from OS/Kubernetes. This is not ideal for event such as scale down or redeployment. 
This PR implements graceful termination as well as update `run.sh` to forward termination signal to the pyfunc server. 

None

**Checklist**

- [X] Tested locally

### DIFF
--- a/python/pyfunc-server/pyfuncserver/protocol/rest/server.py
+++ b/python/pyfunc-server/pyfuncserver/protocol/rest/server.py
@@ -1,4 +1,8 @@
+import asyncio
 import logging
+import os
+import signal
+import sys
 
 import tornado
 from prometheus_client import CollectorRegistry
@@ -7,6 +11,15 @@ from pyfuncserver.config import Config
 from pyfuncserver.metrics.handler import MetricsHandler
 from pyfuncserver.model.model import PyFuncModel
 from pyfuncserver.protocol.rest.handler import HealthHandler, LivenessHandler, PredictHandler
+
+
+async def sig_handler(server):
+    logging.info("Stop accepting new connection")
+    server.stop()
+    logging.info("Closing existing connection")
+    await server.close_all_connections()
+    logging.info("Stopping ioloop")
+    tornado.ioloop.IOLoop.current().stop()
 
 
 class HTTPServer:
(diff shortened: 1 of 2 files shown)


python/pyfunc-server/run.sh
```diff
@@ -13,6 +13,23 @@
...
+  wait "$server_pid"
+}
+
+sigint_handler() {
+  echo "Terminating.."
+  kill -INT "$server_pid"
+  wait "$server_pid"
+}
+
+trap sigterm_handler SIGTERM
```
tiopramayudi, line 28: what is the purpose of trap here?

python/pyfunc-server/pyfuncserver/protocol/rest/server.py
```diff
@@ -32,11 +45,15 @@ def create_application(self):
     def start(self):
         self._http_server = tornado.httpserver.HTTPServer(
             self.create_application())
-
         logging.info("Listening on port %s", self.http_port)
         self._http_server.bind(self.http_port)
         logging.info("Will fork %d workers", self.workers)
         self._http_server.start(self.workers)
+
+        for signame in ('SIGINT', 'SIGTERM'):
```
tiopramayudi, line 53: do we need to handle the same thing for gRPC?

pradithya: Yes, we'll need to do the same thing for grpc too.

python/pyfunc-server/run.sh
pradithya: The trap will allow catching signal that's received by bash and bash can forward it to the process via the handler.

python/pyfunc-server/run.sh
```diff
@@ -13,6 +13,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+sigterm_handler() {
+  echo "Terminating.."
+  kill -TERM "$server_pid"
+  wait "$server_pid"
```
tiopramayudi, line 19: what happen if we don't wait here?

pradithya: Since the bash is the parent process of the pyfunc server then if it exit immediately it will also kill the pyfunc server.

tiopramayudi:
LGTM

tiopramayudi (approved):

PR #346, opened 2023-01-30
Refactor deployed model labels

ariefrahmansyah:

The labels implementation of Merlin resources (model services, batch jobs) is having two major limitations:
1. Hard coded to use "gojek.com/" as prefix so user can change it
2. Only propagate project's labels even though version labels are available via [new_model_version()](https://github.com/gojek/merlin/blob/main/python/sdk/merlin/fluent.py#L157) function.

These drawbacks limit the user to have more flexibility on using label for their tracking.

This PR make following changes:
1. Allow user to specify label prefix
2. Merge version labels with project labels. Version labels have more priority.
3. Introduce "component" labels to tag model-version, model-endpoint, batch-job, and image-builder resources.

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/models/metadata.go
```diff
@@ -30,7 +29,6 @@ const (
 	labelAppName          = "gojek.com/app"
```
pradithya, line 29: Why are these not configurable using the user prefix?
```
	labelAppName          = "gojek.com/app"
	labelOrchestratorName = "gojek.com/orchestrator"
	labelEnvironment      = "gojek.com/environment"
```

pradithya (changes requested):
I think it'll be better to make the prefix consistent across all labels.

pradithya:
Can you review https://github.com/gojek/merlin/pull/348 and rebase this PR against that. I think the e2e flakiness is due to kserve

api/models/metadata.go
```diff
@@ -18,34 +18,47 @@ import (
...
+	LabelComponent        = "component"
+	LabelEnvironment      = "environment"
+	LabelOrchestratorName = "orchestrator"
+	LabelStreamName       = "stream"
+	LabelTeamName         = "team"
+
+	ComponentBatchJob      = "batch-job"
+	ComponentImageBuilder  = "image-builder"
+	ComponentModelEndpoint = "model-endpoint"
+	ComponentModelVersion  = "model-version"
```
ariefrahmansyah, line 37: I introduce a new label, called "component". Currently, only the image builder uses this component label, and this PR adds this component to other resources.

api/models/metadata.go
ariefrahmansyah: Updated

api/models/metadata.go
```diff
@@ -18,34 +18,47 @@ import (
...
+	LabelOrchestratorName: true,
+	LabelStreamName:       true,
+	LabelTeamName:         true,
+}
+
+var prefix string
+
+// InitKubernetesLabeller builds a new KubernetesLabeller Singleton
+func InitKubernetesLabeller(p string) {
+	prefix = p
```
tiopramayudi, line 53: can we do validation whether the prefix is valid or not? e.g "goto/gojek/" is not a valid prefix, if this value passed during init of merlin api, user won't be able to do anything

api/models/metadata.go
pradithya: Noted!

api/models/metadata.go
```diff
@@ -64,28 +77,33 @@ func (metadata *Metadata) Scan(value interface{}) error {
 
 func (metadata *Metadata) ToLabel() map[string]string {
 	labels := map[string]string{
-		labelOrchestratorName: "merlin",
-		labelAppName:          metadata.App,
-		labelEnvironment:      metadata.Environment,
-		labelStreamName:       metadata.Stream,
-		labelTeamName:         metadata.Team,
+		prefix + LabelAppName:          metadata.App,
```
tiopramayudi, line 80: does user need to explictily add `/` in the prefix?

pradithya (approved):
LGTM, thanks @ariefrahmansyah

api/models/metadata.go
```diff
@@ -18,34 +18,47 @@ import (
...
 
 const (
-	labelTeamName         = "gojek.com/team"
-	labelStreamName       = "gojek.com/stream"
-	labelAppName          = "gojek.com/app"
-	labelOrchestratorName = "gojek.com/orchestrator"
-	labelEnvironment      = "gojek.com/environment"
-	labelUsersPrefix      = "gojek.com/%s"
+	LabelAppName          = "app"
+	LabelComponent        = "component"
```
tiopramayudi, line 28: do we need to expose this const?

api/models/metadata.go
ariefrahmansyah: Yes, so it's up to user to set their own delimiter or not at all.

api/models/metadata.go
ariefrahmansyah: Currently, only LabelOrchestratorName that used by other packages, so yeah, we will change other to local variables.

api/models/metadata.go
ariefrahmansyah: Great suggestion. We added the validation for prefix.

api/models/metadata.go
tiopramayudi: but the latest changes, we enforce user to add `/` on their prefix, since prefix "gojek.com" won't pass the validation

ariefrahmansyah: Correct. We've added the validation as you suggested when initializing the prefix to guard this enforcement.

tiopramayudi (approved):

tiopramayudi:
lgtm

PR #351, opened 2023-02-17
Support resource configuration for imagebuilder

terryyylim:

Currently, Merlin image building jobs don't support resource configuration, and requests.cpu is hardcoded. Without requests.memory configured on them, it's common to see jobs failing with OOMKilled error if multiple jobs are running concurrently. This PR supports resource configuration so that some reasonable amount of memory can be set to mitigate this issue. In addition, the GKE autoscaler can spin up another node if multiple jobs are running concurrently, instead of placing all of the jobs onto the same node, and causing them to fight for available memory.

Fixes #

```release-note
NONE
```

**Checklist**


pradithya:
@terryyylim LGTM, can you update the helm chart as well?

pradithya (approved):

charts/merlin/values-e2e.yaml
```diff
@@ -74,6 +74,13 @@ merlin:
         dockerfilePath: "docker/app.Dockerfile"
         buildContextURI: "git://github.com/gojek/merlin.git#refs/tags/v0.19.0"
         mainAppPath: /merlin-spark-app/main.py
+    resourceRequestsLimits:
```
pradithya, line 77: nit: I think `resources` is more concise.

pradithya (approved):
One little nitpick, LGTM thanks @terryyylim

ariefrahmansyah:
Can you also update Merlin helm chart in caraml-dev/helm-charts: https://github.com/caraml-dev/helm-charts/tree/main/charts/merlin ?

terryyylim:
> Can you also update Merlin helm chart in caraml-dev/helm-charts: https://github.com/caraml-dev/helm-charts/tree/main/charts/merlin ?

👌🏻 Yup! I'll update it once this chart has been tested with `0.26.0-rc4`.

PR #353, opened 2023-02-21
Upgrade gota dependency

tiopramayudi:

Previously we faced issue when joining two tables that have different type on their join columns. This is due to recent change in gota optimization that using hash join, due to different key lookup generated by gota the join result is not correct. The fix already merged in this [PR](https://github.com/gojekfarm/gota/pull/6). Hence this PR aims to update the gota dependency and adding test that related to the gota fix

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


pradithya (approved):
LGTM, thanks @tiopramayudi

ariefrahmansyah:
LGTM @tiopramayudi . Thanks!

Btw, you can delete GitHub Actions cache via UI or programmatically: https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28

Here's how I do it:

```
# list cache
curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/gojek/merlin/actions/caches

# delete cache using gh cli
gh api --silent --method DELETE -H "Accept: application/vnd.github+json" "/repos/gojek/merlin/actions/caches?key=${CACHE_KEY}"
```

PR #354, opened 2023-02-23
Add missing method for imagebuilder resource custom decoder

terryyylim:

Resource configuration for imagebuilder was introduced in https://github.com/gojek/merlin/pull/351, but it was missing `Decode` method that is required by envconfig, as described in the documentation - https://github.com/kelseyhightower/envconfig#custom-decoders.

Fixes #

```release-note
NONE
```

### DIFF
--- a/api/config/config.go
+++ b/api/config/config.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/kelseyhightower/envconfig"
 	v1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/apimachinery/pkg/api/resource"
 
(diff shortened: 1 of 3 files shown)


api/config/config.go
```diff
@@ -141,12 +141,32 @@ type Resource struct {
 	Memory string `validate:"required"`
 }
 
+func (r *Resource) Decode(value string) error {
```
tiopramayudi, line 144: I'm not sure you need this, can you check whether this method will be called ?

tiopramayudi:
LGTM, thanks @terryyylim

tiopramayudi (approved):

PR #356, opened 2023-03-03
Pin MLflow version to 1.23.0 on pyfunc-server and batch-prediction base image

ariefrahmansyah:

The current implementation of the new base image for batch predictor upgrade the MLflow version to 1.30.0, however, there's a backward incompatibility in the loading model of SparkModel: https://github.com/mlflow/mlflow/pull/5290

The SparkModelCache.get_or_load in mlflow>=1.25.0 returns a tuple instead of a pyfunc object.

To make it more predictable, we also pin mlflow version in pyfunc's base image to 1.23.0 to match merlin-sdk's requirements.

```release-note
NONE
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/batch-predictor/docker/app.Dockerfile
+++ b/python/batch-predictor/docker/app.Dockerfile
@@ -25,5 +25,6 @@ RUN if [[-z "$GOOGLE_APPLICATION_CREDENTIALS"]]; then gcloud auth activate-servi
 RUN gsutil -m cp -r ${MODEL_URL} .
 # pip 20.2.4 to allow dependency conflicts
 RUN /bin/bash -c ". activate ${CONDA_ENVIRONMENT} && \
+    sed -i 's/mlflow$/mlflow==1.23.0/' ${HOME}/model/conda.yaml && \
     conda env update --name ${CONDA_ENVIRONMENT} --file ${HOME}/model/conda.yaml && \
     python ${HOME}/merlin-spark-app/main.py --dry-run-model ${HOME}/model"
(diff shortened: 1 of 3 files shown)


pradithya:
> The current implementation of the new base image for batch predictor upgrade the MLflow version to 1.30.0

Wher do we specify mlflow 1.30.0?

ariefrahmansyah:
> > The current implementation of the new base image for batch predictor upgrade the MLflow version to 1.30.0
> 
> Where do we specify mlflow 1.30.0?

We don't specify it but the mlflow library when logging the model. In conda.yaml, `mlflow` without any version constraint will be added to the dependencies list.

pradithya:
> We don't specify it but the mlflow library when logging the model. In conda.yaml, mlflow without any version constraint will be added to the dependencies list.

Got it. But why don't we see similar problem for pyfunc server?

pradithya (approved):
LGTM

ariefrahmansyah:
> > We don't specify it but the mlflow library when logging the model. In conda.yaml, mlflow without any version constraint will be added to the dependencies list.
> 
> Got it. But why don't we see similar problem for pyfunc server?

Because pyfunc server already uses `pyfunc.load_model()` directly: https://github.com/gojek/merlin/blob/main/python/pyfunc-server/pyfuncserver/model/model.py#L62
Compared to batch predictor that depends on `SparkModelCache.get_or_load()`: https://github.com/gojek/merlin/blob/main/python/batch-predictor/merlinpyspark/model.py#L70

PR #358, opened 2023-03-06
Change authentication method to use OIDC identity tokens 

deadlycoconuts:
As of present, authentication is performed by the Merlin and Turing SDKs using Google OAuth 2.0 access tokens (generated after authentication with Google's OAuth server) that are added to the headers of requests sent to the Turing API server. These tokens are then used by an internal proxy which will validate the access token and then retrieve the user profile information from the Google API. 

However, in keeping with the move of the Google Identity Services client library from OAuth 2.0 access tokens to Google OpenID Connect (OIDC) identity tokens, we are similarly replacing the use of the former for the latter. These tokens generated would similarly be added to the headers of requests sent to the Turing API server to be validated internally. 

A utils module `caraml_auth` in the package `caraml-auth-google` has been created in the CaraML SDK [repository](https://github.com/caraml-dev/caraml-sdk) to provide this new method of authentication.

This PR serves to update the existing authentication method used by the Turing SDK.

The Merlin SDK will import and use the function `get_default_id_token_credentials` provided by the `caraml-auth-google` package:

```python
from caraml_auth.id_token_credentials import get_default_id_token_credentials

credentials = get_default_id_token_credentials(target_audience="sdk.caraml")
credentials.refresh(Request())
```

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/sdk/docs/requirements_docs.txt
+++ b/python/sdk/docs/requirements_docs.txt
@@ -17,7 +17,6 @@ mlflow>=1.2.0,<=1.23.0
 google-cloud-storage>=1.19.0
 urllib3>=1.23
 PyPrind>=2.11.2
-google-auth>=2.16.2
 Click>=7.0
 cloudpickle==2.0.0
 cookiecutter>=1.7.0
--- a/python/sdk/docs/requirements_docs.txt
+++ b/python/sdk/docs/requirements_docs.txt
@@ -17,7 +17,7 @@ mlflow>=1.2.0,<=1.23.0
 google-cloud-storage>=1.19.0
 urllib3>=1.23
 PyPrind>=2.11.2
-google-auth>=1.11.0,<2.0dev
+google-auth>=2.16.2
 Click>=7.0
 cloudpickle==2.0.0
 cookiecutter>=1.7.0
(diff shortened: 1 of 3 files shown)


python/sdk/docs/requirements_docs.txt
```diff
@@ -17,7 +17,7 @@ mlflow>=1.2.0,<=1.23.0
 google-cloud-storage>=1.19.0
 urllib3>=1.23
 PyPrind>=2.11.2
-google-auth>=1.11.0,<2.0dev
+google-auth>=2.16.2
```
krithika369, line 20: Same comment as https://github.com/caraml-dev/turing/pull/324/files#r1134052734

(Also about the `setup.py` file below.)

krithika369 (approved):
Left a small comment. The rest LGTM. Thanks @deadlycoconuts !

python/sdk/docs/requirements_docs.txt
deadlycoconuts: Yep, I think it's redundant too. Just removed it! (In the `setup.py` file too)

deadlycoconuts:
Thanks a lot for the quick review! Will be merging this once the pipeline succeeds!

PR #364, opened 2023-03-14
Using prefix match header for cmux to handle custom codec content-type

tiopramayudi:

Turing uses custom codec when call merlin standard transformer, that change the http2 content-type header by appending subtype from custom codec, so instead `application/grpc` it will use `application/grpc+codec`. This behavior impacting the multiplexing issue (due to we use cmux), since standard transformer relies on `application/grpc` content-type to route the the connection to gRPC server. Hence we change the match header field to use prefix field matching, so `application/grpc+codec` still can be routed to gRPC server

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


ariefrahmansyah (approved):

pradithya (approved):
Thanks @tiopramayudi

PR #370, opened 2023-03-23
Make API server use its environment label in all deployments

deadlycoconuts:
Kubernetes resources deployed by the Merlin API server (Merlin services, batch jobs, etc.) are currently given the resource label `[label-prefix] + "environment": [environment name]`. This environment name is obtained from the version endpoint's environment name (which reflects the name of the cluster in which the resource is deployed in). 

However, this is inconsistent from the intended value of the label, which should instead be the same as the API server's [environment](https://github.com/gojek/merlin/blob/main/api/config/config.go#L38), i.e. `[label-prefix] + "environment": [environment name found in the API server's config]`. As such this PR introduces some minor refactoring to remove change the said environment name, as well as some other minor refactoring of test case variables.

Fixes #

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/batch/resource_test.go
```diff
@@ -26,12 +26,16 @@ import (
 	"github.com/gojek/merlin/models"
 )
 
+const (
+	environmentName  = "dev"
+	orchestratorName = "merlin"
+)
```
deadlycoconuts, line 32: Sprinkled this here and there wherever I happened to see repetition of the environment name or orchestrator name in the tests.

api/batch/resource_test.go
```diff
@@ -119,8 +123,12 @@ var (
...
 
 func TestCreateSparkApplicationResource(t *testing.T) {
-	models.InitKubernetesLabeller("gojek.com/") //nolint:errcheck
-	defer models.InitKubernetesLabeller("")     //nolint:errcheck
+	err := models.InitKubernetesLabeller("gojek.com/", environmentName)
+	assert.NoError(t, err)
+
+	defer func() {
+		_ = models.InitKubernetesLabeller("", "")
+	}()
```
deadlycoconuts, line 131: Did this to make the error checking explicit (also stops any IDE from pointing out that the error is unhandled).

api/cluster/resource/templater_test.go
```diff
@@ -37,6 +37,11 @@ import (
 	"github.com/gojek/merlin/models"
 )
 
+const (
+	environmentName  = "dev"
```
krithika369, line 41: Nit: Shall we prefix these names with `test` so it doesn't somehow interfere with other variables scoped under the `resource` package? Alternatively (and preferably), this package can be named `resource_test` and focus on only testing the public API - but it seems that the tests are currently closely coupled with the private functions, so this can be a future improvement. (The same comment applies to some other test files in this PR too.)

krithika369 (approved):
@deadlycoconuts thanks for this PR and the refactoring! Left one small comment. The rest LGTM. 🚀

api/cluster/resource/templater_test.go
deadlycoconuts: > Shall we prefix these names with test so it doesn't somehow interfere with other variables scoped under the `resource` package?
Ahh good point, I'll do this then; but yes, seems like separating the packages will be a good idea in the long run.

deadlycoconuts:
Thanks for the quick review! Will be merging this when the pipeline passes!

PR #373, opened 2023-04-03
Add Inference Logger

terryyylim:

This PR introduces Merlin log collector which intercepts all request/response going to/from model and transformer, and submitting the payload to a logging sink, where Kafka and NewRelic is currently supported.

`NONE`

```release-note
NONE
```

**Checklist**

- [x] Tested locally


ariefrahmansyah:
What's the reasoning behind placing `inference-logger` in a new `./pkg` directory?

What do you think about following the standard transformer approach? Which is:
1. Has their Golang code in `api` folder:
   a. Its main file is in `api/cmd/transformer`
   b. Puts their logic code in `api/pkg/transformer`
2. Has transformer.Dockerfile in project's root
3. Has it's own docker build and publish jobs

ariefrahmansyah (approved):
lgtm. thanks @terryyylim !

PR #375, opened 2023-04-04
Replace id token with a more realistic dummy token

deadlycoconuts:
With the change in #358 to make the Merlin SDK use ID tokens for all requests made to the Merlin API, there is a need to update the dummy tokens returned by the mocked Google Identity Services (since we would now expect ID tokens to be returned from them instead of access tokens). 

If these are not updated, we would expect the SDK tests that involve initialising a Merlin client to fail since the returned credentials would contain an access token instead of an ID token. This creates errors in instances like [these](https://github.com/caraml-dev/merlin/blob/cddaf6c3b56cb2ef2adedc28389a75f64d95e270/python/sdk/test/client_test.py#L199), which use a mocked `mock_oauth` [response](https://github.com/caraml-dev/merlin/blob/cddaf6c3b56cb2ef2adedc28389a75f64d95e270/python/sdk/test/client_test.py#L364) that returns credentials with an access token. These errors would then surface as messages such as `google.auth.exceptions.MalformedError: Wrong number of segments in token: b'thisisnotarealtoken'` when the client initialisation fails.

Fixes #

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/sdk/test/client_test.py
+++ b/python/sdk/test/client_test.py
@@ -361,7 +361,7 @@ def mock_oauth():
         "expires_in": 3600,
         "scope": "openid https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/userinfo.email",
         "token_type": "Bearer",
-        "id_token": "thisisnotarealtoken"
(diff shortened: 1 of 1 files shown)


ariefrahmansyah (approved):
LGTM! Thanks, @deadlycoconuts!

PR #376, opened 2023-04-05
Bump Go version to 1.20

terryyylim:

Update Go version to `1.20` prior to integrating inference-logger.

`NONE`

```release-note
NONE
```

### DIFF
--- a/.github/workflows/merlin.yml
+++ b/.github/workflows/merlin.yml
@@ -10,7 +10,7 @@ on:
 env:
   ARTIFACT_RETENTION_DAYS: 7
   DOCKER_BUILDKIT: 1
-  GO_VERSION: 1.18
+  GO_VERSION: '1.20'
 
 jobs:
   create-version:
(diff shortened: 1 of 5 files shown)


.github/workflows/merlin.yml
```diff
@@ -10,7 +10,7 @@ on:
 env:
   ARTIFACT_RETENTION_DAYS: 7
   DOCKER_BUILDKIT: 1
-  GO_VERSION: 1.18
+  GO_VERSION: '1.20'
```
krithika369, line 13: Why the single quotes?

krithika369 (approved):
LGTM. Thanks!

.github/workflows/merlin.yml
terryyylim: It'll be truncated to 1.2 without it and the linting step will fail.

krithika369: Ahh ok.

PR #377, opened 2023-04-06
Disallow changing deployment mode for pending/running endpoints

krithika369:

This PR fixes an issue with switching the deployment mode of a model version, where the K8s resources are not cleaned up properly.

When switching the deployment mode of an existing (running) deployment from serverless -> raw or vice versa, Kserve uses slightly different naming conventions for the underlying deployments and they are not cleaned up properly. Eg:

```
NAME                                                        READY   UP-TO-DATE   AVAILABLE   AGE
tensorflow-sample-s-27-predictor-default                    1/1     1            1           50s
tensorflow-sample-s-27-predictor-default-00001-deployment   1/1     1            1           97s
```

We are already preventing changing the deployment mode of a serving endpoint because it cannot be done gracefully. The same argument can be extended to a running endpoint as well. In addition, the issue with the deployment resource not being cleaned up can be circumvented in the same way. Thus, in all, we now block "updating" an endpoint in any of the 3 states:
* Pending
* Running
* Serving

Fixes #

```release-note
Changed: Deployment mode of model endpoints that are in pending / running state can no longer be changed. Previously, this rules was only applicable to model endpoints in serving state.
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


ariefrahmansyah (approved):
lgtm! thanks, krithika!

PR #380, opened 2023-04-17
Improve Image Builder Jobs Label and Annotation

ariefrahmansyah:

The image-building jobs in Merlin are timing out. After some investigation, we found that one of the root causes is the node pool got scaled down resulting in the image building pods to be rescheduled.

This PR adds "cluster-autoscaler.kubernetes.io/safe-to-evict": "false" to avoid the pod get killed and rescheduled.

In Refactor deployed model labels #346, we managed to propagate user labels to models and batch prediction jobs, but not to image builder jobs (here). This PR includes adding project + model version labels to image builder job.

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cmd/api/setup.go
+++ b/api/cmd/api/setup.go
@@ -144,6 +144,7 @@ func initImageBuilder(cfg *config.Config) (webserviceBuilder imagebuilder.ImageB
 		Tolerations:          cfg.ImageBuilderConfig.Tolerations,
 		NodeSelectors:        cfg.ImageBuilderConfig.NodeSelectors,
 		MaximumRetry:         cfg.ImageBuilderConfig.MaximumRetry,
+		JobSafeToEvict:       cfg.ImageBuilderConfig.JobSafeToEvict,
 
 		ClusterName: cfg.ImageBuilderConfig.ClusterName,
 		GcpProject:  cfg.ImageBuilderConfig.GcpProject,
@@ -164,6 +165,7 @@ func initImageBuilder(cfg *config.Config) (webserviceBuilder imagebuilder.ImageB
 		Tolerations:          cfg.ImageBuilderConfig.Tolerations,
 		NodeSelectors:        cfg.ImageBuilderConfig.NodeSelectors,
 		MaximumRetry:         cfg.ImageBuilderConfig.MaximumRetry,
+		JobSafeToEvict:       cfg.ImageBuilderConfig.JobSafeToEvict,
 
 		ClusterName: cfg.ImageBuilderConfig.ClusterName,
 		GcpProject:  cfg.ImageBuilderConfig.GcpProject,
--- a/api/pkg/imagebuilder/imagebuilder.go
+++ b/api/pkg/imagebuilder/imagebuilder.go
@@ -336,6 +336,11 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
 		Component: models.ComponentImageBuilder,
 		Stream:    project.Stream,
 		Team:      project.Team,
+		Labels:    models.MergeProjectVersionLabels(project.Labels, version.Labels),
+	}
+
+	annotations := map[string]string{
+		"cluster-autoscaler.kubernetes.io/safe-to-evict": fmt.Sprint(c.config.JobSafeToEvict),
 	}
 
 	baseImageTag, ok := c.config.BaseImages[version.PythonVersion]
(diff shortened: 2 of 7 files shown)


api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -336,6 +336,11 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
 		Component: models.ComponentImageBuilder,
 		Stream:    project.Stream,
 		Team:      project.Team,
+		Labels:    models.MergeProjectVersionLabels(project.Labels, version.Labels),
+	}
+
+	annotations := map[string]string{
+		"cluster-autoscaler.kubernetes.io/safe-to-evict": fmt.Sprint(c.config.JobSafeToEvict),
```
krithika369, line 343: Should we conditionally set this label only when it's supposed to be `"false"` ? Though we don't currently plan to use the value `"true"`, I'm afraid that if we do so at some point in the future, it will make it more prone to eviction than usual (doc: https://kubernetes.io/docs/reference/labels-annotations-taints/#cluster-autoscaler-kubernetes-io-safe-to-evict).

scripts/e2e/config/kserve/overlay.yaml
```diff
@@ -1,24 +1,28 @@
...
-            cpu: 1
-            memory: 300Mi
-          requests:
-            cpu: 100m
-            memory: 64Mi
-
+        - name: manager
+          resources:
+            limits:
+              cpu: ~
```
krithika369, line 20: What is the significance of this symbol, and why is this change made?

scripts/e2e/config/kserve/overlay.yaml
```diff
@@ -1,24 +1,28 @@
 apiVersion: apps/v1
 kind: StatefulSet
+metadata:
+  name: kserve-controller-manager
+  namespace: kserve
+$patch: delete
```
krithika369, line 6: Not too familiar with kustomize but is this delete operation intended? What are we trying to delete?

scripts/e2e/config/kserve/overlay.yaml
ariefrahmansyah: In YAML, tilde means NULL value [1]. So in Kustomize, it means we remove resources.limits.cpu

https://stackoverflow.com/questions/51990175/what-is-the-purpose-of-tilde-character-in-yaml

scripts/e2e/config/kserve/overlay.yaml
ariefrahmansyah: This is intended. Kserve 0.9.0 allows us to deploy kserve-controller-manager as deployment other than statefulset. However, their released v0.9.0 manifest has both of deployment and stafulset of kserve-controller-manager which needs more editing from users to delete one of them. Hence, by using this kustomize's $patch: delete, I delete statefulset and just use deployment. I have updated PR's description for this.

api/pkg/imagebuilder/imagebuilder.go
ariefrahmansyah: Ah, TIL. Thanks!

api/cmd/api/setup.go
```diff
@@ -144,6 +144,7 @@ func initImageBuilder(cfg *config.Config) (webserviceBuilder imagebuilder.ImageB
 		Tolerations:          cfg.ImageBuilderConfig.Tolerations,
 		NodeSelectors:        cfg.ImageBuilderConfig.NodeSelectors,
 		MaximumRetry:         cfg.ImageBuilderConfig.MaximumRetry,
+		JobSafeToEvict:       cfg.ImageBuilderConfig.JobSafeToEvict,
```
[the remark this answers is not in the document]
ariefrahmansyah: Got it. I understand your confusion. I have updated this part.

scripts/e2e/config/kserve/overlay.yaml
krithika369: Thanks, Arief!

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -336,6 +336,12 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
 		Component: models.ComponentImageBuilder,
 		Stream:    project.Stream,
 		Team:      project.Team,
+		Labels:    models.MergeProjectVersionLabels(project.Labels, version.Labels),
+	}
+
+	annotations := make(map[string]string)
+	if !c.config.SafeToEvict {
+		annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"] = fmt.Sprint(c.config.SafeToEvict)
```
krithika369, line 344: Super nit: this can just be:

```golang
annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"] = "false"
```

Also, perhaps we can leave an inline comment on why we add this annotation conditionally, linking to the [doc](https://kubernetes.io/docs/reference/labels-annotations-taints/#cluster-autoscaler-kubernetes-io-safe-to-evict)?

krithika369 (approved):
LGTM. Thanks, @ariefrahmansyah !

PR #382, opened 2023-04-18
Add topology spread constraints to deployments

deadlycoconuts:
Similar to how https://github.com/caraml-dev/turing/pull/332 allows an operator to define topology spread constraints for pods deployed by the API server, this PR also introduces changes the Merlin API server operator to specify topology spread constraints within the `environmentConfigs` of its chart values `.yaml` file, that will be propagated to all the model service pods deployed via the the API server: 
```yaml
environmentConfigs:
  - name: "merlin-dev"
    is_default: true
    cluster: "dev-cluster"
    deployment_timeout: "10m"
    namespace_timeout: "2m"
    max_cpu: "8"
    max_memory: "8Gi"
    topology_spread_constraints:
      - maxSkew: 1
        topologyKey: kubernetes.io/hostname
        whenUnsatisfiable: ScheduleAnyway
   ...
```

Similarly,  in order to encourage pods of the same Knative service deployed by the Merlin API server to be scheduled on a largest variety of nodes wherever possible to encourage high availability, an additional match label is added by default to all constraints specified in the Turing API configuration:
```yaml
matchLabels:
    app: [knative-service-name]
```
The above behaviour occurs for all constraints, even those without any other match labels or label selectors specified.

### Some extra details
Since Merlin model service deployments (both predictors and transformers) are deployed as KServe `InferenceService`-s, which themselves are built upon Knative services (see [here](https://kserve.github.io/website/0.10/get_started/first_isvc/) for more info) that can be either of the `Serverless` or `RawDeployment` type, both of which generate different values for the `metadata.labels.app` field (label) in the pod configuration (pod manifest file), based on the `InfererenceService` names:

`Serverless` type:
- `InferenceService` name: `sklearn-sample`
  Autogenerated label: `app=sklearn-sample-1-predictor-default-00001`
  Schema: `[isvc name]-[predictor/transformer]-default-[6 digit 0-padded revision number]`

`RawDeployment` type:
- `InferenceService` name: `sklearn-sample`
  Autogenerated label: `isvc.sklearn-sample-1-predictor-default`
  Schema: `isvc.[isvc name]-[predictor/transformer]-default`

where the `isvc name` includes the version number (of the model service) automatically [generated](https://github.com/caraml-dev/merlin/blob/d455cad616a89093c7db81de9580324f0cf64fe5/api/models/service.go#L52) by the API server.

Hence separate methods are defined to handle these two cases differently. Note in particular that any model redeployment of a `RawDeployment`-type `InferenceService` does not change the `app` label value, though it would change a `Serverless`-type's label, via the increment in the Knative service revision number. As such, whenever a `Severless` deployment is redeployed, we need to determine the to-be-deployed revision number (that will be automatically created by the Knative controller/webhook) by retrieving the `LatestReadyRevision` of the existing deployment and then incrementing it by 1.

This is slightly different from how the Turing API server sets the `app` label value to be exactly the same as that of the Knative service name (the Turing API server deploys routers directly as Knative services).

Fixes #

```release-note
NONE
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cluster/resource/templater.go
+++ b/api/cluster/resource/templater.go
@@ -130,7 +130,7 @@ func (t *InferenceServiceTemplater) CreateInferenceServiceSpec(modelService *mod
 	}
 
 	predictorSpec := createPredictorSpec(modelService, config)
-	predictorTopologySpreadConstraints, err := createNewInferenceServiceTopologySpreadConstraints(
+	predictorSpec.TopologySpreadConstraints, err = createNewInferenceServiceTopologySpreadConstraints(
 		modelService,
 		config,
 		kservev1beta1.PredictorComponent,
@@ -138,7 +138,6 @@ func (t *InferenceServiceTemplater) CreateInferenceServiceSpec(modelService *mod
 	if err != nil {
 		return nil, fmt.Errorf("unable to create predictor topology spread constraints: %w", err)
 	}
-	predictorSpec.TopologySpreadConstraints = predictorTopologySpreadConstraints
 
 	inferenceService := &kservev1beta1.InferenceService{
 		ObjectMeta: objectMeta,
@@ -152,7 +151,7 @@ func (t *InferenceServiceTemplater) CreateInferenceServiceSpec(modelService *mod
 		if err != nil {
 			return nil, fmt.Errorf("unable to create transformer spec: %w", err)
 		}
-		transformerTopologySpreadConstraints, err := createNewInferenceServiceTopologySpreadConstraints(
+		inferenceService.Spec.Transformer.TopologySpreadConstraints, err = createNewInferenceServiceTopologySpreadConstraints(
 			modelService,
 			config,
 			kservev1beta1.TransformerComponent,
@@ -160,7 +159,6 @@ func (t *InferenceServiceTemplater) CreateInferenceServiceSpec(modelService *mod
 		if err != nil {
 			return nil, fmt.Errorf("unable to create transformer topology spread constraints: %w", err)
 		}
-		inferenceService.Spec.Transformer.TopologySpreadConstraints = transformerTopologySpreadConstraints
 	}
 
 	return inferenceService, nil
@@ -175,12 +173,11 @@ func (t *InferenceServiceTemplater) PatchInferenceServiceSpec(orig *kservev1beta
 		return nil, fmt.Errorf("unable to patch inference service spec: %w", err)
 	}
 	orig.ObjectMeta.Annotations = utils.MergeMaps(utils.ExcludeKeys(orig.ObjectMeta.Annotations, configAnnotationKeys), annotations)
-	predictor := createPredictorSpec(modelService, config)
+	orig.Spec.Predictor = createPredictorSpec(modelService, config)
 	if err != nil {
 		return nil, fmt.Errorf("unable to create predictor spec: %w", err)
 	}
-
-	predictor.TopologySpreadConstraints, err = updateExistingInferenceServiceTopologySpreadConstraints(
+	orig.Spec.Predictor.TopologySpreadConstraints, err = updateExistingInferenceServiceTopologySpreadConstraints(
 		orig,
 		config,
 		kservev1beta1.PredictorComponent,
@@ -189,15 +186,13 @@ func (t *InferenceServiceTemplater) PatchInferenceServiceSpec(orig *kservev1beta
 		return nil, fmt.Errorf("unable to create predictor topology spread constraints: %w", err)
 	}
 
-	orig.Spec.Predictor = predictor
-
 	orig.Spec.Transformer = nil
 	if modelService.Transformer != nil && modelService.Transformer.Enabled {
 		orig.Spec.Transformer = t.createTransformerSpec(modelService, modelService.Transformer, config)
 		if err != nil {
 			return nil, fmt.Errorf("unable to create transformer spec: %w", err)
 		}
-		transformerTopologySpreadConstraints, err := updateExistingInferenceServiceTopologySpreadConstraints(
+		orig.Spec.Transformer.TopologySpreadConstraints, err = updateExistingInferenceServiceTopologySpreadConstraints(
 			orig,
 			config,
 			kservev1beta1.TransformerComponent,
@@ -205,7 +200,6 @@ func (t *InferenceServiceTemplater) PatchInferenceServiceSpec(orig *kservev1beta
 		if err != nil {
 			return nil, fmt.Errorf("unable to create transformer topology spread constraints: %w", err)
 		}
-		orig.Spec.Transformer.TopologySpreadConstraints = transformerTopologySpreadConstraints
 	}
 	return orig, nil
 }
--- a/api/cluster/resource/templater.go
+++ b/api/cluster/resource/templater.go
@@ -640,7 +640,7 @@ func updateExistingInferenceServiceTopologySpreadConstraints(
 	} else if modelService.DeploymentMode == deployment.ServerlessDeploymentMode ||
 		modelService.DeploymentMode == deployment.EmptyDeploymentMode {
 		var err error
-		newRevisionName, err = getNewRevisionNameForExistingSeverlessDeployment(
+		newRevisionName, err = getNewRevisionNameForExistingServerlessDeployment(
 			orig.Status.Components[component].LatestCreatedRevision,
 		)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


api/cluster/resource/templater.go
```diff
@@ -288,7 +334,7 @@ func createPredictorSpec(modelService *models.Service, config *config.Deployment
 	return predictorSpec
 }
 
-func (t *InferenceServiceTemplater) createTransformerSpec(modelService *models.Service, transformer *models.Transformer, config *config.DeploymentConfig) *kservev1beta1.TransformerSpec {
+func (t *InferenceServiceTemplater) createTransformerSpec(modelService *models.Service, transformer *models.Transformer) *kservev1beta1.TransformerSpec {
```
deadlycoconuts, line 337: Removed the `config` argument from the signature since it is unused.

api/cluster/resource/templater.go
```diff
@@ -550,6 +596,120 @@ func createLoggerSpec(loggerURL string, loggerConfig models.LoggerConfig) *kserv
...
+) ([]corev1.TopologySpreadConstraint, error) {
+	if len(config.TopologySpreadConstraints) == 0 {
+		var topologySpreadConstraints []corev1.TopologySpreadConstraint
+		return topologySpreadConstraints, nil
+	}
+	var newRevisionName string
+	if modelService.DeploymentMode == deployment.RawDeploymentMode {
+		newRevisionName = fmt.Sprintf("isvc.%s-%s-default", modelService.Name, component)
+	} else if modelService.DeploymentMode == deployment.ServerlessDeploymentMode ||
+		modelService.DeploymentMode == deployment.EmptyDeploymentMode {
```
deadlycoconuts, line 614: I've specified the `EmptyDeploymentMode` here too since the existing unit tests seem to suggest that we're expecting the behaviour of a `ServerlessDeploymentMode` if nothing (empty string `""`) is specified as the `DeploymentMode`.

api/config/environment.go
```diff
@@ -36,8 +38,9 @@ type EnvironmentConfig struct {
 	DeploymentTimeout time.Duration `yaml:"deployment_timeout"`
 	NamespaceTimeout  time.Duration `yaml:"namespace_timeout"`
 
-	MaxCPU    string `yaml:"max_cpu"`
-	MaxMemory string `yaml:"max_memory"`
+	MaxCPU                    string                    `yaml:"max_cpu"`
+	MaxMemory                 string                    `yaml:"max_memory"`
+	TopologySpreadConstraints TopologySpreadConstraints `yaml:"topology_spread_constraints"`
```
krithika369, line 43: Just realised that in Merlin, this is going to be a part of each environment's individual config. Whereas in Turing, it is a common deployment config: https://github.com/caraml-dev/turing/blob/main/api/turing/config/config.go#L233

Something we can think about unifying, when we move the environments to MLP.

api/cluster/resource/templater.go
```diff
@@ -550,6 +596,120 @@ func createLoggerSpec(loggerURL string, loggerConfig models.LoggerConfig) *kserv
...
+		var err error
+		newRevisionName, err = getNewRevisionNameForExistingSeverlessDeployment(
+			orig.Status.Components[component].LatestReadyRevision,
```
krithika369, line 644: Hmm.. What if there was a failed deployment or one in progress? Should we get the latest "created" revision as opposed to ready ?

```
      Latest Created Revision:      std-preprocess-loadtest-3-predictor-default-00002
      Latest Ready Revision:        std-preprocess-loadtest-3-predictor-default-00002
      Latest Rolledout Revision:    std-preprocess-loadtest-3-predictor-default-00002
      Previous Rolledout Revision:  std-preprocess-loadtest-3-predictor-default-00001
```

krithika369: @deadlycoconuts  / @ariefrahmansyah any thoughts on this?

deadlycoconuts: Ah oops my bad; I totally forgot about this. Hmm so if I understand correctly, the latest created revision >= latest rolled out revision and we trying to be as 'safe' as possible by picking the largest possible revision number to increment right? I think that sounds good. We can monitor the behaviour and if it ever creates any bugs/unexpected behaviour, change the implementation later on :)

api/config/environment.go
deadlycoconuts: Yep, the structure of the configs in both apps can be unified and made more consistent 😅

api/cluster/resource/templater.go
```diff
@@ -550,6 +596,120 @@ func createLoggerSpec(loggerURL string, loggerConfig models.LoggerConfig) *kserv
...
+		return nil, fmt.Errorf("Error in type assertion of copied topology spread constraints interface: %w", err)
+	}
+	return topologySpreadConstraints, nil
+}
+
+// getNewRevisionNameForExistingSeverlessDeployment examines the current revision name of an inference service (
+// serverless deployment) app name that is given to it and increments the last value of the revision number by 1, e.g.
+// sklearn-sample-predictor-default-00001 -> sklearn-sample-predictor-default-00002
+func getNewRevisionNameForExistingSeverlessDeployment(currentRevisionName string) (string, error) {
+	revisionNameElements := strings.Split(currentRevisionName, "-")
```
krithika369, line 703: Should we handle the possibility of this split resulting in less than 2 parts (then, the `-1` in the next line would result in an out of bounds panic).

krithika369 (approved):
I left a small comment - the rest LGTM. Thanks for the thorough test cases as well. 🚀

api/cluster/resource/templater.go
deadlycoconuts: Hmm in that case should we just perform a check that `len(revisionNameElements) < 4`? If it's less than 2 the name's probably going to be nonsensical anyway and we wouldn't know how to handle that. I chose 4 because the revision name's expected to be made up of at least 4 parts `[isvc name]-[predictor/transformer]-default-[6 digit 0-padded revision number]`. There may be more parts if the isvc name has additional hyphens.

api/cluster/resource/templater.go
```diff
@@ -152,10 +174,34 @@ func (t *InferenceServiceTemplater) PatchInferenceServiceSpec(orig *kservev1beta
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create predictor topology spread constraints: %w", err)
+	}
 
 	orig.Spec.Transformer = nil
 	if modelService.Transformer != nil && modelService.Transformer.Enabled {
-		orig.Spec.Transformer = t.createTransformerSpec(modelService, modelService.Transformer, config)
+		orig.Spec.Transformer = t.createTransformerSpec(modelService, modelService.Transformer)
+		if err != nil {
```
ariefrahmansyah, line 193: We can remove this since t.createTransformerSpec does not return error

api/cluster/resource/templater.go
```diff
@@ -152,10 +174,34 @@ func (t *InferenceServiceTemplater) PatchInferenceServiceSpec(orig *kservev1beta
 	}
 	orig.ObjectMeta.Annotations = utils.MergeMaps(utils.ExcludeKeys(orig.ObjectMeta.Annotations, configAnnotationKeys), annotations)
 	orig.Spec.Predictor = createPredictorSpec(modelService, config)
+	if err != nil {
```
ariefrahmansyah, line 177: We can remove this since createPredictorSpec does not return error

api/cluster/resource/templater.go
deadlycoconuts: This probably isn't a perfect check either since someone might just happen to create a model with more hyphens than necessary in the `isvc name` , e.g. `sdk-test-1`, and the extra hyphens can count towards the length check (i.e. a revision name with `sdk-test-1-invalid-000001` would still pass the check).

ariefrahmansyah: nit: can we add unit test for this function?

ariefrahmansyah: Also, I'm not a huge fan of regex, but I think regex is more suitable here.

api/cluster/resource/templater.go
deadlycoconuts: Oh good catch; I think the error handling was a remnant of a previous iteration of this PR 😅 I forgot to clean those up after I changed the implementation. Thanks for picking this up!

api/cluster/resource/templater.go
deadlycoconuts: Same comment as above! Thanks again!

krithika369:
Will merge this PR.

PR #383, opened 2023-04-19
Fix regression when redeploying UPI model

pradithya:

Redeploying an existing UPI model causes the UI to crash due to the value of `predictionLogger` passed to `PredictionLoggerPanel` being `null`. `predictionLogger` is a new field introduced in https://github.com/caraml-dev/merlin/pull/361 and existing deployed model doesn't have the property. 
This PR adds guard to check whether the value is null before accessing its internal properties . 

```release-note
None
```

**Checklist**

- [X] Tested locally


deadlycoconuts (approved):
LGTM! 🚀

ariefrahmansyah (approved):

PR #385, opened 2023-04-25
[UI] Replace reach-router with react-router-dom

terryyylim:

In order to upgrade MLP dependency to reference shifted `caraml-dev/mlp` package, we'll need to first remove usage of `@reach/router` in order to use new components such as `CurrentProjectContextProvider` which has been removed from latest `@caraml-dev/ui-lib`.

`N/A`

```release-note
NONE
```

**Checklist**

- [x] Tested locally


mlflow/Dockerfile
```diff
@@ -1,12 +1,11 @@
-FROM python:3.7.2-slim
+FROM python:3.8-slim
```
ariefrahmansyah, line 1: Let's use 3.7-slim here

ariefrahmansyah (approved):
LGTM! Thanks @terryyylim!

PR #386, opened 2023-05-02
Add missing flags for logger

ariefrahmansyah:

When enabling logger, Kserve adds some flags to the logger container. But these flags are not registered by logger app, hence raising error.

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cmd/inference-logger/main.go
+++ b/api/cmd/inference-logger/main.go
@@ -38,6 +38,11 @@ var (
 	logMode          = flag.String("log-mode", string(merlinlogger.LogModeAll), "Whether to log 'request', 'response' or 'all'")
 	inferenceService = flag.String("inference-service", "my-model-1", "The InferenceService name to add as header to log events")
 	namespace        = flag.String("namespace", "my-project", "The namespace to add as header to log events")
+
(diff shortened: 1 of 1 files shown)


api/cmd/inference-logger/main.go
```diff
@@ -38,6 +38,11 @@ var (
 	logMode          = flag.String("log-mode", string(merlinlogger.LogModeAll), "Whether to log 'request', 'response' or 'all'")
 	inferenceService = flag.String("inference-service", "my-model-1", "The InferenceService name to add as header to log events")
 	namespace        = flag.String("namespace", "my-project", "The namespace to add as header to log events")
+
```
krithika369, line 41: On this note, it seems kserve is not passing the `port` and `workers` flags. Do we need them?

krithika369 (approved):
Left one q. LGTM. Thanks, @ariefrahmansyah !

api/cmd/inference-logger/main.go
ariefrahmansyah: We don't need them and kserve also doesn't pass them. However, they didn't raise the error because they have default value configured. I'm keeping them for now as it would introduce more changes and we can ignore them.

PR #389, opened 2023-05-04
Feat/delete api

zenovore:
**What this PR does / why we need it**:  Allowing User to delete model and model version
## Delete Feature
This PR contain the API to allow user to delete model, model version and prediction job.
Deleting an model version have several constraint that must be met, including:
- For `pyfunc_v2` model 
  - No `pending` or `running` prediction jobs that use the model version
  - No  `pending`, `running`, or `serving` endpoints that use the model version
- For non `pyfunc_v2` model 
  - No  `pending`, `running`, or `serving` endpoints that use the model version

If the model version can be deleted, this feature will also delete other related entities such as:
- For `pyfunc_v2` model
  - Prediction job with status `failed`, `failed_submission`, `completed`, and `terminated`
  - Endpoint with status `failed`, `terminated`
- For non `pyfunc_v2`
  - Endpoint with status `failed`, `terminated`

A popup will appear showing the entity that is blocking the deletion process (if the constraints have not been met) and the entities that will be deleted if the constraints are met. Even though that the API is facing some changes on checking the endpoint for the `pyfunc_v2` model, the ui wont check the endpoint status, since current implementation prevent the user for deploying the `pyfunc_v2` model on the frontend.

## Delete Prediction Jobs
This PR change on how the implementation of delete prediction jobs work, instead of changing the status to `terminated` on `StopJob`, now the `StopPredictionJob` function will also delete the job data from the `spark` and from the `database`. If the job **doesn't have spark application**, then instead of throwing an error, the deletion process will proceed with deleting the data from the database

**Does this PR introduce a user-facing change?**: No

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/api/models_api.go
```diff
@@ -100,3 +105,157 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+	if err != nil {
+		return InternalServerError(err.Error())
+	}
+
+	var allJobInModel []*models.PredictionJob
+	var allEndpointInModel []*models.VersionEndpoint
+
+	// iterate for every version, check the active prediction job / active endpoint
+	for _, version := range versions {
+		if model.Type == "pyfunc_v2" {
```
krithika369, line 140: @ariefrahmansyah, just to confirm, could you clarify if `pyfunc_v2` can only be used for batch jobs? Or is it also possible to use them for real-time models?

api/api/versions_api.go
```diff
@@ -145,6 +146,108 @@ func (c *VersionsController) CreateVersion(r *http.Request, vars map[string]stri
...
+	model, version, err := c.getModelAndVersion(ctx, modelID, versionID)
+	if err != nil {
+		if !gorm.IsRecordNotFoundError(err) {
+			return InternalServerError(err.Error())
+		}
+		return NotFound(err.Error())
+	}
+
+	// handle for pyfunc v2, since the prediction job feature is only available for pyfunc_v2 model
+	if model.Type == "pyfunc_v2" {
```
krithika369, line 164: Since much of this logic (and `checkActivePredictionJobs` and `checkActiveEndpoints`) is largely this same as with the model delete, can we extract it into a private method of the `VersionsController` and let model delete call it too? In this case, the `ModelsController` will hold a reference to the `VersionsController` like:

```golang
// ModelsController controls models API.
type ModelsController struct {
	*AppContext
        *VersionsController
}

func (c *ModelsController) DeleteModel(r *http.Request, vars map[string]string, body interface{}) *Response {
    for _, version := range versions {
            // This function can be defined in versions_api.go and handle
            // deletion of a single version / mlflow run
            err := c.VersionsController.deleteVersion(version)
            if err != nil {
                    // Handle type of error and return Response 
            }
    }

    // Delete mlflow experiment and artifact
    // ...

    // Delete model data from DB
    // ...

    return Ok(model.ID)
}
```

Wdyt?

krithika369 (commented):
Thanks for the PR and the detailed unit tests, @zenovore ! The changes look great. :)

I left some comments. Will let @ariefrahmansyah / @deadlycoconuts have a look as well.

api/api/versions_api.go
```diff
@@ -145,6 +146,108 @@ func (c *VersionsController) CreateVersion(r *http.Request, vars map[string]stri
...
+			return 400, nil, fmt.Errorf("There are active prediction job that still using this model version")
+		}
+	}
+
+	return 200, jobs, nil
+}
+
+func (c *VersionsController) checkActiveEndpoints(ctx context.Context, model *models.Model, version *models.Version) (int, []*models.VersionEndpoint, error) {
+
+	endpoints, err := c.EndpointsService.ListEndpoints(ctx, model, version)
```
krithika369, line 237: In order to reduce the latency of this API, rather than getting all endpoints and iterating over them, it will be useful if we add some filter options to the `ListEndpoints` method, so we can directly query the DB for active endpoints. Better yet, we can have an API to only `CountEndpoints` so there is no need to fetch the actual data yet. Similarly, there can be a `DeleteEndpoints` API that accepts the `model_id` and `version_id`.

Does it make sense?

Similar comment for the predictions jobs counting. These needn't be tackled in this PR right away - they can be in a new PR.

api/api/models_api.go
```diff
@@ -100,3 +105,157 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+	}
+	if model.Type != "pyfunc_v2" {
+		// undeploy all existing model endpoint
+		modelEndpoints, err := c.ModelEndpointsService.ListModelEndpoints(ctx, modelID)
+		if err != nil {
+			return InternalServerError(err.Error())
+		}
+
+		for _, modelEndpoint := range modelEndpoints {
+			_, err := c.ModelEndpointsService.UndeployEndpoint(ctx, model, modelEndpoint)
```
deadlycoconuts, line 202: Hmm it seems like `UndeployEndpoint` will only call `modelEndpointStorage.Save`, which only [updates](https://github.com/caraml-dev/merlin/blob/d0d1ecdda927482b666ca189662c5419b16169a2/api/service/model_endpoint_service.go#L204) the model endpoint in the endpoints database instead of actually deleting it completely, should there instead be a new function for the `ModelEndpointStorage` interface that's actually implemented to `Delete` endpoints? 🤔 Interestingly, the `EndpointsService.DeleteEndpoint` being called above actually performs a `Delete` operation on the version endpoints database.

A slightly unrelated question, but if all the version endpoints have already been deleted (or undeployed rather) by this point, is it still possible that a model endpoint still exists? Or is this section redundant (there is logically no way a model endpoint can still exist if all the version endpoints have been deleted)?

api/batch/controller.go
```diff
@@ -285,6 +285,9 @@ func (c *controller) syncStatus(ctx context.Context, key string) error {
 		c.cleanup(ctx, predictionJob, sparkApp.Namespace)
 		modelName := getModelName(predictionJob.Name)
 		BatchCounter.WithLabelValues(sparkApp.Namespace, modelName, string(predictionJob.Status)).Inc()
+		if predictionJob.Status == models.JobTerminated {
+			return c.store.Delete(predictionJob)
+		}
```
deadlycoconuts, line 290: Is this a new behaviour we're implementing? i.e. a job gets deleted automatically and immediately after it's terminated?

deadlycoconuts (commented):
Thanks a lot for this PR! Krithika seems to have already left some comments so I've just added 2 more! In general it's looking good though I agree that we can perhaps consider refactoring the model version deletion logic into some method that can be reused when deleting a model :)

api/batch/controller.go
zenovore: yes, i was thinking the same implementation from turing, the runner will delete any job that in terminated state

api/service/prediction_job_service.go
```diff
@@ -173,6 +173,11 @@ func (p *predictionJobService) StopPredictionJob(ctx context.Context, env *model
 		return nil, err
 	}
 
+	if job.Status.IsTerminal() && job.Status != models.JobTerminated {
```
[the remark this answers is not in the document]
zenovore: because if we let the ctl.Stop handle some of the terminal status (failed, failed_submission) which doesnt have any spark application, this leads to error when we delete the job

api/api/versions_api.go
deadlycoconuts: I think it's safer to perform first a check on all the versions (in a loop) to determine if there are any versions that have active endpoint/prediction jobs. If none of them are active, we can then proceed to delete them in a loop.

api/service/prediction_job_service.go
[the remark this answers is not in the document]
deadlycoconuts: Hmm where this this change go? Why was this removed? 🤔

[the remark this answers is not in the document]
zenovore: Update: i kinda changed the implementation, since the runner is updating for every job, lets just use ctl.Stop to do the deletion, the implementation is kinda different than turing, so when user are terminating the job, it will immediately get deleted, the runner on merlin is working differently than turing, since the worker only works if the spark application is existing (the worker function is to update job status from the spark application status)

The current implementation will immediately cleanup the spark application + do the deletion from the database

api/api/models_api.go
zenovore: oops, sorry for missing this out, yes, it should be DeleteModelEndpoint, since the prerequisite of model deletion is no more active deployment, so we can just immediately delete it from the db. I already update it so it will immediately delete the model endpoint data from the database.

deadlycoconuts: Sounds good! Thanks for making that change!

api/batch/controller.go
deadlycoconuts: @krithika369 @ariefrahmansyah, there should be no concerns with this change right? The only reason we used to leave terminated jobs dangling in the DB was for debugging?

api/service/prediction_job_service.go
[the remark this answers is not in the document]
deadlycoconuts: That makes sense, thanks for the changes!

api/batch/controller.go
zenovore: Since i already change the implementation to [this](https://github.com/caraml-dev/merlin/pull/389#discussion_r1236503432), the runner now doesn't delete the terminated jobs, since the StopFunction will immediately Delete the jobs. But still, is it okay if i do this? wdyt?

deadlycoconuts: So if I understand correctly, these are the two workflows for stopping prediction jobs?
- runner: (if it operates on a 'terminated' job) -> updates db that the job is terminated 
- stop API endpoint: (if called) -> stop job -> delete job (changed)

zenovore: no there's only one way to delete the prediction job, from stop API endpoint
the runner will just update the prediction job status according the spark application, so the runner will only update the prediction job status, eg: pending -> running -> completed

api/api/models_api.go
```diff
@@ -99,3 +104,126 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+			allEndpointInModel = append(allEndpointInModel, endpoints...)
+
+		}
+	}
+
+	// DELETING ALL THE RELATED ENTITY
+	for _, version := range versions {
+		if model.Type == "pyfunc_v2" {
+			// DELETE PREDICTION JOBS
+			response := c.VersionsController.deleteInactivePredictionJobs(ctx, allJobInModel, model, version)
```
deadlycoconuts, line 175: This refactoring looks good, though there's a tiny issue with the logic behind the loops. When you check for active prediction jobs/endpoints, you loop through each version of the model, and append all those jobs/endpoints in the list `allJobInModel ` or `allEndpointInModel`. That means that these lists will have a mixture of jobs/endpoints from different versions, e.g. `[v1.job1, v1.job2, v2.job1, v2.job2]`.

However, when you loop through each version again to delete the jobs/endpoints, you're passing this entire list together with with 1 version at a time to the `deleteInactivePredictionJobs` or `deleteInactiveVersionEndpoints` call, i.e. you're calling `deleteInactivePredictionJobs` with version `v1` but these jobs `[v1.job1, v1.job2, v2.job1, v2.job2]`, and then with version `v2` but also with the jobs `[v1.job1, v1.job2, v2.job1, v2.job2]`. 

This might cause an error when the deletion method searches for those inexistent jobs right (that don't correspond to their versions)?

api/api/models_api.go
```diff
@@ -99,3 +104,126 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+
+	var allJobInModel []*models.PredictionJob
+	var allEndpointInModel []*models.VersionEndpoint
+
+	// iterate for every version, check the active prediction job / active endpoint
+	for _, version := range versions {
+		if model.Type == "pyfunc_v2" {
+			// check active prediction job
+			// if there are any active prediction job using the model version, deletion of the model version are prohibited
+			jobs, response := c.VersionsController.checkActivePredictionJobs(ctx, model, version)
```
deadlycoconuts, line 148: I just realised the names `checkActivePredictionJobs` and `checkActiveEndpoints` are misleading because what they actually do two things, 1) retrieve **inactive** prediction jobs/endpoints and 2) return an error if there are active jobs/endpoints. 

Can we rename this to something like, `getInactivePredictionJobsForDeletion`? Same thing for the other method.

api/batch/controller.go
deadlycoconuts: Oops my bad, I was thinking of two different things when writing the first sentence and when listing the bullet points, but yes I had my doubt clarified 😅

deadlycoconuts (approved):
🚀🚀🚀 Thanks a lot for this new big feature! We're hoping it'll be super useful for our users! @ariefrahmansyah @krithika369, I believe most of the comments have been addressed; you may wish to look at the outstanding ones to see if everything's okay :) 

P.S. Please only merge this after you've addressed the comments ahaha 😅

.github/workflows/merlin.yml
```diff
@@ -369,7 +369,7 @@ jobs:
       LOCAL_REGISTRY_PORT: 12345
       LOCAL_REGISTRY: "dev.localhost"
       INGRESS_HOST: "127.0.0.1.nip.io"
-      MERLIN_CHART_VERSION: 0.10.16
+      MERLIN_CHART_VERSION: 0.10.19
```
krithika369, line 372: @deadlycoconuts between this PR and https://github.com/caraml-dev/merlin/pull/397, guess we may need to coordinate a bit.

api/api/models_api.go
krithika369: @ariefrahmansyah, could you clarify this?

api/api/models_api.go
```diff
@@ -99,3 +104,126 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+			allEndpointInModel = append(allEndpointInModel, endpoints)
+
+		}
+	}
+
+	// DELETING ALL THE RELATED ENTITY
+	for index, version := range versions {
+		if model.Type == "pyfunc_v2" {
+			// DELETE PREDICTION JOBS
+			response := c.VersionsController.deleteInactivePredictionJobs(ctx, allJobInModel[index], model, version)
```
krithika369, line 175: CMIIW, but @zenovore, this `[index]` is only correct if each version has exactly 1 job and 1 endpoint. Eg: If there are 0 endpoints for a version or 2 (because the same version can be deployed in multiple environments), this logic fails.

krithika369: Ok scratch that comment. I just realised that the `allJobInModel` and `allEndpointInModel` are array of array. So, this is fine.

.github/workflows/merlin.yml
deadlycoconuts: I'll probably merge after this one 😅 So yep, I'll create the config in my PR. (Thanks for the reminder!)

api/api/models_api.go
```diff
@@ -99,3 +104,126 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+		if model.Type == "pyfunc_v2" {
+			// check active prediction job
+			// if there are any active prediction job using the model version, deletion of the model version are prohibited
+			jobs, response := c.VersionsController.getInactivePredictionJobsForDeletion(ctx, model, version)
+			if response != nil {
+				return response
+			}
+
+			// save all prediction jobs for deletion process later on
+			allJobInModel = append(allJobInModel, jobs)
```
krithika369, line 154: Nit for renaming:
* `allJobInModel` -> `inactiveJobsInModel`
* `allEndpointInModel` -> `inactiveEndpointsInModel`
* `deleteInactivePredictionJobs` -> `deletePredictionJobs` (this function's logic is not checking whether the job is inactive, so it needn't be in the name)
* `deleteInactiveVersionEndpoints` -> `deleteVersionEndpoints` (same reasoning as above)

api/api/versions_api.go
krithika369: Sorry I missed this discussion earlier. Let's keep this implementation for now.

api/api/models_api.go
krithika369: Same comment on the Version API.

api/api/models_api.go
ariefrahmansyah: Sorry I missed this. Technically, it's possible to use `pyfunc_v2` for batch jobs and real-time models. However, in Gojek itself, no real-time models are using `pyfunc_v2`.

deadlycoconuts: > Technically, it's possible to use pyfunc_v2 for batch jobs and real-time models.

TIL 😮

api/api/models_api.go
```diff
@@ -99,3 +104,126 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+	}
+
+	// DELETING ALL THE RELATED ENTITY
+	for index, version := range versions {
+		if model.Type == "pyfunc_v2" {
+			// DELETE PREDICTION JOBS
+			response := c.VersionsController.deleteInactivePredictionJobs(ctx, allJobInModel[index], model, version)
+			if response != nil {
+				log.Errorf("failed to stop prediction job %v", response.data)
+				//return InternalServerError(fmt.Sprintf("Failed stopping prediction job: %s", err))
```
krithika369, line 178: Related to https://github.com/caraml-dev/merlin/pull/389#discussion_r1223651355, you also don't need to log the error from the model/versions controller. It will be done automatically: https://github.com/caraml-dev/merlin/pull/396

Could you remove the explicit logs from the `api/api/models_api.go` and `api/api/versions_api.go`?

api/batch/controller.go
krithika369: @ariefrahmansyah do you see any concerns with this change (`api/batch/controller.go`) - immediately deleting the terminated jobs from DB / cluster?

api/api/models_api.go
krithika369: Got it. This being the case, I think we should look for:
* Prediction jobs if pyfunc_v2
* Endpoints always (including pyfunc_v2)

zenovore: does this mean, pyfunc_v2 model could be deployed?

deadlycoconuts: Yep. So we need to process endpoints for all models, and if it's a `pyfunc_v2` model, we process prediction jobs for it too.

api/batch/controller.go
ariefrahmansyah: I have no concern with immediately deleting the terminated jobs. Technically, it shall be safe with no side effect.

api/api/models_api.go
deadlycoconuts: Resolving this conversation because it has been fixed in the newer commits.

api/api/models_api.go
zenovore: Based on my observation on the UI, the `pyfunc_v2` model is only for prediction jobs only and not having the endpoint function (deploy), please refer [here](https://github.com/caraml-dev/merlin/blob/main/ui/src/version/VersionListTable.js#L491-L536) for the current implementation.

I will add the endpoint check for the API, but since the frontend suggest that we are not letting users deploy `pyfunc_v2` endpoints anyway. we don't need to add the endpoint table on the delete popup confirmation for deleting `pyfunc_v2` model version, wdyt about this?

deadlycoconuts: Thanks for checking that out! Yep, I don't think we'll need to expose the fact that `pyfunc_v2` models can be deployed as an endpoint to our end-users, so we can just make the change in the API server. The same can be said also for the SDK. So I think just having this change on the API would be enough :)

api/api/models_api.go
```diff
@@ -17,6 +17,9 @@ package api
 import (
 	"fmt"
 	"net/http"
+	"strconv"
+
+	"github.com/caraml-dev/merlin/service"
```
krithika369, line 22: Nit. The convention with ordering imports is:
* Standard packages
* Public packages, Like `"github.com/jinzhu/gorm"`
* Local packages. Here, `"github.com/caraml-dev/merlin/service"` is a local package, for this app.

api/api/models_api.go
```diff
@@ -99,3 +103,119 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+		// delete inactive endpoints for all model type
+		response := c.VersionsController.deleteVersionEndpoints(inactiveEndpointsInModel[index], version)
+		if response != nil {
+			return response
+		}
+
+		// DELETE VERSION
+		err = c.VersionsService.Delete(version)
+		if err != nil {
+			return InternalServerError(fmt.Sprintf("Delete Failed: %s", err.Error()))
```
krithika369, line 188: Nit: Delete version %d failed. It's useful to indicate what resource's deletion failed since we are trying to delete multiple versions here.
```golang
return InternalServerError(fmt.Sprintf("Delete version %d failed: %s", version.ID, err.Error()))
```

api/api/models_api.go
```diff
@@ -99,3 +103,119 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+			return InternalServerError(fmt.Sprintf("Unable to delete model endpoint: %s", err.Error()))
+		}
+	}
+
+	// Delete mlflow experiment and artifact if there are model version
+	if len(versions) != 0 {
+		s := strconv.FormatUint(uint64(model.ExperimentID), 10)
+		err = c.MlflowDeleteService.DeleteExperiment(ctx, s, true)
+		if err != nil {
+			return InternalServerError(fmt.Sprintf("Delete Failed: %s", err.Error()))
```
krithika369, line 210: Nit:
```
return InternalServerError(fmt.Sprintf("Delete mlflow experiment failed: %s", err.Error()))
```

api/api/models_api.go
```diff
@@ -99,3 +103,119 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+		}
+	}
+
+	// Delete model data from DB
+	err = c.ModelsService.Delete(model)
+	if err != nil {
+		return InternalServerError(fmt.Sprintf("Delete Failed: %s", err.Error()))
```
krithika369, line 217: Nit:

```
return InternalServerError(fmt.Sprintf("Delete model failed: %s", err.Error()))
```

api/api/versions_api.go
```diff
@@ -141,6 +142,124 @@ func (c *VersionsController) CreateVersion(r *http.Request, vars map[string]stri
...
+	// delete inactive endpoint
+	response = c.deleteVersionEndpoints(inactiveEndpoints, version)
+	if response != nil {
+		return response
+	}
+
+	// delete mlflow run and artifact
+	err = c.MlflowDeleteService.DeleteRun(ctx, version.RunID, version.ArtifactURI, true)
+	if err != nil {
+		return InternalServerError(fmt.Sprintf("Delete Failed: %s", err.Error()))
```
krithika369, line 196: Same comments in this file, as in the models API file, about the error messages.

krithika369 (approved):
Left some small comments. Besides that, LGTM. Thanks for this work and the elaborate tests! 🚀

api/api/models_api.go
zenovore: i was using the goimports package to sort the import, let me change it 👍

api/api/models_api.go
```diff
@@ -99,3 +102,119 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+	versions, _, err := c.VersionsService.ListVersions(ctx, modelID, c.MonitoringConfig, service.VersionQuery{})
+	if err != nil {
+		return InternalServerError(err.Error())
+	}
+
+	var inactiveJobsInModel [][]*models.PredictionJob
+	var inactiveEndpointsInModel [][]*models.VersionEndpoint
+
+	// iterate for every version, check the active prediction job / active endpoint
+	for _, version := range versions {
```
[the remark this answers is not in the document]
ariefrahmansyah: I'm thinking of a case when a model has many versions, let say model M1 has 500 versions. The current implementation will query the table 500 times * 2 (each for prediction job table and version table). We can modify the query so we can just query each table once so the query to table is only 2.

ariefrahmansyah (commented):
I left some nit comments and a discussion on the implementation of iterating over model versions. Overall, the current implementation is looking good for me.

api/api/models_api.go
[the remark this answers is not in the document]
krithika369: Agree with this concern. And it was one of the discussions earlier in our internal channel as well. Perhaps we can get some testing done around the performance and address this in a separate PR.

swagger.yaml
```diff
@@ -342,6 +368,29 @@ paths:
...
+        200:
+          description: "OK"
+          schema:
+            type: "integer"
+        404:
+          description: "Model with given `model_id` not found"
+        400:
+          description: "There are other entity still using this model (model version endpoint / prediction jobs)"
+        500:
+          description: "Deletion failed"
```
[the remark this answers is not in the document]
zenovore: yes, it was already done on the sdk [pr](https://github.com/caraml-dev/merlin/pull/406)

api/api/models_api.go
```diff
@@ -99,3 +102,119 @@ func (c *ModelsController) GetModel(r *http.Request, vars map[string]string, bod
...
+		if model.Type == "pyfunc_v2" {
+			// check active prediction job
+			// if there are any active prediction job using the model version, deletion of the model version are prohibited
+			jobs, response := c.VersionsController.getInactivePredictionJobsForDeletion(ctx, model, version)
```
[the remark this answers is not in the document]
zenovore: yes, it always return error, let me change it to `errResponse`, the reason that i don't use `error`, is because the return type is different, with the function returnting struct `response` immediately, so thats why i don't name it `error`

api/api/models_api.go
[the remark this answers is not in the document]
ariefrahmansyah: Got it. I agree on moving forward with this PR first and try to improve it later.

ariefrahmansyah (approved):
lgtm. thanks for the amazing work, alex!

PR #394, opened 2023-05-12
Update inference-logger kafka message delivery implementation

terryyylim:

Introduction of inference-logger previously - https://github.com/caraml-dev/merlin/pull/373, saw merlin models using kafka logger to be crashing with panics that are related to closed channels.

In the current logic, there is a sleep duration which closes the delivery channel, resulting in future go-routines to panic when working on a closed channel. This PR simplifies the logic for error handling and uses the producer's `Events()` for message delivery. This is also more aligned with Merlin's existing `pkg/kafka` (https://github.com/caraml-dev/merlin/blob/main/api/pkg/kafka/kafka.go#L97) implementation for ease of convergence in the future.

`NONE`

```release-note
NONE
```

**Checklist**

- [x] Tested locally

### DIFF
--- a/api/pkg/inference-logger/logger/handler_test.go
+++ b/api/pkg/inference-logger/logger/handler_test.go
@@ -187,6 +187,7 @@ func Test(t *testing.T) {
 					zapLogger := l.Sugar()
 
 					mockKafkaProducer := &mocks.KafkaProducer{}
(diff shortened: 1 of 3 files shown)


krithika369 (approved):
LGTM. Thanks, @terryyylim !

PR #396, opened 2023-05-18
chore: Refactor error logging from API handlers

krithika369:

This PR refactors the errors logged by the API handlers to do the following:
* `api/utils/stacktrace.go` - Utility function for capturing stacktrace (largely inspired by the examples in the [official docs](https://pkg.go.dev/runtime#Frames))
* `api/api/response.go` - Add `stacktrace string` to the `Response` struct. All utility functions for non-successful responses will capture the stacktrace (which is closer to the origin of the error) in the response.
* `api/api/router.go` - The handler function will log an error to the console will relevant details about the API request / response (such as the route name, error response and stacktrace), when the response indicates failure. The `stacktrace` in the log will be taken from the `Response` data and overrides the default stacktrace printed by the `zap` logger (which would not have been of use in this case, as it points to the logging line in the `HandlerFunc`).

Screenshot 2023-06-06 at 9 44 38 AM

* `api/` - The following changes are made to the error messages and logging.
     - All error messages in API responses now include the original error received from the service layer. On the other hand, some info that can be inferred from the request has been removed from the response (eg: For a non-existent model version 2 under a model with id=1, GET `/models/1/versions/2` will return `Model version not found` as opposed to  `Model version 2 not found`). Besides this, there is a general clean up of error messages to make them more meaningful (eg: previously we were showing errors like `Model version 1 for version 2`; now it is `Model version not found`).
     - Individual log statements for the errors to the console have been removed (as the router is expected to handle this for all failures, via the wrapper `HandlerFunc`). The only exception is `api/api/log_api.go` which is added as a raw handler to the router.

The additional error details are meant to help with debugging API errors.

```release-note
Propagate original error in the response, back to the caller, for non 2XX responses.
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/api/version_endpoints_api.go
```diff
@@ -53,7 +53,7 @@ func (c *EndpointsController) ListEndpoint(r *http.Request, vars map[string]stri
 
 	version, err := c.VersionsService.FindByID(ctx, modelID, versionID, c.MonitoringConfig)
 	if err != nil {
-		return NotFound(fmt.Sprintf("Version with given `version_id: %d` not found", versionID))
+		return NotFound(fmt.Sprintf("Version with given `version_id: %d` not found: %s", versionID, err))
```
pradithya, line 56: Shall we also log the error to the console for any successful return?

krithika369: Yes, sounds good. I'm thinking about logging from the utility methods for `NotFound` and `InternalServerError` [here](https://github.com/caraml-dev/merlin/blob/main/api/api/response.go#L97), and clean up the individual log statements from the API handlers. Wdyt?

pradithya: I think it's better to do it in the [router](https://github.com/caraml-dev/merlin/blob/75853498bb3cd3069553858f2332d4101119e9c8/api/api/router.go#LL121C4-L121C10). Although need to check whether the stack trace is still preserved at that point.

api/api/model_endpoint_alerts_api.go
```diff
@@ -63,9 +61,9 @@ func (c *AlertsController) GetModelEndpointAlert(r *http.Request, vars map[strin
 	modelEndpointAlert, err := c.ModelEndpointAlertService.GetModelEndpointAlert(modelID, modelEndpointID)
 	if err != nil {
 		if gorm.IsRecordNotFoundError(err) {
-			return NotFound(fmt.Sprintf("GetModelEndpointAlert: Alert for model endpoint with id %s not found", modelEndpointID))
+			return NotFound(fmt.Sprintf("Model alert not found: %v", err))
```
deadlycoconuts, line 64: Should this be `fmt.Sprintf("Model endpoint alert not found: %v", err)`, just like in your changes to `UpdateModelEndpointAlert`?

api/api/model_endpoint_alerts_api.go
```diff
@@ -63,9 +61,9 @@ func (c *AlertsController) GetModelEndpointAlert(r *http.Request, vars map[strin
...
 		}
-		return InternalServerError(fmt.Sprintf("GetModelEndpointAlert: Error while getting alert for model endpoint with id %s", modelEndpointID))
+		return InternalServerError(fmt.Sprintf("Error getting alert for model: %v", err))
```
deadlycoconuts, line 66: Should this be `fmt.Sprintf("Error finding alert for model endpoint: %v", err)`, just like in your changes to `UpdateModelEndpointAlert`?

api/api/model_endpoints_api.go
```diff
@@ -166,26 +152,25 @@ func (c *ModelEndpointsController) UpdateModelEndpoint(r *http.Request, vars map
...
 
-			return NotFound(fmt.Sprintf("Environment not found: %s", newEndpoint.EnvironmentName))
 		}
 	}
 	newEndpoint.Environment = env
 
 	currentEndpoint, err := c.ModelEndpointsService.FindByID(ctx, modelEndpointID)
 	if err != nil {
-		return NotFound(fmt.Sprintf("Model Endpoint with given `model_endpoint_id: %d` not found", modelEndpointID))
+		return NotFound(fmt.Sprintf("Environment not found: %v", err))
```
deadlycoconuts, line 173: Should there be a `if gorm.IsRecordNotFoundError(err)` check here, failing which we return `NotFound(fmt.Sprintf("Model endpoint not found: %v", err))`, otherwise we return `InternalServerError(fmt.Sprintf("Error getting model endpoint: %v", err))`? In any case, the error message shouldn't be about an `environment` 😅

api/api/prediction_job_api.go
```diff
@@ -39,10 +38,10 @@ func (c *PredictionJobController) Create(r *http.Request, vars map[string]string
 
 	model, version, err := c.getModelAndVersion(ctx, modelID, versionID)
 	if err != nil {
-		if !gorm.IsRecordNotFoundError(err) {
-			return InternalServerError(err.Error())
+		if gorm.IsRecordNotFoundError(err) {
+			return NotFound(fmt.Sprintf("Model / version not found: %v", err))
```
deadlycoconuts, line 42: Whoa thanks for catching the misplaced `!`! Ditto for the change below (in the same file).

api/api/prediction_job_api.go
```diff
@@ -103,21 +100,20 @@ func (c *PredictionJobController) Get(r *http.Request, vars map[string]string, _
...
 	}
 
 	env, err := c.AppContext.EnvironmentService.GetDefaultPredictionJobEnvironment()
 	if err != nil {
-		return InternalServerError("Unable to find default environment, specify environment target for deployment")
+		return InternalServerError(fmt.Sprintf("Unable to find default environment, specify environment target for deployment: %v", err))
 	}
 
 	job, err := c.PredictionJobService.GetPredictionJob(ctx, env, model, version, id)
 	if err != nil {
```
deadlycoconuts, line 115: It's interesting how we're not performing the `if gorm.IsRecordNotFoundError(err)` check here for `PredictionJobService.GetPredictionJob` and also in other instances below, but I guess it's not super important if we don't do it, just pointing this out 🤔

api/api/version_endpoints_api.go
deadlycoconuts: It seems like previously you were consistently using `%v` to format the error, but it's `%s` in this case. Just wondering if it's intentional 🤔

deadlycoconuts (approved):
A bit late to review this but thanks a lot for this PR! LGTM! 🚀 I have left some comments but they are pretty minor, and overall everything looks good, though I'm a little confused why the `if gorm.IsRecordNotFoundError(err)` check isn't performed consistently everywhere before returning an `InternalServerError` 🤔 (nothing wrong with the refactoring in this PR, but probably with just curious why they weren't performed originally)

.gitignore
```diff
@@ -184,4 +184,6 @@ Temporary Items
 # End of https://www.toptal.com/developers/gitignore/api/linux,macos,intellij,visualstudiocode,jupyternotebooks
 
 # Binary
-bin/
\ No newline at end of file
+bin/
+
+.vscode/
```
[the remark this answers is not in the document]
krithika369: Haha. I use VS Code for development. Without this, the local settings folder will always be shown in the diff. It's common practice to add IDE related artifacts to the `.gitignore` file.

api/api/model_endpoint_alerts_api.go
```diff
@@ -47,9 +45,9 @@ func (c *AlertsController) ListModelEndpointAlerts(r *http.Request, vars map[str
 	modelEndpointAlerts, err := c.ModelEndpointAlertService.ListModelAlerts(modelID)
 	if err != nil {
 		if gorm.IsRecordNotFoundError(err) {
-			return NotFound(fmt.Sprintf("ListModelAlerts: Alerts for Model ID %s not found", modelID))
+			return NotFound(fmt.Sprintf("Model alert not found: %v", err))
```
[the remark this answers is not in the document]
krithika369: The error messages created in the service / api layers will eventually only be logged by the API router. The ID is expected to be covered by the "path" field in the log created by the router (as in the screenshot). But yes, it some scenarios, it could be confusing (eg: when looping through a list of models and one of them has an issue). If such scenarios arise, we could be more explicit about the error messages.

api/api/prediction_job_api.go
krithika369: Good point. Corrected these instances.

api/api/version_endpoints_api.go
krithika369: For default errors, they are equivalent. But I correct this to `%v` to be consistent. Thanks for catching!

krithika369:
Thanks for the detailed review @deadlycoconuts.

> I'm a little confused why the if gorm.IsRecordNotFoundError(err) check isn't performed consistently everywhere before returning an InternalServerError

I checked these instances and corrected the relevant ones now. Thanks for catching!

krithika369:
Thanks for the reviews. Will be merging this shortly.

.gitignore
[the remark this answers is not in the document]
deadlycoconuts: Ahahah got it xD 

> common practice to add IDE related artifacts to the .gitignore file

Was just wondering how common was it if there were 101 common IDEs but I guess that's not a real concern 😅

api/api/model_endpoint_alerts_api.go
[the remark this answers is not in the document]
deadlycoconuts: Okay, got it! Thanks for clarifying! :)

deadlycoconuts:
LGTM! Thanks for the changes and for the clarifications! Feel free to merge whenever you like! :)

PR #397, opened 2023-05-18
Convert API configs from env vars to config file values

deadlycoconuts:
In line with the general direction of moving away from using environment variables to pass configuration to applications, to using configuration files (.yaml), the Merlin API server has been refactored in order to read its configuration values off a specified file instead of environment variables. This change also requires an accompanying change to its Helm chart, as it would no longer be able to utilise the same environment variables to derive its configuration. See https://github.com/caraml-dev/helm-charts/pull/254 for the accompanying PR for this change (that PR has to be merged first before this PR).

Most of this PR involves removing the `envconfig` tags from the various fields in `api/config/config.go`. A couple of additional helper methods (copied over from Turing) are introduced in order to perform the parsing of the configs from a `.yaml` config file.

~~NOTE that the e2e tests have been temporarily been commented out to allow the publishing of a Merlin API server image to allow the changes in the Helm chart repo to pass its CICD tests and to get merged.~~

NONE

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


Dockerfile
```diff
@@ -59,4 +59,4 @@ ENV ZONEINFO=/zoneinfo.zip
 # UI must be built outside docker
 COPY ./ui/build ./ui/build
 
-CMD ["merlin_api"]
+ENTRYPOINT ["sh", "-c", "merlin_api \"$@\"", "--"]
```
deadlycoconuts, line 62: This needs to be changed in order to allow other command line arguments to be passed (like the `-config` flag).

api/cluster/resource/templater_test.go
```diff
@@ -121,8 +121,8 @@ var (
 		},
 		FeastServingKeepAlive: &config.FeastServingKeepAliveConfig{
 			Enabled: true,
-			Time:    time.Duration(30 * time.Second),
-			Timeout: time.Duration(1 * time.Second),
```
deadlycoconuts, line 125: These are redundant and removed.

api/config/config.go
```diff
@@ -396,35 +402,223 @@ func (u *FeastServingURLs) URLs() []string {
...
+// and can be overridden by the values from environment variables. Nested keys in the config
+// can be set from environment variable name separed by "_". For instance the config value for
+// "DbConfig.Port" can be overridden by environment variable name "DBCONFIG_PORT". Note that
+// all environment variable names must be upper case.
+//
+// If no config file is provided, only the default config values and config values from environment
+// varibales will be loaded.
+//
+// Refer to example.yaml for an example of config file.
+func Load(filepaths ...string) (*Config, error) {
```
deadlycoconuts, line 446: This new method is largely copied over from Turing.

api/config/environment.go
```diff
@@ -89,16 +89,20 @@ type ResourceRequestConfig struct {
...
-	err = yaml.Unmarshal(cfgFile, &configs)
-	if err != nil {
+	if err = yaml.Unmarshal(cfgFile, &configs); err != nil {
 		log.Panicf("unable to unmarshall deployment config file:\n %s,\nDue to: %v", cfgFile, err)
 	}
+	for _, env := range configs {
+		if env.K8sConfig == nil {
+			log.Panicf("Error, k8sConfig for %s is nil", env.Name)
+		}
+	}
```
deadlycoconuts, line 105: This new check ensures that all the k8s credentials configs are not nil (a similar check exists in Turing).

api/config/feast_storage_config.go
```diff
@@ -12,35 +11,6 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
-type Duration time.Duration
```
deadlycoconuts, line 15: This type has been removed because it's interfering with how viper is parsing the configs from a `.yaml` file.

scripts/e2e/setup-cluster.sh
```diff
@@ -41,9 +41,9 @@ store_cluster_secret() {
   cat <<EOF |  yq -P - > /tmp/temp_k8sconfig.yaml
 {
     "k8s_config": {
-        "name": $(k3d kubeconfig get "$CLUSTER_NAME" | yq .clusters[0].name -o json -),
+        "name": $(k3d kubeconfig get "$CLUSTER_NAME" | yq '.clusters[0].name' -o json -),
```
deadlycoconuts, line 44: Added these because the expressions (e.g. `.clusters[0].name`) occasionally throw and error if they are not expressed as an explicit string (depends on the version of `yq` used).

api/config/config_test.go
```diff
@@ -54,26 +53,27 @@ func TestFeastServingURLs_URLs(t *testing.T) {
...
 	simulationBigtableURL := "online-bt-serving.dev"
 	testCases := []struct {
-		desc                  string
-		redisConfig           *string
-		bigtableConfig        *string
-		simulationRedisURL    *string
-		simulationBigtableURL *string
-		feastStorageCfg       feast.FeastStorageConfig
+		desc                   string
+		redisConfigFilePath    *string
```
krithika369, line 63: Thanks for this refactoring! A lot more readable. :)

krithika369 (approved):
Thanks for this refactoring, @deadlycoconuts ! I left some small comments. The rest LGTM.

CC: @ariefrahmansyah

api/config/config.go
```diff
@@ -16,36 +16,42 @@ package config
...
-	LoggerDestinationURL  string          `envconfig:"LOGGER_DESTINATION_URL"`
-	Sentry                sentry.Config   `envconfig:"SENTRY" split_words:"false"`
-	NewRelic              newrelic.Config `envconfig:"NEWRELIC" split_words:"false" `
-	EnvironmentConfigPath string          `envconfig:"DEPLOYMENT_CONFIG_PATH" required:"true"`
-	NumOfQueueWorkers     int             `envconfig:"NUM_OF_WORKERS" default:"2"`
-	SwaggerPath           string          `envconfig:"SWAGGER_PATH" default:"./swagger.yaml"`
-
-	DeploymentLabelPrefix string `envconfig:"DEPLOYMENT_LABEL_PREFIX" default:"gojek.com/"`
-	PyfuncGRPCOptions     string `envconfig:"PYFUNC_GRPC_OPTIONS" default:"{}"`
+	Environment           string `default:"dev"`
```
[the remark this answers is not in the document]
deadlycoconuts: Okay I've added that `reflectViperConfig` method though it needed a tiny change to not call `v.SetDefault` under the hood for **_all_** fields.  In particular, the `mlp.K8sConfig` field/struct (of the image builder config) gets initialised with zero-values when `v.SetDefault` gets called on it (it stays as such if it doesn't get populated by non-zero values in the other config files). This causes a problem later when the struct gets converted to an interface, and then marshalled into a yaml byte string, which causes a panic with the following error:
```
panic: reflect.Value.Interface: cannot return value obtained from unexported field or method
```
Previously, when `reflectViperConfig` wasn't used, this wasn't an issue because the `v.SetDefault` wouldn't be creating a zero-valued `mlp.K8sConfig` field that would be parsed by the marshaller. 

A quick fix to this was to create a unique `default:"-"` tag-value pair that simply causes `reflectViperConfig` to ignore that particular field and not call `v.SetDefault` on it. This is applied to the `mlp.K8sConfig` field.

api/config/config.go
```diff
@@ -16,36 +16,42 @@ package config
...
+	Port                  int    `default:"8080"`
+	LoggerDestinationURL  string
+	Sentry                sentry.Config
+	NewRelic              newrelic.Config
+	EnvironmentConfigPath string `validate:"required"`
```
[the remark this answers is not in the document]
deadlycoconuts: Ok, I've just added them wherever relevant (there might be some fields that are probably not 'required' but I guess we can refine these tags over time.

api/config/config.go
```diff
@@ -61,38 +66,38 @@ type Config struct {
...
-	HomePage          string         `envconfig:"REACT_APP_HOMEPAGE" json:"REACT_APP_HOMEPAGE,omitempty"`
-	MaxAllowedReplica int            `envconfig:"REACT_APP_MAX_ALLOWED_REPLICA" default:"20" json:"REACT_APP_MAX_ALLOWED_REPLICA,omitempty"`
-	MerlinURL         string         `envconfig:"REACT_APP_MERLIN_API" json:"REACT_APP_MERLIN_API,omitempty"`
-	MlpURL            string         `envconfig:"REACT_APP_MLP_API" json:"REACT_APP_MLP_API,omitempty"`
-	OauthClientID     string         `envconfig:"REACT_APP_OAUTH_CLIENT_ID" json:"REACT_APP_OAUTH_CLIENT_ID,omitempty"`
-	SentryDSN         string         `envconfig:"REACT_APP_SENTRY_DSN" json:"REACT_APP_SENTRY_DSN,omitempty"`
-	UPIDocumentation  string         `envconfig:"REACT_APP_UPI_DOC_URL" json:"REACT_APP_UPI_DOC_URL,omitempty"`
-	CPUCost           string         `envconfig:"REACT_APP_CPU_COST" json:"REACT_APP_CPU_COST,omitempty"`
-	MemoryCost        string         `envconfig:"REACT_APP_MEMORY_COST" json:"REACT_APP_MEMORY_COST,omitempty"`
+	DocURL            Documentations `json:"REACT_APP_MERLIN_DOCS_URL,omitempty"`
```
krithika369, line 74: Are we not setting the defaults here anymore?

krithika369 (approved):

api/config/config.go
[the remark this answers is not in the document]
krithika369: TIL. I agree that the first method is better.

Btw, whey are we setting `default:"-"` for some other fields too, like `FeastRedisConfig`, `FeastBigtableConfig`, etc. ?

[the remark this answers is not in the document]
deadlycoconuts: Because (I believe) these fields are meant to be optional (as seen from some of the unit tests), and if we don't skip calling `v.SetDefault` on them, those fields will have their nested fields initialised with zero values, like `FeastRedisConfig.ServingURL` (a string), and their own `validate:"required"` tags will invalidate themselves because the tag forces their values to be non-zero.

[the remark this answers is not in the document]
krithika369: Ahh got it. Makes sense. Thanks for the clarification.

api/config/config.go
deadlycoconuts: Hmm, there weren't default values furnished on the app-level previously 🤔 Unless I missed something... Otherwise, I can provide default values from the Helm chart values file. Would that work?

ariefrahmansyah: We are not setting default value of DocURL from the API, but the UI: https://github.com/caraml-dev/merlin/blob/main/ui/src/config.js#L32-L35

So, I think it's okay to keep the status quo. No need to set default value here.

api/config/config.go
```diff
@@ -403,35 +420,209 @@ func (u *FeastServingURLs) URLs() []string {
...
+// These config files will override the default config values (refer to setDefaultValues function)
+// and can be overridden by the values from environment variables. Nested keys in the config
+// can be set from environment variable name separated by "_". For instance the config value for
+// "DbConfig.Port" can be overridden by environment variable name "DBCONFIG_PORT". Note that
+// all environment variable names must be upper case.
+//
+// If no config file is provided, only the default config values and config values from environment
+// variables will be loaded.
+//
+// Refer to example.yaml for an example of config file.
```
ariefrahmansyah, line 463: Thanks for detailed comment!

ariefrahmansyah (approved):
LGTM. Thanks for amazing work, Zi Yi!

api/config/config.go
deadlycoconuts: Ahaha I just copied that entire chunk of code from Turing. In the long run we'll remove this entire function into a separate helper function in the MLP package so that Merlin, Turing and XP can all use it.

deadlycoconuts:
Thanks a lot everyone for the comments! I'll be merging this once the pipeline passes! 🚀🚀

PR #398, opened 2023-05-21
Update mlp client and remove decryption when retrieving secrets

pradithya:
MLP's Secret API removes the application-level encryption when [returning the secret data](https://github.com/caraml-dev/mlp/pull/85). Moving forward, we'll rely on TLS to secure the secrets while in transit.
This PR updates the MLP client used by Merlin and removes the decryption process when retrieving a secret from MLP.

```release-note

```

**Checklist**

- [x] Tested in dev environment with MLP version https://github.com/caraml-dev/mlp/pull/85
- [N.A.] Added unit test, integration, and/or e2e tests


api/client/api_alert.go
```diff
@@ -126,8 +126,8 @@ func (a *AlertApiService) AlertsTeamsGet(ctx context.Context) ([]string, *http.R
 
 /*
 AlertApiService Lists alerts for given model.
- * @param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
- * @param modelId
+  - @param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
```
krithika369, line 129: Why is this `*` removed from the auto-generated code now? Is it simply because of a newer version of `swagger-codegen` ?

scripts/quick_install.sh
```diff
@@ -170,7 +170,7 @@ helm upgrade --install merlin caraml/merlin --namespace=mlp --values=./values-e2
   --version ${MERLIN_CHART_VERSION} \
   --set deployment.image.tag=${VERSION} \
   --set ui.oauthClientID=${OAUTH_CLIENT_ID} \
-  --set mlpApi.apiHost=http://mlp.mlp.${INGRESS_HOST}.nip.io/v1 \
+  --set mlpApi.apiHost=http://mlp.mlp.${INGRESS_HOST}.nip.io \
```
krithika369, line 173: Thanks for updating the host to drop the `/v1`.

(Guess we can update the global / default values for the charts, once this change is made to turing / xp as well. Will make a note of this.)

krithika369 (approved):
LGTM. Thanks, @pradithya !

api/client/api_alert.go
pradithya: `gofmt` did that and not swagger

scripts/quick_install.sh
pradithya: I am updating chart for both merlin and turing as well.

PR #401, opened 2023-05-23
Add column expression documentation

tiopramayudi:

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/docs/user-guide/transformer_expressions.md
+++ b/docs/user-guide/transformer_expressions.md
@@ -20,6 +20,20 @@ Standard Transformer provides several built-in functions that are useful for com
 | Time       | [FormatTimestamp](#formattimestamp)                          |
 | Time       | [ParseTimestamp](#parsetimestamp)                            |
 | Time       | [ParseDateTime](#parsedatetime)                              |
+| Column     | [Get](#get)                                                  |
+| Column     | [IsIn](#isin)                                                |
+| Column     | [StdDev](#stddev)                                            |
+| Column     | [Mean](#mean)                                                |
+| Column     | [Median](#median)                                            |
+| Column     | [Max](#max)                                                  |
+| Column     | [MaxStr](#maxstr)                                            |
+| Column     | [Min](#min)                                                  |
+| Column     | [MinStr](#minstr)                                            |
+| Column     | [Quantile](#quantile)                                        |
+| Column     | [Sum](#sum)                                                  |
+| Column     | [Flatten](#flatten)                                          |
+| Column     | [Unique](#unique)                                            |
+
 
 ## Geospatial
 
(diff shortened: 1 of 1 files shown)


docs/user-guide/transformer_expressions.md
```diff
@@ -535,3 +549,397 @@ variables:
...
+
+Standard Transformer Config:
+```
+variables:
+- name: max
+  expression: yourTableName.Col("avg_order_1_day").Max()
+```
+Output: 4000
+
+### MaxStr
```
krithika369, line 732: Curious - what is the motivation for having this function? Is it not possible for users to typecast the value of `max` to string using other functions? (From the doc, it seems we don't have type conversion functions, but confirming...)

docs/user-guide/transformer_expressions.md
```diff
@@ -535,3 +549,397 @@ variables:
...
+Standard Transformer Config:
+```
+variables:
+- name: sum
+  expression: yourTableName.Col("avg_order_1_day").Sum()
+```
+Output: 9000
+
+### Flatten
+`Flatten` is a function to flatten all values in a column, this is suitable for column that has list type, for non list the result will be the same with the original columns
```
krithika369, line 880: In a table, what is the result of flatten on the other columns? For example, after flatten on `nearby_restaurant_ids` column below, can the user select `restaurant_id` too? Will its value be preserved? Like:
```

| restaurant_id | restaurant_ids |
|---------------|-----------------|
| 1 | 2 |
| 1 | 3 |
| 1 | 4 |
| 2 | 4 |
...
```

krithika369 (approved):
Left some questions. LGTM. Thanks, @tiopramayudi

docs/user-guide/transformer_expressions.md
```diff
@@ -535,3 +549,397 @@ variables:
...
+| 2 | 3000 | 0.005 |
+| 3 | 4000 | 0.006 |
+
+Users try to retrieve index 2 for column `restaurant_id`
+
+Standard Transformer Config:
+```
+variables:
+- name: restaurant_2
+  expression: yourTableName.Col("restaurant_id").Get(2)
```
pradithya, line 583: Can we use `avg_order_1_day` as example? And also, shouldn't it return `3` in this case since it's 0-index

docs/user-guide/transformer_expressions.md
```diff
@@ -20,6 +20,20 @@ Standard Transformer provides several built-in functions that are useful for com
 | Time       | [FormatTimestamp](#formattimestamp)                          |
 | Time       | [ParseTimestamp](#parsetimestamp)                            |
 | Time       | [ParseDateTime](#parsedatetime)                              |
+| Column     | [Get](#get)                                                  |
```
pradithya, line 23: Strictly speaking, these are methods supported by `Series` (type returned by `table.Col` ).

docs/user-guide/transformer_expressions.md
```diff
@@ -535,3 +549,397 @@ variables:
...
+#### Output
+New Column that has unique value for each row
+
+#### Examples
+
+Suppose users have table `yourTableName`
+
+| restaurant_id | rating |
+|---------------|-----------------|
+| 1 | [2, 3, 4] |
```
pradithya, line 931: What happen to the second sub-column of `restaurant_id` ?

docs/user-guide/transformer_expressions.md
tiopramayudi: I think this is to expose all the method that already provided by `gota` for series that maybe useful for user, cmiiw @ariefrahmansyah. Yes unfortunately we don't have type conversion functions

docs/user-guide/transformer_expressions.md
tiopramayudi: > In a table, what is the result of flatten on the other columns? For example, after flatten on `nearby_restaurant_ids` column below, can the user select `restaurant_id` too? Will its value be preserved? Like:
> 
> ```
> 
> | restaurant_id | restaurant_ids |
> |---------------|-----------------|
> | 1 | 2 |
> | 1 | 3 |
> | 1 | 4 |
> | 2 | 4 |
> ...
> ```

the `flatten` function will return new series (column), if you use this expression in `variables` operation, the output will be variable that type is series(column). CMIIW, your question is if user use the expression in the `updateColumns` operation like below
```
 inputTable: yourTableName
 outputTable: yourOutputTableName
  steps:
      - updateColumns:
         - column: flatten_restaurant_id
            expression: yourTableName.Col("nearby_restaurant_ids").Flatten()
```
? If that the case, the `yourOutputTableName` has the same value with the `yourTableName` and plus one column `flatten_restaurant_id` (if the dimension of new column is the same with existing table otherwise the operation will be fail)

docs/user-guide/transformer_expressions.md
tiopramayudi: oh yeah 🤦 , I start count from 1

docs/user-guide/transformer_expressions.md
tiopramayudi: what do you mean by sub-column of `restaurant_id` ? do you mean what happen if `Unique` is run for `rating` column?

pradithya: nvm, I thought `restaurant_id` has a nested column since I am only reading the raw markdown. 
Follow up question, what is the output of `yourTableName.Col("rating").Unique()` expression?

docs/user-guide/transformer_expressions.md
pradithya: We have implicit type conversion in some parts of the standard transformer. But I think it'll be useful to have an explicit expression as well.

docs/user-guide/transformer_expressions.md
tiopramayudi: I already updated in the docs

PR #405, opened 2023-06-06
feat: Add caching for MLP projects

krithika369:

This PR introduces a read-aside in-memory cache for the MLP projects API, specifically, for getting a project by ID. The implementation is largely the same as in [Turing](https://github.com/caraml-dev/turing/blob/main/api/turing/service/mlp_service.go), with some existing minor differences:
* The Merlin codebase relies on an [internal MLP client](https://github.com/caraml-dev/merlin/blob/main/api/mlp/client.go) instead of directly using the one exposed by MLP. This is retained.
* Turing uses `MLPService` to cover both projects and secrets MLP APIs (in addition to Merlin's environments API). In Merlin's codebase this is split into the `ProjectsService` and `SecretService`.

With this change, in most cases, look up of individual projects can be done in-memory, bypassing the MLP API server. The MLP List projects API will only be called under the following scenarios:
* When Merlin's List projects API is called - this will get the full list of projects from the MLP API and (re-)cache them in-memory
* When a specific project look up in the cache fails, the full list of projects will be retrieved from MLP again and re-cached

```release-note
Cache MLP projects for faster responses.
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


Makefile
```diff
@@ -42,7 +42,6 @@ init-dep-ui:
 .PHONY: init-dep-api
 init-dep-api:
 	@echo "> Initializing API dependencies ..."
-	@cd ${API_PATH} && go mod tidy -v
```
krithika369, line 45: `tidy` is no longer available in Go 1.20:

> go.mod file indicates go 1.20, but maximum version supported by tidy is 1.18

api/service/project_service.go
```diff
@@ -1,49 +1,132 @@
...
+)
+
+// ProjectsService provides a set of methods to interact with the MLP / Merlin APIs
 type ProjectsService interface {
-	List(ctx context.Context, projectName string) (mlp.Projects, error)
+	// List lists available projects, optionally filtered by given project `name`
+	List(ctx context.Context, name string) (mlp.Projects, error)
+	// GetByID gets the project matching the provided id.
 	GetByID(ctx context.Context, projectID int32) (mlp.Project, error)
-	GetByName(ctx context.Context, projectName string) (mlp.Project, error)
```
krithika369, line 26: This method is unused.

api/service/project_service.go
```diff
@@ -1,49 +1,132 @@
...
+type projectsService struct {
+	mlpClient mlp.APIClient
+	cache     *cache.Cache
+}
+
+// NewProjectsService returns a service that retrieves information that is shared across MLP projects.
+func NewProjectsService(mlpAPIClient mlp.APIClient) (ProjectsService, error) {
+	svc := &projectsService{
+		mlpClient: mlpAPIClient,
+		cache:     cache.New(projectCacheExpirySeconds*time.Second, projectCacheCleanUpSeconds*time.Second),
```
ariefrahmansyah, line 41: The implementation is lgtm.

However, can we re-use `pkg/transformer/cache` that uses `github.com/coocood/freecache` instead of introducing new dependency (`github.com/patrickmn/go-cache`)? If needed, we can refactor it to `pkg/cache`.

krithika369:
@ariefrahmansyah could you have another look? Thanks.

api/service/project_service.go
```diff
@@ -1,49 +1,143 @@
...
+		if strings.HasPrefix(key, projectCacheKeyPrefix) {
+			var project mlpClient.Project
+			if err := json.Unmarshal(entry.Value, &project); err != nil {
+				return mlp.Projects{}, fmt.Errorf("Malformed project info found in the cache for key %s: %w", key, err)
+			} else {
+				// If either the name filter is not set or the project's name matches
+				if name == "" || project.Name == name {
+					projects = append(projects, project)
+				}
+			}
```
ariefrahmansyah, line 95: nit, I prefer fewer branching here and faster return when filtered project found:

```
if err := json.Unmarshal(entry.Value, &project); err != nil {
        return mlp.Projects{}, fmt.Errorf("Malformed project info found in the cache for key %s: %w", key, err)
}

if project.Name == name {
	return mlp.Projects{project}
}

projects = append(projects, project)
```

ariefrahmansyah (approved):
LGTM with nit comments. Thanks, Krithika!

PR #408, opened 2023-06-13
Fail fast if cannot initialize logSink

ariefrahmansyah:

When trying to use Kafka LogSink but failed to initialize NewKafkaSink, the getLogSink function returns nil and the program continue running. This can lead to panic because the program trying to use nil value when trying to Send the payload log:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xf2359d]

goroutine 31 [running]:
github.com/caraml-dev/merlin/pkg/inference-logger/logger.(*Worker).Send(...)
	/src/log-collector/pkg/inference-logger/logger/worker.go:106
github.com/caraml-dev/merlin/pkg/inference-logger/logger.(*Worker).Start.func1()
	/src/log-collector/pkg/inference-logger/logger/worker.go:79 +0x5dd
created by github.com/caraml-dev/merlin/pkg/inference-logger/logger.(*Worker).Start
	/src/log-collector/pkg/inference-logger/logger/worker.go:40 +0x56
```

This PR modifies getLogSink to also return error and fail faster if it failed to initialize KafkaSink.

Fixes inference-logger raise panic.

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cmd/inference-logger/Makefile
+++ b/api/cmd/inference-logger/Makefile
@@ -1,0 +1,15 @@
+APP_NAME=inference-logger
+LOG_URL?=kafka:localhost:9092
+
+.PHONY: build
+build:
+	@go build -o bin/${APP_NAME} ./main.go
+
+.PHONY: run
+run:
+	@rm /tmp/agent.sock || true
+	@SERVING_READINESS_PROBE='{"tcpSocket":{"port":8080,"host":"127.0.0.1"},"successThreshold":1}' UNIX_SOCKET_PATH="/tmp/agent.sock" ./bin/${APP_NAME} -log-url="${LOG_URL}"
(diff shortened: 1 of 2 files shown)


krithika369 (approved):

krithika369:
Btw, how can we fail fast for NewRelic logger?

ariefrahmansyah:
> Btw, how can we fail fast for NewRelic logger?

NewRelic library does not have a built-in Ping API like Kafka, so instead, I think we can build and send our own "ping" log to NewRelic after NewRelicSink initialization.

PR #411, opened 2023-06-15
Bugfix: Add image building job deletion delay

deadlycoconuts:
As of present, when the Merlin API server triggers an image building job as part of launching a batch prediction job/service, it checks for the presence of an existing image building job in the cluster with the same name. The status of the image building job found (if it is found), corresponds to a variety of situations:
- Found; running -> an existing image building job is already taking place and still running -> wait for the job to complete
- Found; failed -> an existing image building job has failed -> delete the old job -> begin a new one
- Not found -> begin a new one

In particular, we're interested in the **_second_** scenario, where there a new job is triggered right after the deletion of the old (failed) run (see [this](https://github.com/caraml-dev/merlin/blob/9d2689df94ba0f622ce01bdb60ed27a9af37a603/api/pkg/imagebuilder/imagebuilder.go#L142) for more details): 

```go
err = jobClient.Delete(ctx, job.Name, metav1.DeleteOptions{})
if err != nil {
	log.Errorf("error deleting job: %v", err)
return "", ErrDeleteFailedJob
}

jobSpec, err := c.createKanikoJobSpec(project, model, version)
if err != nil {
	log.Errorf("unable to create job spec %s, error: %v", imageRef, err)
	return "", ErrUnableToCreateJobSpec{
		Message: err.Error(),
	}
}

job, err = jobClient.Create(ctx, jobSpec, metav1.CreateOptions{})
if err != nil {
	log.Errorf("unable to build image %s, error: %v", imageRef, err)
	return "", ErrUnableToBuildImage{
		Message: err.Error(),
	}
}
```

A potential problem can occur whereby the creation of the new job happens _too quickly_ after the deletion of the job, due to the k8s resources having insufficient time to be removed from the cluster (the kube API server would return an error saying that the creation of the new job has failed due to the existence of another job with the very same name). When this happens, the (new) image building job fails immediately.

This PR thus introduces a new method to wait for the deletion to be complete before proceeding with the creation of another image builder job.

#### Main Modifications
- `api/pkg/imagebuilder/imagebuilder.go` - Addition of an additional helper method to wait for image building jobs to be completely removed from the cluster

Fixes #

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/imagebuilder/imagebuilder.go
+++ b/api/pkg/imagebuilder/imagebuilder.go
@@ -71,7 +71,7 @@ const (
 	// jobCompletionTickDurationSecond is the interval at which the API server checks if a job has completed
 	jobCompletionTickDurationSecond = 5
 	// jobDeletionTimeoutSeconds is the maximum time to wait for a job to be deleted from a cluster
-	jobDeletionTimeoutSeconds = 10
+	jobDeletionTimeoutSeconds = 30
 	// jobDeletionTickDurationMilliseconds is the interval at which the API server checks if a job has been deleted
 	jobDeletionTickDurationMilliseconds = 100
 
--- a/api/pkg/imagebuilder/imagebuilder.go
+++ b/api/pkg/imagebuilder/imagebuilder.go
@@ -350,7 +350,7 @@ func (c *imageBuilder) waitJobDeleted(ctx context.Context, job *batchv1.Job) err
 		case <-ticker.C:
 			_, err := jobClient.Get(ctx, job.Name, metav1.GetOptions{})
 			if err != nil {
-				if !kerrors.IsNotFound(err) {
+				if kerrors.IsNotFound(err) {
 					return nil
 				}
 				log.Errorf("unable to get job status for job %s: %v", job.Name, err)
--- a/api/pkg/imagebuilder/imagebuilder.go
+++ b/api/pkg/imagebuilder/imagebuilder.go
@@ -68,8 +68,8 @@ type imageBuilder struct {
 const (
 	containerName    = "pyfunc-image-builder"
 	kanikoSecretName = "kaniko-secret"
-	// jobDeletionTickDurationMilliseconds is the interval at which the API server checks if a job has completed
-	jobCompletedTickDurationSecond = 5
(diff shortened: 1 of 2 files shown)


api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -66,9 +66,14 @@ type imageBuilder struct {
 }
 
 const (
-	containerName      = "pyfunc-image-builder"
-	kanikoSecretName   = "kaniko-secret"
-	tickDurationSecond = 5
+	containerName    = "pyfunc-image-builder"
+	kanikoSecretName = "kaniko-secret"
+	// jobDeletionTickDurationMilliseconds is the interval at which the API server checks if a job has completed
+	jobCompletedTickDurationSecond = 5
```
krithika369, line 72: Nit: `jobCompletedTickDurationSecond` -> `jobCompletionTickDurationSecond` for consistency?

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -66,9 +66,14 @@ type imageBuilder struct {
...
+	// jobDeletionTimeoutSeconds is the maximum time to wait for a job to be deleted from a cluster
+	jobDeletionTimeoutSeconds = 10
```
krithika369, line 74: Is this sufficient? Or should we make it 30 seconds, to give ample time, just in case?

krithika369 (approved):
LGTM. Thanks for this fix, @deadlycoconuts ! Left some small comments.

api/pkg/imagebuilder/imagebuilder.go
ariefrahmansyah: +1 on making it 30 seconds

ariefrahmansyah (approved):

api/pkg/imagebuilder/imagebuilder.go
deadlycoconuts: Oops nice catch!

deadlycoconuts:
Thanks a lot everyone for the quick review! I'll be merging this once the pipeline passes!

PR #415, opened 2023-06-20
fix: Check cached projects first when listProjects by a specific name

krithika369:

Follow up from: https://github.com/caraml-dev/merlin/pull/405

With the project caching changes made earlier, the `ListProjects` API will end up refreshing all the projects in the cache every single time. This worsens the performance of `merlin.set_project(name)` SDK method, that relies on this function under the hood (especially when running multiple e2e test cases in parallel and there are hundreds on MLP projects). 

To overcome this performance degradation, a slight modification is made to the implementation. We first search the cache for matching projects and if > 0 projects are found, they are returned, before attempting to refresh cache and try again.

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


deadlycoconuts (approved):

PR #416, opened 2023-06-26
Bugfix: Remove DictEnv custom type

deadlycoconuts:
PR #397 refactored multiple config fields used in the Merlin API server though some of the fields refactored were not cleanly removed. The `NodeSelectors` field was using a custom `DictEnv` type that was created to help unmarshall a map object stored in a json string (that was passed to the API server via an environment variable); however, with the refactoring done in the said PR, there is now no need for that custom type to exist, since Viper (the package used to unmarshall yaml configs) is able to directly parse a yaml map.

This PR removes that custom type, as well as two other unused fields, `BuildContextURI` and `PredictionJobBuildContextURI` from the `ImageBuilderConfig` struct.

NONE

Fixes #

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/config/config.go
+++ b/api/config/config.go
@@ -216,7 +216,7 @@ type ImageBuilderConfig struct {
 	// How long to keep the image building job resource in the Kubernetes cluster. Default: 2 days (48 hours).
 	Retention     time.Duration `validate:"required" default:"48h"`
 	Tolerations   Tolerations
-	NodeSelectors DictEnv
+	NodeSelectors map[string]string
 	MaximumRetry  int32                 `validate:"required" default:"3"`
 	K8sConfig     *mlpcluster.K8sConfig `validate:"required" default:"-"`
 	SafeToEvict   bool                  `default:"false"`
@@ -231,18 +231,6 @@ func (spec *Tolerations) Decode(value string) error {
 		return err
 	}
 	*spec = tolerations
-	return nil
-}
-
-type DictEnv map[string]string
-
-func (d *DictEnv) Decode(value string) error {
-	var dict DictEnv
-
-	if err := json.Unmarshal([]byte(value), &dict); err != nil {
-		return err
-	}
-	*d = dict
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


api/config/config.go
```diff
@@ -197,26 +197,25 @@ type ClusterConfig struct {
 }
 
 type ImageBuilderConfig struct {
-	ClusterName                  string `validate:"required"`
-	GcpProject                   string
-	BuildContextURI              string `validate:"required"`
-	ContextSubPath               string
-	DockerfilePath               string           `validate:"required" default:"./Dockerfile"`
-	BaseImages                   BaseImageConfigs `validate:"required"`
-	PredictionJobBuildContextURI string           `validate:"required"`
```
deadlycoconuts, line 206: Removed the `PredictionJobBuildContextURI` since it's not used anywhere.

api/config/config.go
```diff
@@ -197,26 +197,24 @@ type ClusterConfig struct {
 }
 
 type ImageBuilderConfig struct {
-	ClusterName                  string `validate:"required"`
-	GcpProject                   string
-	BuildContextURI              string `validate:"required"`
```
deadlycoconuts, line 202: Removed the `BuildContextURI` field since it's not used anywhere.

api/config/config.go
```diff
@@ -315,15 +301,15 @@ type ModelClientKeepAliveConfig struct {
...
 	ImageName             string `validate:"required"`
 	FeastServingURLs      FeastServingURLs
-	FeastCoreURL          string                       `validate:"required"`
-	FeastCoreAuthAudience string                       `validate:"required"`
-	EnableAuth            bool                         `default:"false"`
-	FeastRedisConfig      *FeastRedisConfig            `default:"-"`
-	FeastBigtableConfig   *FeastBigtableConfig         `default:"-"`
-	FeastGPRCConnCount    int                          `validate:"required" default:"10"`
-	FeastServingKeepAlive *FeastServingKeepAliveConfig `default:"-"`
-	ModelClientKeepAlive  *ModelClientKeepAliveConfig  `default:"-"`
```
deadlycoconuts, line 325: @krithika369 I've removed the tags from these two configs is never expected to be `nil`. This also allows their nested `default` tags to work and set the correct default values on them.

krithika369 (approved):
LGTM. Thanks for the fix, @deadlycoconuts !

PR #418, opened 2023-07-01
chore: Update cache library

krithika369:

This PR replaces the use of `"github.com/coocood/freecache"` in the codebase with `"github.com/patrickmn/go-cache"` for the following reasons:
* The `freecache` library needs us to declare the size of the cache ahead of time. `go-cache`, on the other hand, is map-based and simpler.
* The `freecache` library requires each entry to be less than 1/1024 of cache size. This would require us to calculate the cache size ahead of time and does not allow for the occasional large-sized entries. `go-cache` has no such restriction.

With this change, the transformer config `CACHE_SIZE_IN_MB` is no longer required and has been removed.

**Other changes**:

`.env.sample` is replaced by `env-config.yaml`, based on the changes introduced by https://github.com/caraml-dev/merlin/pull/397

```release-note
Replace Project and Transformer caching library with one that can grow flexibly. Support for the transformer env var config CACHE_SIZE_IN_MB is now removed and if set, will be unused.
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/service/project_service.go
```diff
@@ -58,7 +57,7 @@ func (ps projectsService) List(ctx context.Context, name string) (mlp.Projects,
 		}
 	}
 	// Refresh cache and try filtering
-	err := ps.refreshProjects(ctx)
+	err := ps.refreshProjects(ctx, name)
```
krithika369, line 60: When retrieving the projects list, use the name filter if supplied. Without the name filter, all the projects will be retrieved which may be wasteful if we are only interested in one.

Makefile
```diff
@@ -1,4 +1,3 @@
-include .env.sample
 export
```
ariefrahmansyah, line 1: I think we can remove this export as well.

api/cmd/transformer/Makefile
```diff
@@ -1,4 +1,3 @@
-include .env.sample
 export
```
ariefrahmansyah, line 1: Same as above. Should be safe to remove this line.

env-config.yaml
```diff
@@ -0,0 +1,75 @@
+ClusterConfig:
```
ariefrahmansyah, line 1: nit: to not be confused with environment.yaml, i think it's better to rename this file to `config.yaml`

ariefrahmansyah (approved):
LGTM with nit comments.

I found go-cache library is simpler because (1) it uses interface instead of []byte. Allowing us to just cast the cache value instead of encoding and decoding json. And (2) We don't need to specify cache size.

tiopramayudi:
what happens if the cache keeps increasing? Can it cause OOM?

krithika369:
> what happens if the cache keeps increasing? Can it cause OOM?

@tiopramayudi, that's a good question. Most users set the TTL (using `FEAST_CACHE_TTL`) to 10 minutes or less. As long as the keys are set to expire frequently, they will be cleaned up routinely and there shouldn't be an explosion of memory.

That said, in theory, if there is lots of data added to the cache with a long expiration period, there can be OOM. So we can add a warning to the users when the feature is released, that they should allocate sufficient memory to the transformer containers if they query a lot of data in their transformation steps.

PR #421, opened 2023-07-05
Fix getting model and model endpoint id for rendering model alert page

ariefrahmansyah:

There's a bug on Model Endpoint Alert page where users 1) Cannot open alert page if they write the endpoint on the browser address bar and 2) Cannot create alert.

The first issue is due to the page cannot get the model data from location state, hence I updated it to fetch the model data from the API.

And the second issue is due to endpointId is undefined because endpointId is not passed correctly as well. I updated it to get the value via `useParams` hook instead of via component props.

Fixes Model Alert UI

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/ui/src/model/ModelDetails.js
+++ b/ui/src/model/ModelDetails.js
@@ -44,7 +44,7 @@ export const ModelDetails = () => {
     },
     {
       text: model.name,
-      href: `/merlin/projects/${projectId}/models/${model.id}`,
+      href: `/merlin/projects/${projectId}/models/${modelId}`,
     },
   ];
 
--- a/ui/src/model/ModelDetails.js
+++ b/ui/src/model/ModelDetails.js
@@ -14,17 +14,13 @@
  * limitations under the License.
  */
 
-import React, { useEffect, useState } from "react";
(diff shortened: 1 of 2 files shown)


ui/src/model/ModelDetails.js
```diff
@@ -34,34 +30,23 @@ const LoadingContent = () => (
...
-      ]);
-  }, [projectId, model]);
+  const breadcrumbs = [
+    {
+      text: "Models",
+      href: `/merlin/projects/${projectId}/models`,
+    },
+    {
+      text: model.name,
+      href: `/merlin/projects/${projectId}/models/${model.id}`,
```
krithika369, line 47: Though the component will be reloaded when the `model` value changes, shall we simply use the constant `modelId` here instead of `model.id`?

krithika369 (approved):
Left a small comment. LGTM. Thanks for the fix, @ariefrahmansyah !

ui/src/model/ModelDetails.js
ariefrahmansyah: You are right. Thanks for the suggestion. Updated.

PR #424, opened 2023-07-07
Add viperKeyDelimiter; Add unit test for imagebuilder nodeselectors

ariefrahmansyah:

Previous implementation of Viper cannot load the YAML key that contains ".". For example, give this config file:

```
NodeSelectors:
    cloud.google.com/gke-nodepool: image-build-job-node-pool
```

config.Load will raise error:

```
Failed initializing config: failed to unmarshal config values: 1 error(s) decoding:

* 'ImageBuilderConfig.NodeSelectors[cloud]' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[google:map[com/gke-nodepool:image-build-job-node-pool]]'
```

This PR fixes this issue by setting viper key delimiter to `::`

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/config/config.go
+++ b/api/config/config.go
@@ -37,6 +37,8 @@ import (
 )
 
 const (
+	viperKeyDelimiter = "::"
+
 	MaxDeployedVersion = 2
 )
 
@@ -451,7 +453,7 @@ func (cfg *Config) Validate() error {
 //
 // Refer to example.yaml for an example of config file.
 func Load(spec interface{}, filepaths ...string) (*Config, error) {
-	v := viper.New()
+	v := viper.NewWithOptions(viper.KeyDelimiter(viperKeyDelimiter))
 
 	err := reflectViperConfig("", spec, v)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


api/config/config.go
```diff
@@ -37,6 +37,8 @@ import (
 )
 
 const (
+	viperKeyDelimiter = "::"
```
deadlycoconuts, line 40: Nice, thanks for making this a constant!

api/config/testdata/valid-imagebuilder-nodeselectors.yaml
```diff
@@ -0,0 +1,3 @@
+ImageBuilderConfig:
+  NodeSelectors:
+    "cloud.google.com/gke-nodepool": image-build-job-node-pool
```
deadlycoconuts, line 3: Not really a comment on the PR but with the changes I don't think you need the quotes for `cloud.google.com/gke-nodepool` anymore.

deadlycoconuts (approved):
LGTM! 🚀 Thanks for discovering this and fixing this! :DD

PR #425, opened 2023-07-10
Fix PDB implementation

ariefrahmansyah:

1. Fix PDB apply by supplying `FieldManager: "application/apply-patch"` https://github.com/kubernetes/client-go/issues/1036#issuecomment-989485582
2. Fix PDB config to always have % as suffix

Fixes PDB implementation

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cluster/pdb.go
+++ b/api/cluster/pdb.go
@@ -20,7 +20,6 @@ type PodDisruptionBudget struct {
 	Labels                   map[string]string
 	MaxUnavailablePercentage *int
 	MinAvailablePercentage   *int
-	Selector                 *metav1.LabelSelector
 }
 
 func NewPodDisruptionBudget(modelService *models.Service, componentType string, pdbConfig config.PodDisruptionBudgetConfig) *PodDisruptionBudget {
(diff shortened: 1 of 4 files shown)


api/cluster/pdb.go
```diff
@@ -90,7 +89,7 @@ func (c *controller) deployPodDisruptionBudget(ctx context.Context, pdb *PodDisr
 	pdbCfg.WithLabels(pdb.Labels)
 	pdbCfg.WithSpec(pdbSpec)
 
-	_, err = c.policyClient.PodDisruptionBudgets(pdb.Namespace).Apply(ctx, pdbCfg, metav1.ApplyOptions{})
+	_, err = c.policyClient.PodDisruptionBudgets(pdb.Namespace).Apply(ctx, pdbCfg, metav1.ApplyOptions{FieldManager: "application/apply-patch"})
```
krithika369, line 92: Missed this in the previous PR. But what's the reason for using `Apply` instead of `Create`? https://github.com/kubernetes/client-go/blob/v0.27.3/kubernetes/typed/policy/v1/poddisruptionbudget.go#L118

ariefrahmansyah: So we can use Apply for both Undeploy and Redeploy. Otherwise, we need to implement a function that call Update.

krithika369: > So we can use Apply for both Undeploy and Redeploy

Do you mean for both Deploy and Redeploy?

krithika369 (approved):
Thanks for the fix, @ariefrahmansyah ! I imagine we will be adding it to https://github.com/caraml-dev/turing/pull/346 too.

LGTM.

api/cluster/pdb.go
ariefrahmansyah: Oops, yes, you're right. It's for Deploy and Redeploy.

PR #427, opened 2023-07-11
Fix panic due to use incorrect value for error message

tiopramayudi:

We faced panic issue when update column in ST that involve conditional , this is happened due to accessing nil pointer when construct error message. 

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/transformer/pipeline/table_transform_op.go
+++ b/api/pkg/transformer/pipeline/table_transform_op.go
@@ -229,7 +229,7 @@ func updateColumns(env *Environment, specs []*spec.UpdateColumn, resultTable *ta
 
 			columnValue, err := subsetSeriesFromExpression(env, condition.Expression, rowIndex)
 			if err != nil {
-				return nil, fmt.Errorf("error evaluation column rule value for column %s: %v. Err: Ts", columnName, condition.Default.Expression)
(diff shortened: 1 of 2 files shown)


api/pkg/transformer/pipeline/table_transform_op.go
```diff
@@ -229,7 +229,7 @@ func updateColumns(env *Environment, specs []*spec.UpdateColumn, resultTable *ta
 
 			columnValue, err := subsetSeriesFromExpression(env, condition.Expression, rowIndex)
 			if err != nil {
-				return nil, fmt.Errorf("error evaluation column rule value for column %s: %v. Err: Ts", columnName, condition.Default.Expression)
+				return nil, fmt.Errorf("error evaluation column rule value for column %s: %v. Err: %v", columnName, condition.Expression, err)
```
ariefrahmansyah, line 232: nit: please use %w for err.

ariefrahmansyah (approved):
lgtm with nit comments

api/pkg/transformer/pipeline/table_transform_op.go
tiopramayudi: updated

PR #404, opened 2023-06-04
merlin-sdk docker image

ariefrahmansyah:
**What this PR does / why we need it**:

This PR introduces new Docker image, namely `merlin-sdk` that contains the SDK source code and its test suite. This docker enables faster development iteration.

This PR also updates mock_oauth to cater for running e2e-test on machine that use Google Workload Identity.

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Checklist**

- [ ] Added unit test, integration, and/or e2e tests
- [x] Tested locally
- [ ] Updated documentation
- [ ] Update Swagger spec if the PR introduce API changes
- [ ] Regenerated Golang and Python client if the PR introduce API changes

PR #429, opened 2023-07-13
Use pytest_configure to set mock responses as global variable

ariefrahmansyah:

In #404, we merge mock_oauth in client_test.py and merlin_test.py to one pytest fixture in conftest.py. However, there's an issue where the mock responses in conftest is not used by client_test or merlin_test because it was using different instance of Reponses object.

This PR introduce [pytest_configure()](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_configure) function where it sets mock responses as global variable that can be called by other files.

```release-note
NONE
```

**Checklist**

- [x] Updated unit test, integration, and/or e2e tests
- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/sdk/test/client_test.py
+++ b/python/sdk/test/client_test.py
@@ -18,7 +18,6 @@ from sys import version_info
 from unittest import mock
 
 import pytest
-from urllib3_mock import Responses
 
 import client as cl
 from client import ApiClient, Configuration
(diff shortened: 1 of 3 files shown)


krithika369 (approved):

PR #431, opened 2023-07-16
Set initial scale during redeployment of a model version

krithika369:

This PR implements setting the initial scale of a model version when it is redeployed, to match the current scale it is at. This is done using the [Knative Initial Scale](https://knative.dev/docs/serving/autoscaling/scale-bounds/#initial-scale) annotation, as the Predictor / Transformer specs do not directly support it. One caveat is that annotations applied to inference services will be applied to all components (predictor and transformer) and thus, it is not possible to individually control their initial scales (see related issue: https://github.com/kserve/kserve/issues/666). Thus, the max of the 2 values (current scale of predictor, and transformer, if enabled), will be used as the initial scale. Autoscaling will eventually correct the replicas as needed, and within the max threshold for the specific component.

Note that this change is only applicable to **serverless deployments**.

Main changes:
* `api/cluster/controller.go` - Add method `GetCurrentDeploymentScale` to the controller interface, to retrieve the current scale from the Knative revision. If patching an existing deployment that is serverless, get the current scale of the components and pass it down to the templater.
* `api/cluster/resource/templater.go` - `PatchInferenceServiceSpec` now takes in the current replicas of the existing inference service deployment and applies the relevant annotation.

Related Turing PR: https://github.com/caraml-dev/turing/pull/350

```release-note
When re-deploying a model version using serverless deployment, its current scale will be applied at start up.
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


config.yaml
```diff
@@ -69,6 +69,8 @@ StandardTransformerConfig:
     FeastBigtableURL: https://caraml.dev/feast/bigtable
   Kafka:
     Brokers: localhost:9092
+  Jaeger:
```
krithika369, line 72: Adding default value based on: https://github.com/caraml-dev/merlin/pull/417

python/sdk/setup.py
```diff
@@ -25,7 +25,7 @@
     "boto3>=1.9.84",
     "caraml-upi-protos>=0.3.1",
     "certifi>=2017.4.17",
-    "Click>=7.0",
+    "Click>=7.0,<8.1.4",
```
krithika369, line 28: Newer [versions](https://click.palletsprojects.com/en/8.1.x/changes/#changes) of this library has a breaking change causing errors in the CI. Eg: https://github.com/caraml-dev/merlin/actions/runs/5567016741/jobs/10168580897

So, pinning the version to <8.1.4

api/cluster/controller.go
```diff
@@ -142,75 +167,93 @@ func newController(kfservingClient kservev1beta1client.ServingV1beta1Interface,
...
+				deploymentScale.Predictor = &predictorScale
+			}
+			if _, ok := s.Status.Components[kservev1beta1.TransformerComponent]; ok {
+				if transformerScale, err := c.GetCurrentDeploymentScale(
+					ctx, modelService.Namespace,
+					s.Status.Components[kservev1beta1.TransformerComponent].LatestCreatedRevision); err != nil {
+					deploymentScale.Transformer = &transformerScale
+				}
+			}
+		}
```
ariefrahmansyah, line 229: Have you considered to merge getting prediction and transformer's deployment scale in one function?

So, in .Deploy() function, it would be something like this:

```
// Get current scale of the existing deployment
		deploymentScale := resource.DeploymentScale{}
		if modelService.DeploymentMode == deployment.ServerlessDeploymentMode ||
			modelService.DeploymentMode == deployment.EmptyDeploymentMode {
			deploymentScale, err = c.getCurrentDeploymentScale(ctx, modelService.Namespace,
				s.Status.Components)
			...
		}
```

Then, the function c.getCurrentDeploymentScale() will be the one that have the logic on whether the transformer component exists or not. I prefer to have .Deploy() function to have higher level of asbtraction.

api/cluster/controller.go
```diff
@@ -291,3 +334,18 @@ func (c *controller) ListPods(ctx context.Context, namespace, labelSelector stri
 func (c *controller) StreamPodLogs(ctx context.Context, namespace, podName string, opts *corev1.PodLogOptions) (io.ReadCloser, error) {
 	return c.clusterClient.Pods(namespace).GetLogs(podName, opts).Stream(ctx)
 }
+
+func (c *controller) GetCurrentDeploymentScale(ctx context.Context, namespace, knRevisionName string) (int, error) {
```
ariefrahmansyah, line 338: I think we can have this as private method since it's not being used by other package. Then we can remove it from Controller interface as well.

api/cluster/controller.go
```diff
@@ -142,75 +167,93 @@ func newController(kfservingClient kservev1beta1client.ServingV1beta1Interface,
 	}, nil
 }
 
-func (k *controller) Deploy(ctx context.Context, modelService *models.Service) (*models.Service, error) {
+func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (*models.Service, error) {
```
ariefrahmansyah, line 170: Thanks for renaming this receiver 😬

ariefrahmansyah (approved):
lgtm. with nit comments. thanks krithika!

api/cluster/controller.go
krithika369: I did consider that. The only downside to this is, it makes it harder to unit test it. Since the integration tests are also not able to test the initial scale annotation (as we don't return it from the [Deploy function](https://github.com/caraml-dev/merlin/blob/main/api/cluster/controller.go#L221)), I decided to keep it as a part of the interface and add a separate test case for it.

krithika369:
The CI is failing with this error which is strange - it looks like some outdated code 🤔 
Screenshot 2023-07-17 at 6 40 14 PM

I tried clearing the relevant CI cache and it didn't seem to help. Will make another dummy commit to re-try.

krithika369:
> The CI is failing with this error which is strange - it looks like some outdated code 🤔 Screenshot 2023-07-17 at 6 40 14 PM
> 
> I tried clearing the relevant CI cache and it didn't seem to help. Will make another dummy commit to re-try.

Think I finally figured out the issue. The checkout action is actually attempting to merge this PR against main, and the new test case introduced on main branch was inconsistent with the changes from this PR.

Screenshot 2023-07-18 at 9 57 29 AM

PR #433, opened 2023-07-19
Revert changes to caching library used in transformer

krithika369:

https://github.com/caraml-dev/merlin/pull/418 replaced the in-memory caching library used in the code with `github.com/patrickmn/go-cache` everywhere. However, this could be problematic for the transformer, as indicated by [this comment](https://github.com/caraml-dev/merlin/pull/418#issuecomment-1617219461). So, this PR reverts the changes to the caching library for the transformer, so it can continue to rely on `github.com/coocood/freecache`, which can bound the memory allocated for the caching.

```release-note
Replace Transformer caching library. Support for the transformer env var config CACHE_SIZE_IN_MB is now re-enabled.
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


deadlycoconuts (approved):
LGTM, though now that I read that [comment](https://github.com/caraml-dev/merlin/pull/418#issuecomment-1617219461), I'm wondering if whatever you [suggested](https://github.com/caraml-dev/merlin/pull/418#issuecomment-1621406343) subsequently would be insufficient 🤔

krithika369:
> LGTM, though now that I read that [comment](https://github.com/caraml-dev/merlin/pull/418#issuecomment-1617219461), I'm wondering if whatever you [suggested](https://github.com/caraml-dev/merlin/pull/418#issuecomment-1621406343) subsequently would be insufficient 🤔

Yes, it could be insufficient. To cover all possibilities, we must add a warning to the users in the release notes. But if they fail to take note of it, it could cause OOM. So, after thinking about it further, it seems best not to alter the existing behavior.

PR #434, opened 2023-07-21
Set DB logger level to silent

ariefrahmansyah:

In #413, when updating gorm version, the database's default logger level was set to info which made merlin log to be quite noisy. This PR set the logger level to silent.

```release-note
NONE
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


krithika369 (approved):
Thanks for this change @ariefrahmansyah !

PR #435, opened 2023-07-25
fix: delete query for deployments table

ariefrahmansyah:

In GORM 2.0, we must specify WHERE clause/function first before the DELETE function.

Reference: https://gorm.io/docs/delete.html#Delete-a-Record

This PR fixes failed deletion of model version that has existing deployments.

Fixes failed deleting model version with deployments.

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/storage/deployment_storage.go
+++ b/api/storage/deployment_storage.go
@@ -15,8 +15,9 @@
 package storage
 
 import (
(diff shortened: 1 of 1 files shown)


krithika369 (approved):
Thanks for catching this, @ariefrahmansyah !

PR #436, opened 2023-07-27
fix: Add precautionary logic for PDB

krithika369:

PDB support was introduced in https://github.com/caraml-dev/merlin/pull/420. In some scenarios, the PDB can create problems. As an example, if PDB is configured to have a minimum of 20% replicas available and there is only one replica in the deployment, it's not possible to drain that node. To prevent such issues, this PR introduces some precautionary checks, - ensuring that at least 1 replica of the deployment can be removed:

```
Let
===
# replicas: r
minAvailablePercent / 100: p

We want
=======
r - ceil(r * p) >= 1 // ceil because k8s rounds up to the nearest int
=> ceil(r * p) + 1 <= r
=> ceil(r * p) < r // We can say this because all terms in the above inequality are int
```

It's enough to check `ceil(r * p) < r` where r is the minReplicas because:

```
r - ceil(r * p) >= 1
=> r * (1 - ceil(p)) >= 1

We can see that, as r increases, LHS will increase. So, it's enough to ensure this inequality holds for the smallest possible value of r, which is the minReplicas configured for the deployment.
```

```release-note
Disable PDB when the min replicas is very low, so that it doesn't block node drain.
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cluster/controller.go
+++ b/api/cluster/controller.go
@@ -232,7 +232,7 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 	}
 
 	if c.deploymentConfig.PodDisruptionBudget.Enabled {
-		pdbs := c.createPodDisruptionBudgets(modelService, c.deploymentConfig.PodDisruptionBudget)
+		pdbs := createPodDisruptionBudgets(modelService, c.deploymentConfig.PodDisruptionBudget)
 		if err := c.deployPodDisruptionBudgets(ctx, pdbs); err != nil {
 			log.Errorf("unable to create pdb %v", err)
 			return nil, ErrUnableToCreatePDB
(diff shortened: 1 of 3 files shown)


api/cluster/controller.go
```diff
@@ -232,7 +232,7 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 	}
 
 	if c.deploymentConfig.PodDisruptionBudget.Enabled {
-		pdbs := c.createPodDisruptionBudgets(modelService, c.deploymentConfig.PodDisruptionBudget)
+		pdbs := createPodDisruptionBudgets(modelService, c.deploymentConfig.PodDisruptionBudget)
```
krithika369, line 235: Removing `createPodDisruptionBudgets`  from the controller so that it's easy to unit test it.

ariefrahmansyah (approved):
LGTM! Thanks for making it better, Krithika!

PR #440, opened 2023-08-01
Add FeastClientMaxConcurrentRequests to Transformer Simulator

deadlycoconuts:
This PR introduces a new field in the Merlin API server's config, `SimulatorFeastClientMaxConcurrentRequests`, which will be used to initialise a standard transformer executor whenever the transformer simulator is used. Previously, this value is not passed as options to the Feast transformer, causing the Hystrix max concurrent request value to default as `10`, which is much lower than the [default](https://github.com/caraml-dev/merlin/blob/main/api/pkg/transformer/feast/feature_retriever.go#L106) value which a regular standard transformer is initialised with, `100`.

By introducing this new field, all simulate requests will have the same max concurrent request limit set as 100 by default (unless specified otherwise by the Merlin API server's operator). 

A bug whereby a user might not experience the same behaviour from a regular standard transformer and from the simulator, if the number of concurrent requests to Feast is different.

Fixes #

NONE

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/config/config.go
+++ b/api/config/config.go
@@ -323,6 +323,8 @@ type StandardTransformerConfig struct {
 	Jaeger             JaegerConfig          `validate:"required"`
 	SimulationFeast    SimulationFeastConfig `validate:"required"`
 	Kafka              KafkaConfig           `validate:"required"`
+	// Simulator configs
+	SimulatorFeastClientMaxConcurrentRequests int `validate:"required" default:"100"`
 }
 
 // KafkaConfig configuration for publishing prediction log
--- a/api/service/transformer_simulation_service.go
+++ b/api/service/transformer_simulation_service.go
@@ -69,10 +69,11 @@ func (ts *transformerService) createTransformerExecutor(ctx context.Context, sim
 		executor.WithTraceEnabled(true),
 		executor.WithModelPredictor(executor.NewMockModelPredictor(mockModelResponseBody, mockModelRequestHeaders, simulationPayload.Protocol)),
 		executor.WithFeastOptions(feast.Options{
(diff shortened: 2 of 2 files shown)


api/config/config.go
```diff
@@ -323,6 +323,8 @@ type StandardTransformerConfig struct {
 	Jaeger             JaegerConfig          `validate:"required"`
 	SimulationFeast    SimulationFeastConfig `validate:"required"`
 	Kafka              KafkaConfig           `validate:"required"`
+	// Simulator configs
+	SimulatorFeastClientMaxConcurrentRequests int `validate:"required" default:"100"`
```
tiopramayudi, line 327: just curious, if we don't set the value, will it call the validate first or populate the default first?

tiopramayudi (commented):
add one question, the rest is LGTM. Thanks @deadlycoconuts for the fix

api/config/config.go
deadlycoconuts: It will call default first 😅 (so actually there's no real need for `validate`)

deadlycoconuts: Unless somehow someone tries to set it as `nil` in between when `default` is set and when `validate` is called.

`default` is set over here: https://github.com/caraml-dev/merlin/blob/main/api/cmd/api/main.go#L84, via `Load`, which calls `reflectViperConfig`, which then sets all the defaults with the `default` tag.

`validate` is only called in that same file I shared, at line 100, [`reflectViperConfig](https://github.com/caraml-dev/merlin/blob/main/api/cmd/api/main.go#L100.

tiopramayudi (approved):

PR #442, opened 2023-08-04
fix: Updating association tables correctly

ariefrahmansyah:

GORM V2 switched to using upsert to save associations, so saving changed fields won't works in V2. (to avoid save uncompleted associations to protect your data) [[ref](https://github.com/go-gorm/gorm/issues/3487#issuecomment-694889052)].

However, this breaks Merlin's behavior.

For example, when redeploying a model version with a standard transformer, the redeployment will not update the transformer configuration. The resulting SQL is:

```
INSERT INTO "transformers" (...) VALUES (...)
ON CONFLICT ("id")
DO UPDATE SET 
  "version_endpoint_id"="excluded"."version_endpoint_id"
RETURNING "id"
```

The ON CONFLICT DO UPDATE SET clause is ignoring other transformer fields.

This PR adds gorm session with FullSaveAssociations: true to gorm query so we will have the same behavior as gorm.v1 when updating associations table. [[ref](https://github.com/go-gorm/gorm/issues/3487#issuecomment-698303344)]. The resulting SQL is now:

```
INSERT INTO "transformers" (...) VALUES (...)
ON CONFLICT ("id")
DO UPDATE SET
  "updated_at"='2023-08-04 18:35:10.261',
  "enabled"="excluded"."enabled",
  "version_endpoint_id"="excluded"."version_endpoint_id",
  "transformer_type"="excluded"."transformer_type",
  "image"="excluded"."image",
  "command"="excluded"."command",
  "args"="excluded"."args",
  "resource_request"="excluded"."resource_request",
  "env_vars"="excluded"."env_vars"
RETURNING "id"
```

Fixes updating association tables

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/storage/version_endpoint_storage.go
+++ b/api/storage/version_endpoint_storage.go
@@ -54,7 +54,12 @@ func (v *versionEndpointStorage) Get(uuid uuid.UUID) (*models.VersionEndpoint, e
 
 func (v *versionEndpointStorage) Save(endpoint *models.VersionEndpoint) error {
 	sanitizeEndpoint(endpoint)
-	return v.db.Session(&gorm.Session{FullSaveAssociations: true}).Save(&endpoint).Error
+
+	if err := v.db.Save(&endpoint).Error; err != nil {
+		return err
+	}
+
+	return v.db.Save(endpoint.Transformer).Error
 }
 
 func sanitizeEndpoint(endpoint *models.VersionEndpoint) {
--- a/api/storage/version_endpoint_storage.go
+++ b/api/storage/version_endpoint_storage.go
@@ -59,7 +59,11 @@ func (v *versionEndpointStorage) Save(endpoint *models.VersionEndpoint) error {
 		return err
 	}
 
-	return v.db.Save(endpoint.Transformer).Error
+	if endpoint.Transformer != nil {
+		return v.db.Save(endpoint.Transformer).Error
+	}
+
+	return nil
 }
 
 func sanitizeEndpoint(endpoint *models.VersionEndpoint) {
--- a/api/storage/version_endpoint_storage.go
+++ b/api/storage/version_endpoint_storage.go
@@ -54,7 +54,7 @@ func (v *versionEndpointStorage) Get(uuid uuid.UUID) (*models.VersionEndpoint, e
 
 func (v *versionEndpointStorage) Save(endpoint *models.VersionEndpoint) error {
 	sanitizeEndpoint(endpoint)
-	return v.db.Save(&endpoint).Error
+	return v.db.Session(&gorm.Session{FullSaveAssociations: true}).Save(&endpoint).Error
 }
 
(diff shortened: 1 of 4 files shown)


deadlycoconuts (approved):
LGTM! 🚀 Thanks for catching and investigating this!

api/storage/version_endpoint_storage.go
```diff
@@ -54,7 +54,7 @@ func (v *versionEndpointStorage) Get(uuid uuid.UUID) (*models.VersionEndpoint, e
 
 func (v *versionEndpointStorage) Save(endpoint *models.VersionEndpoint) error {
 	sanitizeEndpoint(endpoint)
-	return v.db.Save(&endpoint).Error
+	return v.db.Session(&gorm.Session{FullSaveAssociations: true}).Save(&endpoint).Error
```
ariefrahmansyah, line 57: I have tried using GROM Clauses but to no avail, since clauses cannot be used to update the associations. @deadlycoconuts @krithika369

krithika369: @ariefrahmansyah will this also trigger a save on the `environments` table?

Instead of using `FullSaveAssociations`, I'm thinking if we should explicitly issue `Save` on the transformer record. Like:

```golang
func (v *versionEndpointStorage) Save(endpoint *models.VersionEndpoint) error {
    sanitizeEndpoint(endpoint)
    if err := v.db.Save(&endpoint).Error; err != nil {
        return err
    }
    
    return v.db.Save(endpoint.Transformer).Error
}
```

krithika369 (approved):
LGTM. Thanks, @ariefrahmansyah !

luizcarlosrodrigues:
I´m still having problem with this same code because when i use the statement below, if my association are just updating one column, it updates all columns with null values:

`database.Db.WithContext(ctx).Session(&gorm.Session{FullSaveAssociations:true}).Clauses(clause.Returning{}).Updates(&client)`

For this example I have a table and an association that is updating just one field and gorm is updating all the other fields to null

```
type Persons struct {
	ClientId          *string            `gorm:"column:uuid;primaryKey;not null;default:null" json:"clientId"`
	Name              *string            `gorm:"column:name;not null;default:null" json:"name"`
	Documents         []*PersonDocuments `gorm:"foreignKey:PersonId" json:"documents,omitempty"`
}
```

```
type PersonAddress struct {
	AddressId        *string `gorm:"column:uuid;primaryKey;not null;default:null" json:"addressId"`
	PersonId         *string `gorm:"column:person_id;not null;default:null" json:"clientId"`
	CountryCode      *string `gorm:"column:country_code;not null;default:null" json:"countryCode"`
	StatusCode       *int    `gorm:"column:status_code;not null;default:1" json:"statusCode"`
}
```
I´m sending the person object but i just need to update the statusCode of the Address association

PR #443, opened 2023-08-08
fix: order the output rows based on the order of the input entities

khorshuheng:
Currently, the rows for the Feast Table are not guaranteed to be ordered based on the original order of the entities, which is a counter intuitive behaviour, consider that the official feast serving client returns the result based on the original entities order. It also affect usability as users need to perform extra work to reorder the table based on the original entity rows.

Fixes #

```release-note
Feast Table output rows are now guaranteed to follow the order of entity rows.
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation

### DIFF
--- a/api/pkg/transformer/feast/call.go
+++ b/api/pkg/transformer/feast/call.go
@@ -60,7 +60,7 @@ func newCall(
 }
 
 // do create request to feast and return the result as table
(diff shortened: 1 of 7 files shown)


ariefrahmansyah (approved):
LGTM. Thanks, @khorshuheng!

PR #447, opened 2023-08-11
fix: PDB not deleted and multiple PDBs targeting single deployment

ariefrahmansyah:

This PR fixes two issues:

1. Fail to delete predictor PDB because the resource request is not passed to createPodDisruptionBudgets() function hence, the pdb for the predictor is not appended to pdbs to be deleted because it's not meet this [condition](https://github.com/caraml-dev/merlin/blob/main/api/cluster/pdb.go#L78):
```
if modelService.ResourceRequest != nil &&
```

2. Because the PDB is not deleted, there would be multiple PDBs in the cluster and all of them would be targeting the same active deployments because of incorrect Selector spec.

For example, we have deployed sklearn-sample-r version 54 and 55. Then we undeploy version 54. The remaining pods are only for version 55:

```
$ kubectl get po
NAME                                                              READY   STATUS    RESTARTS   AGE
sklearn-sample-r-55-predictor-default-779754655d-cgvqj            1/1     Running   0          42m
sklearn-sample-r-55-predictor-default-779754655d-dw6hc            1/1     Running   0          42m
```

 However, the PDB for version 54 still exists:

```
$ kubectl get pdb
NAME                             MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
sklearn-sample-r-54-model-pdb    20%             N/A               2                     44m
sklearn-sample-r-55-model-pdb    20%             N/A               3                     44m
```

And both of them have the same selector:

```
  spec:
    minAvailable: 20%
    selector:
      matchLabels:
        gojek.com/app: sklearn-sample-s
        gojek.com/component: model-version
        gojek.com/environment: staging
        gojek.com/orchestrator: merlin
        gojek.com/stream: dsp
        gojek.com/team: dsp
```

This PR adds more labels for the selector.

Fixes fail to delete predictor PDB
Fixes multiple PDBs targeting single deployment

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/cluster/pdb.go
+++ b/api/cluster/pdb.go
@@ -25,7 +25,7 @@ type PodDisruptionBudget struct {
 
 func NewPodDisruptionBudget(modelService *models.Service, componentType string, pdbConfig config.PodDisruptionBudgetConfig) *PodDisruptionBudget {
 	return &PodDisruptionBudget{
-		Name:                     fmt.Sprintf("%s-%s-%s", modelService.Name, componentType, models.PDBComponentType),
+		Name:                     fmt.Sprintf("%s-%s-%s", modelService.ModelName, componentType, models.PDBComponentType),
 		Namespace:                modelService.Namespace,
 		Labels:                   modelService.Metadata.ToLabel(),
 		MaxUnavailablePercentage: pdbConfig.MaxUnavailablePercentage,
--- a/api/cluster/pdb.go
+++ b/api/cluster/pdb.go
@@ -24,10 +24,14 @@ type PodDisruptionBudget struct {
 }
 
 func NewPodDisruptionBudget(modelService *models.Service, componentType string, pdbConfig config.PodDisruptionBudgetConfig) *PodDisruptionBudget {
+	labels := modelService.Metadata.ToLabel()
+	labels["component"] = componentType
+	labels["serving.kserve.io/inferenceservice"] = modelService.Name
+
 	return &PodDisruptionBudget{
-		Name:                     fmt.Sprintf("%s-%s-%s", modelService.ModelName, componentType, models.PDBComponentType),
+		Name:                     fmt.Sprintf("%s-%s-%s", modelService.Name, componentType, models.PDBComponentType),
 		Namespace:                modelService.Namespace,
-		Labels:                   modelService.Metadata.ToLabel(),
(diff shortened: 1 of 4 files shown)


api/models/container.go
```diff
@@ -27,6 +27,7 @@ const (
 
 	ImageBuilderComponentType     = "image_builder"
 	ModelComponentType            = "model"
+	PredictorComponentType        = "predictor"
```
ariefrahmansyah, line 30: I'm adding a new predictor component type because Kserve is using `predictor` for their inference service component labels.

api/cluster/pdb.go
```diff
@@ -78,7 +82,7 @@ func createPodDisruptionBudgets(modelService *models.Service, pdbConfig config.P
 	if modelService.ResourceRequest != nil &&
 		math.Ceil(float64(modelService.ResourceRequest.MinReplica)*
 			minAvailablePercent) < float64(modelService.ResourceRequest.MinReplica) {
-		predictorPdb := NewPodDisruptionBudget(modelService, models.ModelComponentType, pdbConfig)
+		predictorPdb := NewPodDisruptionBudget(modelService, models.PredictorComponentType, pdbConfig)
```
deadlycoconuts, line 85: Wow thanks for catching this 😅

api/models/container.go
deadlycoconuts: Just curious, is it possible to remove `ModelComponentType` completely? (or replace/rename/refactor it with the new `PredictorComponentType`?) I think all the previous instances where we're using that constant is when they were referring to a predictor right?

deadlycoconuts (approved):
LGTM! 🚀 Thanks for looking at these fixes; I just left one small comment but it's not too important 😅

api/models/container.go
ariefrahmansyah: Yes, it's possible. For Merlin itself, it's mainly being used for Log UI. I have also considered this and I think it should be addressed in separate PR.

deadlycoconuts: Ahh I see, okay then! Thanks for clarifying!

krithika369 (approved):
LGTM. Thanks for the quick fix, @ariefrahmansyah !

PR #452, opened 2023-08-17
fix: allow post method for standard transformer simulator

khorshuheng:
POST method for standard transformer has not been whitelisted

Fixes #

```release-note
Fix permission issue where users are not allowed to make changes to standard transformer simulator.
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/middleware/authorization_test.go
+++ b/api/middleware/authorization_test.go
@@ -30,6 +30,7 @@ func TestAuthorizer_RequireAuthorization(t *testing.T) {
 	}{
 		{"All authenticated users can list projects", "/projects", "GET", false},
 		{"All authenticated users can list environments", "/environments", "GET", false},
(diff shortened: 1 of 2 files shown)


api/middleware/authorization_test.go
```diff
@@ -30,6 +30,8 @@ func TestAuthorizer_RequireAuthorization(t *testing.T) {
 	}{
 		{"All authenticated users can list projects", "/projects", "GET", false},
 		{"All authenticated users can list environments", "/environments", "GET", false},
+		{"All authenticated users can read simulator", "/standard-transformer/simulate", "GET", false},
```
krithika369, line 33: Why do we still have the GET here ? We only need POST for this endpoint, right?

deadlycoconuts (approved):
Thanks for this fix!

krithika369 (approved):
LGTM. Thanks for this fix!

PR #455, opened 2023-08-18
Enable model redeployments with SDK

deadlycoconuts:
This PR introduces a minor tweak the the SDK which allows the redeployment of existing Merlin model versions using the `deploy` method.

Previously, performing the following operations below will result in the second call to `deploy` returning a `{"error":"There is 'running' deployment for the model version"}` error: 
```python
    # Create new model version
    with merlin.new_model_version() as v:
        merlin.log_model(model_dir=model_dir)

    # Deploy 
    initial_endpoint = merlin.deploy(v)

    # Deploy using raw_deployment
    new_endpoint = merlin.deploy(v, deployment_mode=DeploymentMode.RAW_DEPLOYMENT)
```

In the new implementation introduced by this PR, the second `deploy` call will simply redeploy the router version (represented by `v`) with the new configuration (i.e. with the raw deployment setting).

### Unrelated Modifications
This PR also introduces an additional check in the Update Endpoint controller to ensure that an endpoint being updated is not in the `pending` state (attempting to do so will return a 400 error). This changes the current behaviour of this endpoint (the Update Endpoint endpoint 🥴), as you might notice from some of the changes made to the unit tests.

Fixes #

```release-note
1. Calling `ModelVersion.deploy` with an already deployed model version passed as a parameter will redeploy the endpoint with the new configuration of the given version
2. Updating an endpoint that is in the pending state is no longer allowed
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


python/sdk/merlin/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-VERSION = "0.25.0"
+VERSION = "0.31.3"
```
deadlycoconuts, line 15: Somehow this version hasn't been updated for a loooong time. I don't know when we'll cut these SDK changes into a new release, but I've just bumped up this version to the latest tag that we have.

docs/connecting-to-merlin/python-sdk.md
```diff
@@ -15,6 +15,10 @@ from merlin.model import ModelType
 # Connect to an existing Merlin deployment
 merlin.set_url("merlin.example.com")
 
+# Set the active model to the name given by parameter, if the model with the given name is not found, a new model will 
+# be created.
+merlin.set_model("example-model", ModelType.PYFUNC)
+
```
deadlycoconuts, line 21: Added this section here because trying to actually call `list_model_endpoints` without setting a model will throw an error.

api/api/version_endpoints_api.go
```diff
@@ -392,9 +392,13 @@ func validateUpdateRequest(prev *models.VersionEndpoint, new *models.VersionEndp
...
 	}
 
+	if prev.Status == models.EndpointPending {
+		return fmt.Errorf("Updating endpoint status to %s is not allowed when the endpoint is currently in the pending state", new.Status)
+	}
+
 	if new.Status != prev.Status {
 		if prev.Status == models.EndpointServing {
-			return fmt.Errorf("Updating endpoint status to %s is not allowed when the endpoint is in serving state", new.Status)
+			return fmt.Errorf("Updating endpoint status to %s is not allowed when the endpoint is currently in the serving state", new.Status)
```
deadlycoconuts, line 401: Made this change to make the error message clearer about which state it is referring to.

python/sdk/merlin/version.py
ariefrahmansyah: Thanks for updating this. This line is actually being updated when GitHub CI releases new SDK (refer to this line: https://github.com/caraml-dev/merlin/blob/main/.github/workflows/release.yml#L40)

deadlycoconuts: Ahh... thanks for sharing that! Seems like the last time that GitHub job was run successfully was a year ago 👀 https://github.com/caraml-dev/merlin/actions/workflows/publish-python-sdk.yml 

There's probably something wrong with the GitHub CICD workflow; lemme go fix that

deadlycoconuts: Ahhh got it, there's actually no need to update that value 😅 should I just set it as `VERSION = "0.0.0"` to make it obvious it doesn't need to be bumped?

deadlycoconuts:
Unpopular opinion: can I `black` the entire Merlin SDK repo? Seeing how the `model_test.py` file got reformatted in https://github.com/caraml-dev/merlin/pull/456/files, I was wondering if we should just stick to some standard formatting :/

python/sdk/merlin/version.py
ariefrahmansyah: -> should I just set it as VERSION = "0.0.0" to make it obvious it doesn't need to be bumped?

That's a good idea. I'm fine with that. 👍🏻

ariefrahmansyah:
> Unpopular opinion: can I `black` the entire Merlin SDK repo? Seeing how the `model_test.py` file got reformatted in https://github.com/caraml-dev/merlin/pull/456/files, I was wondering if we should just stick to some standard formatting :/

Doing this will make the PR huge with unrelated changes to the first intention. I rather 1) have separate formatting PR or 2) only format the file you originally modify for this PR.

ariefrahmansyah (commented):
Zi Yi, can you also add new integration test for redeployment case?

deadlycoconuts:
> 1) have separate formatting PR

Let's do this then :) But in any case, I just updated my PR to match the formatting in that test file that was modified in that PR, so it's not urgent.

deadlycoconuts:
> Zi Yi, can you also add new integration test for redeployment case?

Just added a new integration test for the model redeployment workflow :)

python/sdk/test/integration_test.py
```diff
@@ -947,5 +946,55 @@ def test_deployment_mode_for_serving_model(integration_test_url, project_name, u
...
+    assert endpoint.autoscaling_policy.metrics_type == MetricsType.RPS
+    assert endpoint.autoscaling_policy.target_value == 20
+
+    # Deploy v2 using raw_deployment with CPU autoscaling policy
+    new_endpoint = merlin.deploy(v1, autoscaling_policy=merlin.AutoscalingPolicy(
+         metrics_type=merlin.MetricsType.CPU_UTILIZATION,
+         target_value=10))
+
+    resp = requests.post(f"{new_endpoint.url}", json=tensorflow_request_json)
+
```
ariefrahmansyah, line 987: One small request, Zi Yi, can we `assert endpoint.url == new_endpoint.url` as well?

deadlycoconuts: Done! :)

ariefrahmansyah (approved):
LGTM! Thanks @deadlycoconuts!

PR #464, opened 2023-09-14
Remove Python3.7 support in merlin-sdk

krithika369:

Python 3.7 has reached its end of life ([ref](https://devguide.python.org/versions/)). This MR removes support for Python 3.7 from the SDK. For the time being the Pyfunc model service and batch job apps will still support Python 3.7 - this is achieved by pinning their dependency on the `merlin-sdk`. A future MR will also remove support for Python 3.7 from these components.

## Summary of changes
WIP

Related PR for Turing: https://github.com/caraml-dev/turing/pull/362

Fixes #

```release-note
Remove support for Python 3.7 from merlin-sdk
```

**Checklist**

- [ ] Tested locally
- [x] Updated documentation

### DIFF
--- a/.github/workflows/merlin.yml
+++ b/.github/workflows/merlin.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"] #TODO: Remove Python 3.7 support
     env:
       PIPENV_DEFAULT_PYTHON_VERSION: ${{ matrix.python-version }}
     steps:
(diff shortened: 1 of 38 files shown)


python/batch-predictor/setup.py
```diff
@@ -21,11 +21,12 @@
...
-merlin_sdk_package = "merlin-sdk"
-for index, item in enumerate(REQUIRE):
-    if merlin_sdk_package in item:
-        REQUIRE[index] = f"{merlin_sdk_package} @ file://localhost/{merlin_path}#egg={merlin_sdk_package}"
+# TODO: Uncomment below lines after Pyfunc server stops supporting Python 3.7
+# merlin_path = os.path.join(os.getcwd(), "../sdk")
+# merlin_sdk_package = "merlin-sdk"
+# for index, item in enumerate(REQUIRE):
+#     if merlin_sdk_package in item:
+#         REQUIRE[index] = f"{merlin_sdk_package} @ file://localhost/{merlin_path}#egg={merlin_sdk_package}"
```
ariefrahmansyah, line 29: If my understanding is correct, you commented this block because batch predictor is using 3.7 but merlin-sdk requires >3.7 hence if we keep this block the installation will fail.

ariefrahmansyah (approved):
LGTM! Thanks, Krithika!

python/batch-predictor/setup.py
krithika369: That is correct. For now, I've pinned the Pyfunc server and batch predictor dependency on `merlin-sdk` to `0.33.0`, which is the last version that's supporting Python 3.7.

PR #466, opened 2023-09-21
fix(sdk): Fix SDK model redeployment

deadlycoconuts:
This PR introduces a fix to PR #455, which itself introduced a new feature to allow the redeployment of existing Merlin model versions using the `deploy` method. 

While this new feature assumes that a user would be able to only **_specify fields he/she intends to update_** in the model version configuration when calling the `deploy` method, the existing implementation however unexpectedly overwrites some of the existing configuration with certain default configuration values if they are not explicitly specified. These configuration values include:
- protocol
- deployment mode
- resources
- autoscaling policy

This happens because the 'deploy' method indirectly sets default values for the aforementioned configuration if they aren't specified (this was primarily intended to make the **initial** deployment process simpler so users wouldn't have to specify too many config values), and does not distinguish between a user calling `deploy` on an existing model version or on a new model version.

This PR thus makes changes to the SDK to set these defaults selectively. 

A separate change is made to the `override` method used in the Merlin API server update model version endpoint to override the existing model protocol ONLY when the new model version explicitly specifies a protocol. This change is a **_partial reversion_** of PR #300 (change 2) which means that from now on users need to explicitly specify the HTTP protocol (as opposed to leaving it as `null`) when updating a UPI model to use HTTP instead. It's ultimately a minor change since the main fix that had been introduced in that PR involves change 3 (in that PR).

Fixes #455

```release-note
Users need to explicitly specify the HTTP protocol (as opposed to leaving it as `null`) when updating a UPI model to use HTTP instead
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/service/version_endpoint_service.go
+++ b/api/service/version_endpoint_service.go
@@ -231,13 +231,10 @@ func (k *endpointService) override(left *models.VersionEndpoint, right *models.V
 		left.EnvVars = models.MergeEnvVars(left.EnvVars, right.EnvVars)
 	}
 
-	// default to HttpJson if not provided
-	if right.Protocol == "" {
-		right.Protocol = protocol.HttpJson
-	}
-
 	// override protocol
-	left.Protocol = right.Protocol
+	if right.Protocol != "" {
+		left.Protocol = right.Protocol
+	}
 
 	return nil
 }
--- a/api/service/version_endpoint_service.go
+++ b/api/service/version_endpoint_service.go
@@ -236,6 +236,11 @@ func (k *endpointService) override(left *models.VersionEndpoint, right *models.V
 		left.Protocol = right.Protocol
 	}
 
+	// set default protocol if it is not specified explicitly
+	if left.Protocol == "" {
+		left.Protocol = protocol.HttpJson
+	}
+
 	return nil
 }
 
--- a/python/sdk/merlin/model.py
+++ b/python/sdk/merlin/model.py
@@ -1095,16 +1095,16 @@ class ModelVersion:
         target_deployment_mode = None
         if deployment_mode is None:
             if current_endpoint is None:
-                target_deployment_mode = DeploymentMode.SERVERLESS
+                target_deployment_mode = DeploymentMode.SERVERLESS.value
         else:
-            target_deployment_mode = deployment_mode
+            target_deployment_mode = deployment_mode.value
 
         target_protocol = None
         if protocol is None:
             if current_endpoint is None:
-                target_protocol = Protocol.HTTP_JSON
+                target_protocol = Protocol.HTTP_JSON.value
         else:
-            target_protocol = protocol
+            target_protocol = protocol.value
 
         target_env_name = environment_name
         if target_env_name is None:
--- a/python/sdk/merlin/model.py
+++ b/python/sdk/merlin/model.py
@@ -1224,7 +1224,6 @@ class ModelVersion:
             autoscaling_policy=target_autoscaling_policy,
             protocol=target_protocol,
         )
-        current_endpoint = self.endpoint
         if current_endpoint is not None:
             # This allows a serving deployment to be update while it is serving
             if current_endpoint.status == Status.SERVING:
--- a/python/sdk/merlin/model.py
+++ b/python/sdk/merlin/model.py
@@ -1095,10 +1095,10 @@ class ModelVersion:
 
         target_deployment_mode = None
         target_protocol = None
-        target_env_name = None
+        target_env_name = ModelVersion._get_default_target_env_name(env_list)
         target_resource_request = None
         target_autoscaling_policy = None
-        target_env_vars = []
+        target_env_vars: List[client.models.Environment] = []
         target_transformer = None
         target_logger = None
 
@@ -1107,10 +1107,9 @@ class ModelVersion:
         if current_endpoint is None:
             target_deployment_mode = DeploymentMode.SERVERLESS.value
             target_protocol = Protocol.HTTP_JSON.value
-            target_env_name = ModelVersion._get_default_target_env_name(env_list)
             target_resource_request = ModelVersion._get_default_resource_request(target_env_name, env_list)
             target_autoscaling_policy = ModelVersion._get_default_autoscaling_policy(
-                deployment_mode.value if deployment_mode.value else target_deployment_mode
+                deployment_mode.value if deployment_mode is not None else target_deployment_mode
             )
 
         if deployment_mode is not None:
@@ -1587,7 +1586,7 @@ class ModelVersion:
         return target_env_name
 
     @staticmethod
-    def _get_default_resource_request(env_name: str, env_list: List[client.models.Environment]) -> ResourceRequest:
+    def _get_default_resource_request(env_name: str, env_list: List[client.models.Environment]) -> client.ResourceRequest:
         resource_request = None
         for env in env_list:
             if env.name == env_name:
@@ -1603,10 +1602,12 @@ class ModelVersion:
             raise ValueError("default resource request not found in the environment config")
 
         resource_request.validate()
-        return resource_request
+        return client.ResourceRequest(
+            resource_request.min_replica, resource_request.max_replica,
+            resource_request.cpu_request, resource_request.memory_request)
 
     @staticmethod
-    def _get_default_autoscaling_policy(deployment_mode: DeploymentMode) -> client.AutoscalingPolicy:
+    def _get_default_autoscaling_policy(deployment_mode: str) -> client.AutoscalingPolicy:
         # If a previous model version does not exist, we set up a default autoscaling_policy
         if deployment_mode == DeploymentMode.RAW_DEPLOYMENT:
             autoscaling_policy = RAW_DEPLOYMENT_DEFAULT_AUTOSCALING_POLICY
@@ -1630,15 +1631,14 @@ class ModelVersion:
     def _create_transformer_spec(transformer: Transformer, target_env_name: str, env_list: List[client.models.Environment]) -> client.Transformer:
         resource_request = transformer.resource_request
         if resource_request is None:
-            resource_request = ModelVersion._get_default_resource_request(target_env_name, env_list)
+            target_resource_request = ModelVersion._get_default_resource_request(target_env_name, env_list)
         else:
             resource_request.validate()
+            target_resource_request = client.ResourceRequest(
+                resource_request.min_replica, resource_request.max_replica,
+                resource_request.cpu_request, resource_request.memory_request)
 
-        target_resource_request = client.ResourceRequest(
-            resource_request.min_replica, resource_request.max_replica,
-            resource_request.cpu_request, resource_request.memory_request)
-
-        target_env_vars = []
+        target_env_vars: List[client.models.Environment] = []
         if transformer.env_vars is not None:
             target_env_vars = ModelVersion._add_env_vars(target_env_vars, transformer.env_vars)
 
--- a/python/sdk/merlin/model.py
+++ b/python/sdk/merlin/model.py
@@ -1085,9 +1085,9 @@ class ModelVersion:
         :param env_vars: List of environment variables to be passed to the model container.
         :param transformer: The service to be deployed alongside the model for pre/post-processing steps.
         :param logger: Response/Request logging configuration for model or transformer.
-        :param deployment_mode: mode of deployment for the endpoint (default: DeploymentMode.SERVERLESS)
+        :param deployment_mode: mode of deployment for the endpoint (default: None)
         :param autoscaling_policy: autoscaling policy to be used for the deployment (default: None)
-        :param protocol: protocol to be used for deploying the model (default: HTTP_JSON)
+        :param protocol: protocol to be used for deploying the model (default: None)
         :return: VersionEndpoint object
         """
         current_endpoint = self.endpoint
@@ -1607,8 +1607,7 @@ class ModelVersion:
 
     @staticmethod
     def _get_default_autoscaling_policy(deployment_mode: str) -> client.AutoscalingPolicy:
-        # If a previous model version does not exist, we set up a default autoscaling_policy
-        if deployment_mode == DeploymentMode.RAW_DEPLOYMENT:
+        if deployment_mode == DeploymentMode.RAW_DEPLOYMENT.value:
             autoscaling_policy = RAW_DEPLOYMENT_DEFAULT_AUTOSCALING_POLICY
         else:
             autoscaling_policy = SERVERLESS_DEFAULT_AUTOSCALING_POLICY
--- a/python/sdk/merlin/model.py
+++ b/python/sdk/merlin/model.py
@@ -1565,7 +1565,7 @@ class ModelVersion:
     def _get_env_list(self) -> List[client.models.Environment]:
         return EnvironmentApi(self._api_client).environments_get()
 
-    def _get_endpoint_in_environment(self, environment_name: str) -> Optional[VersionEndpoint]:
+    def _get_endpoint_in_environment(self, environment_name: Optional[str]) -> Optional[VersionEndpoint]:
         """
         Return the FIRST endpoint of this model version that is deployed in the environment specified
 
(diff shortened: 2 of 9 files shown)


api/service/version_endpoint_service.go
```diff
@@ -231,13 +231,10 @@ func (k *endpointService) override(left *models.VersionEndpoint, right *models.V
...
-	// default to HttpJson if not provided
-	if right.Protocol == "" {
-		right.Protocol = protocol.HttpJson
-	}
-
 	// override protocol
-	left.Protocol = right.Protocol
+	if right.Protocol != "" {
+		left.Protocol = right.Protocol
+	}
```
deadlycoconuts, line 237: This change is a partial reversion of the changes made in https://github.com/caraml-dev/merlin/pull/300 (change no. 2).

python/sdk/merlin/model.py
```diff
@@ -1029,6 +1029,21 @@ def deploy(self, environment_name: str = None,
         :param protocol: protocol to be used for deploying the model (default: HTTP_JSON)
         :return: VersionEndpoint object
         """
+        current_endpoint = self.endpoint
```
deadlycoconuts, line 1032: This line checks for the existence of a prior endpoint that has been deployed.

python/sdk/test/integration_test.py
```diff
@@ -1115,6 +1123,13 @@ def test_redeploy_model(integration_test_url, project_name, use_google_oauth, re
 
     # Check that the endpoint remains the same
     assert endpoint.url == new_endpoint.url
+
+    # Check that the deployment configs of v2 have remained the same as those used in v1
+    assert endpoint.resource_request.cpu_request == "123m"
+    assert endpoint.resource_request.memory_request == "234Mi"
+    assert endpoint.env_vars == {"green": "TRUE"}
+    assert endpoint.deployment_mode == DeploymentMode.RAW_DEPLOYMENT
```
deadlycoconuts, line 1131: Additional checks to ensure the original configs do not get overwritten by the default configs.

api/service/version_endpoint_service.go
deadlycoconuts: That's a good point 🤔 I think after looking at the code, it seems like the only this `override` method is only [called](https://github.com/caraml-dev/merlin/blob/072325c6e1ae1226216db7186757b22a391f6c72/api/service/version_endpoint_service.go#L121) to update 2 types of endpoint configs (the `left` argument), 1) an existing endpoint config stored in the DB and 2) a 'default' endpoint config that will be overwritten by a user request. 

In case 2, we're actually already [setting](https://github.com/caraml-dev/merlin/blob/main/api/models/version_endpoint.go#L103) this `Protocol` field as `protocol.HttpJson` by default so `left.Protocol` wouldn't be empty in such cases.

For case 1, well, I don't think it's possible we'll encounter any empty strings for the `Protocol` field of the endpoints retrieved from the DB, if we assume that the mechanism mentioned in case 2 is working correctly 😅 Unless maybe there's some DB corruption or something similar 🤔

So yeah I don't really think it's necessary to include the extra step to set a default value for `left.Protocol` (it seems a little strange to set a default value in a method called 'override' also. What do you think?

python/sdk/test/integration_test.py
deadlycoconuts: Ahh, that's actually not how the API server works right now 👀 It seems like the intended behaviour (from the implementation in the API server) for the update endpoint workflow is to only 'override' or 'add on' additional configs to the existing endpoint configs. In the case of the transformer, if there's a transformer in the original 'configs' but no transformer specified in the new 'configs', the existing transformer will simply [not get overwritten](https://github.com/caraml-dev/merlin/blob/072325c6e1ae1226216db7186757b22a391f6c72/api/service/version_endpoint_service.go#L200) and remain in the newly deployed endpoint. 

Actually this behaviour exists for all the other endpoint configs, even the 'null-able' ones, like env vars and the logger. I suppose this is a design choice for the update endpoint workflow and not a bug? In order to remove the transformer, env vars or the logger, I think user has to create an entirely new model version right? 🤔

python/sdk/merlin/model.py
```diff
@@ -1113,10 +1145,9 @@ def deploy(self, environment_name: str = None,
...
                                           logger=target_logger,
-                                          deployment_mode=deployment_mode.value,
-                                          autoscaling_policy=client.AutoscalingPolicy(autoscaling_policy.metrics_type.value,
-                                                                                      autoscaling_policy.target_value),
-                                          protocol=protocol.value
+                                          deployment_mode=target_deployment_mode,
+                                          autoscaling_policy=target_autoscaling_policy,
+                                          protocol=target_protocol,
                                           )
         current_endpoint = self.endpoint
```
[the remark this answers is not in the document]
deadlycoconuts: Oh yeah wow nice catch! :D Removing this! Thanks!

python/sdk/merlin/model.py
```diff
@@ -1029,6 +1029,21 @@ def deploy(self, environment_name: str = None,
...
+                target_deployment_mode = DeploymentMode.SERVERLESS.value
+        else:
+            target_deployment_mode = deployment_mode.value
+
+        target_protocol = None
+        if protocol is None:
+            if current_endpoint is None:
+                target_protocol = Protocol.HTTP_JSON.value
+        else:
+            target_protocol = protocol.value
```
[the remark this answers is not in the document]
deadlycoconuts: Oh yes, that's an awesome suggestion 😁 I was also thinking the `deploy` method was starting to look a little messy with the setting of each parameter manually but yeah I think your suggestion looks really clean; let me try refactoring this a little, thanks!

python/sdk/merlin/model.py
```diff
@@ -1029,82 +1029,67 @@ def deploy(self, environment_name: str = None,
...
-                        env.default_resource_request.memory_request,
-                        )
-
-            # This case is when the default resource request is not specified in the environment config
-            if resource_request is None:
-                raise ValueError("resource request must be specified")
-
-        resource_request.validate()
+        # target_env_name can never be set as null
+        target_env_name = environment_name if environment_name is not None else ModelVersion._get_default_target_env_name(env_list)
```
deadlycoconuts, line 1036: Not sure if this is an intended or not, but it was never possible to make API calls via the SDK with `environment_name` being `None` or `""`. When creating or updating an endpoint, the SDK would always either use the user-given env name, or set a default env name, or raise an error if it can't find one.

The behaviour of this field is also different from the other fields when updating an endpoint - you can't leave the new endpoint config's environment name empty and expect the API server to redeploy the endpoint in the same environment (an error will be returned from the API server instead).

As such, since this field works differently from the other config fields (e.g. deployment mode, protocol, logger, etc.), I'm assigning the value to this variable separately.

python/sdk/merlin/model.py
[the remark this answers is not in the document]
deadlycoconuts: Okay I refactored the `deploy` method quite a bit so now it looks like what you had suggested, and also by creating some helper functions here and there to make the method easier to understand.

python/sdk/merlin/model.py
```diff
@@ -1042,42 +1057,51 @@ def deploy(self, environment_name: str = None,
...
+            target_resource_request = client.ResourceRequest(
+                resource_request.min_replica, resource_request.max_replica,
+                resource_request.cpu_request, resource_request.memory_request)
 
-        target_resource_request = client.ResourceRequest(
-            resource_request.min_replica, resource_request.max_replica,
-            resource_request.cpu_request, resource_request.memory_request)
+            if resource_request.gpu_request is not None and resource_request.gpu_name is not None:
+                env_api = EnvironmentApi(self._api_client)
+                env_list = env_api.environments_get()
```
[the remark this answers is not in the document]
deadlycoconuts: That's a good point, while refactoring the `deploy` method, I actually went to simplify it also so it does not repeat the ListEnvironments API call multiple times. It now calls that API once and reuses the response of that call throughout the entire `deploy` method.

python/sdk/merlin/model.py
```diff
@@ -1029,82 +1029,67 @@ def deploy(self, environment_name: str = None,
         :param protocol: protocol to be used for deploying the model (default: HTTP_JSON)
         :return: VersionEndpoint object
         """
+        current_endpoint = self.endpoint
+        env_list = self._get_env_list()
```
deadlycoconuts, line 1033: This call reduces all the redundant calls to the ListEnvironments API to just 1 call at the start of the entire method.

python/sdk/merlin/model.py
```diff
@@ -1525,6 +1469,82 @@ def _wait_build_complete(self, logs):
...
+                f"env_vars should be dictionary, got: {type(new_env_vars)}")
+
+        if len(new_env_vars) > 0:
+            for name, value in new_env_vars.items():
+                target_env_vars.append(
+                    client.EnvVar(str(name), str(value)))
+        return target_env_vars
+
+    @staticmethod
+    def _create_transformer_spec(transformer: Transformer, target_env_name: str, env_list: List[client.models.Environment]) -> client.Transformer:
```
deadlycoconuts, line 1529: This is actually almost the same as the original `create_transformer_spec` method - the only changes I made was make this method static and reused some other helper methods.

api/service/version_endpoint_service.go
ariefrahmansyah: When we added the `protocol` column to version_endpoints and model_endpoints tables, we didn't specify the default value ([ref](https://github.com/caraml-dev/merlin/blob/main/db-migrations/29_protocol.up.sql)). So, it's possible that the `protocol` field is empty here, hence the `left.Protocol` would also be empty. I think it's better to set the default value for `left.Protocol` if it's empty.

python/sdk/merlin/model.py
ariefrahmansyah: In CreateEndpoint API, if env_name is empty, we have logic to get default environment ([ref](https://github.com/caraml-dev/merlin/blob/main/api/api/version_endpoints_api.go#L138)). However, for UpdateEndpoint API, the env_name must be specified ([ref](https://github.com/caraml-dev/merlin/blob/main/api/api/version_endpoints_api.go#L262)). So, I think for redeployment, the `target_env_name` must be specified with current_endpoint's env_name if it's None.

ariefrahmansyah: Otherwise, if a model already deployed to non-default env and got redeployment without setting env_name will be redeployed to the default env. For example:

```
# First deployment will deploy the model version to non-default-env
merlin.deploy(v, environment_name="non-default-env")

# Redeployment without environment_name will target default-env
merlin.deploy(v)
```

python/sdk/test/integration_test.py
ariefrahmansyah: Oh, I see. I have forgotten this. Thanks for checking it.

-> I suppose this is a design choice for the update endpoint workflow and not a bug?
Yes, this is a design choice.

-> In order to remove the transformer, env vars or the logger, I think the user has to create an entirely new model version right?
It's not necessarily to create an entirely new model version. You can still using redeploy but with empty or disabled value. For example:

```
merlin.deploy(v, transformer=StandardTransformer(enabled=False))
```

python/sdk/merlin/model.py
deadlycoconuts: Sounds good, I'll make `target_env_name` be set to the current endpoint's `env_name` if it's None. Thanks for the really clear explanation! :D

api/service/version_endpoint_service.go
deadlycoconuts: Got it, thanks for digging that up 😅 I've just added that additional step to set the default protocol too!

python/sdk/merlin/model.py
deadlycoconuts: Okay sorry to bring this up, but while testing the changes I noticed a bug, and I wanted to find out if there was a reason why the `model.endpoint` attribute needs to always be from a default env? Currently the [implementation](https://github.com/caraml-dev/merlin/blob/main/python/sdk/merlin/model.py#L734) of the `endpoint` getter looks up all the endpoints of a model version to retrieve the first endpoint that's in the default env. I see that this attribute is used for setting the traffic of the model but I'm not sure why it can't be from a non-default env.

The implication of this is that in order to allow a model already deployed to a non-default env to be redeployed to the same env when the `environment_name` is not specified, I'll be unable to check the endpoint returned by the current `endpoint` getter (since the endpoint of the model deployed to the non-default env wouldn't be returned). Actually this would probably affect the way the entire `deploy` function works when redeploying a model 🤔

ariefrahmansyah: -> I wanted to find out if there was a reason why the model.endpoint attribute needs to always be from a default env?

Good question. I honestly forget why we did it this way in the first place, but inferring from how it is implemented now, I think because ModelVersion's endpoint() method is being called without the environment name parameter, we expect to return the default endpoint. (Again, why the environment name parameter is missing, I forget 🙏🏻)

As you described, the implication of using the ModelVersion.endpoint() method is leading to non-expected behavior. I think we should create new method here to get the correct previous version endpoint that to be redeployed.

deadlycoconuts: Okay, I've created a new helper method `_get_deployed_endpoint` to return an existing running (it will select the _first_ by default)/serving endpoint. This endpoint is the one that gets updated by calling the `deploy` method. This should allow users to redeploy their endpoints in a non-default environment in the same environment.

However, that still doesn't properly address the situation (even if we don't face this now) whereby a model version is already deployed to 2 different environments (1 endpoint in each environment) and if the user doesn't specify any environment name as an argument to the `deploy` method - which endpoint should the SDK update? That said I think this scenario isn't too pressing to address right now, we might perhaps want to align ourselves on how to determine the SDK's behaviour in such a scenario. I'll add a card somewhere so we can perhaps pick this up again in the future.

python/sdk/test/conftest.py
```diff
@@ -136,7 +136,7 @@ def requests():
     retry_strategy = Retry(
         total=5,
         status_forcelist=[429, 500, 502, 503, 504, 404],
-        method_whitelist=["POST"],
+        allowed_methods=["POST"],
```
deadlycoconuts, line 139: The change in this argument is done because the keyword argument `method_whitelist` has been [deprecated](https://github.com/GeneralMills/pytrends/issues/591) as of `urllib3>=2`. `allowed_methods` was introduced as a [replacement](https://github.com/urllib3/urllib3/issues/2092) from `urllib3>=1.26` onwards (the arguments for `method_whitelist` get passed to `allowed_methods` under the hood anyway).

python/sdk/setup.py
```diff
@@ -36,7 +36,7 @@
     "python_dateutil>=2.5.3",
     "PyYAML>=5.4",
     "six>=1.10",
-    "urllib3>=1.23",
+    "urllib3>=1.26",
```
deadlycoconuts, line 39: Bumped up min version of `urllib3`. Context: https://github.com/caraml-dev/merlin/pull/466/files#r1356056883

ariefrahmansyah: Thanks Zi Yi!

python/sdk/test/conftest.py
ariefrahmansyah: Thanks for fixing this!

ariefrahmansyah (approved):
LGTM! Thanks for improving the codebase as well, Zi Yi!

deadlycoconuts:
Thanks a lot for the detailed review and the suggestions @ariefrahmansyah and @tkpd-hafizhan ! Merging this now! 🚀

python/sdk/merlin/model.py
deadlycoconuts: In the end, we've decided to replace `_get_deployed_endpoint` with a similar helper method `_get_endpoint_in_environment` that instead returns an endpoint in the environment specified. This environment will be the one specified as an argument in the `deploy` method, otherwise (or if it's none) it will be the default environment. This implementation will make the `deploy` method work in the same way as the `serve_traffic` method and addresses all the concerns raised above.

PR #471, opened 2023-10-10
Introducing model version revision

ariefrahmansyah:

The current implementation of model version redeployment patches the existing inference service however, if the patch fails, it will delete the existing inference service ([ref](https://github.com/caraml-dev/merlin/blob/v0.33.0/api/cluster/controller.go#L243-L250)).

This PR reworks how we do the redeployment by introducing model version revision which itself is a new inference service that deployed for every redeployment. Then we will have a new Istio VirtualService that routes the traffic to the latest revision (latest inference service). Note that the new VirtualService adds a `Content-Type` header to the request because we need it for graceful migration to Kserve 0.11 where it requires the `Content-Type` header to be set, but our clients might not using this header.

It also changes the naming convention for the deployed inference service:

From: `{model_name}-{model_version}` to `{model_name}-{model_version}-{model_version_revision}`

### Serverless Deployment Before

image

### Serverless Deployment After

image

### Raw Deployment Before

image

### Raw Deployment After

image

### Model Version Deployment History UI

Screen Shot 2023-10-16 at 14 56 22

Changes this PR introduce:

1. Add `revision_id` column to `version_endpoints` table
2. Add revision number to model service name, prefixed with `r`
3. Add model version virtual service that routes to deployed model service with revision number to maintain backward compatibility
   a. New model service with revision number's URL is something like: my-model-name-1-r1.domain.com
   b. To maintain backward compatibility, the new model version virtual service have this URL: my-model-name-1.domain.com
4. Add Deployment History tab in the UI
5. Add Deployments API
6. Add redeploy label to deploymentCounter Prometheus metrics
7. Update e2e-test to use environment's default_deployment_config; Increase github e2e-test cluster environment's default_deployment_config

Fixes model version redeployment got deleted if the redeployment fails.

```release-note
Add model version deployment history UI
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/cluster/controller.go
```diff
@@ -250,6 +252,33 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 	}
 
 	inferenceURL := models.GetInferenceURL(s.Status.URL, isvcName, modelService.Protocol)
+
+	// Deploy virtual service
```
krithika369, line 256: Nit: `// Create / update virtual service` may be clearer since we are handling both here.

api/cluster/virtual_service.go
```diff
@@ -0,0 +1,188 @@
...
+	v1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/caraml-dev/merlin/models"
+	"github.com/caraml-dev/merlin/pkg/protocol"
+)
+
+const (
+	knativeIngressGateway          = "knative-serving/knative-ingress-gateway"
```
krithika369, line 20: Seems risky to hardcode these (what if knative / istio were installed in a different namespace?). But I realise we are doing the same thing in [Turing](https://github.com/caraml-dev/turing/blob/main/api/turing/cluster/servicebuilder/router.go#L82) and something similar is done in the kserve codebase too. Even if we don't deal with it now, I guess it would be better to add a TODO comment to make these configurable.

api/cluster/virtual_service.go
```diff
@@ -0,0 +1,188 @@
...
+	}, nil
+}
+
+func (cfg VirtualService) BuildVirtualServiceSpec() (*v1beta1.VirtualService, error) {
+	modelVersionHost, err := cfg.getModelVersionHost()
+	if err != nil {
+		return nil, err
+	}
+
+	// We manually create model version revision to avoid Kserve's modification to transformer url if the name contains "transformer"
```
krithika369, line 65: I didn't understand this comment.. Which is the line that does "We manually create model version revision" ?

api/cluster/virtual_service.go
```diff
@@ -0,0 +1,188 @@
...
+	switch cfg.Protocol {
+	case protocol.UpiV1:
+		return []*istiov1beta1.HTTPRoute{
+			{
+				Route: routeDestinations,
+			},
+		}
+
+	default:
+		// TODO: Remove setting Content-Type header after all clients are updated to set the header
```
krithika369, line 144: IMO, we can keep it as is - we are going to create this new VS anyway, it can be the fallback handling if no content-type is set for HTTP protocol.

krithika369:
Thanks for the PR, @ariefrahmansyah! Left some comments. Will have another look once you've reviewed them.

api/cluster/virtual_service.go
krithika369: Btw, can we only set the "Content-Type" header optionally, for those requests that are not setting this header? Reason is, if someone is sending binary data or so for some of their models, we shouldn't change the value to `"application/json"`. In general, we can ask the users to set the `Content-Type` header for all requests to GRPC models and especially to HTTP models that are not using `"application/json"`.

api/models/service.go
```diff
@@ -99,8 +101,8 @@ func MergeProjectVersionLabels(projectLabels mlp.Labels, versionLabels KV) mlp.L
 	return projectLabels
 }
 
-func CreateInferenceServiceName(modelName string, versionID string) string {
-	return fmt.Sprintf("%s-%s", modelName, versionID)
+func CreateInferenceServiceName(modelName, versionID, revisionID string) string {
+	return fmt.Sprintf("%s-%s-%s", modelName, versionID, revisionID)
```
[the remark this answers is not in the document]
ariefrahmansyah: Updated. Thanks for catching this.

api/service/version_endpoint_service.go
```diff
@@ -123,6 +123,9 @@ func (k *endpointService) DeployEndpoint(ctx context.Context, environment *model
 		return nil, err
 	}
 
+	// Increment endpoint revision id
+	endpoint.RevisionID++
```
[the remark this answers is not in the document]
ariefrahmansyah: Not necessarily. I expected the revision id already be incremented when we saved it to DB because the revision that failed is actually the incremented one.

api/cluster/controller.go
```diff
@@ -190,51 +201,42 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
...
-		// Get current scale of the existing deployment
-		deploymentScale := resource.DeploymentScale{}
+	// Get current scale of the existing deployment
+	deploymentScale := resource.DeploymentScale{}
+	if modelService.RevisionID > 1 {
 		if modelService.DeploymentMode == deployment.ServerlessDeploymentMode ||
 			modelService.DeploymentMode == deployment.EmptyDeploymentMode {
-			deploymentScale = c.GetCurrentDeploymentScale(ctx, modelService.Namespace, s.Status.Components)
-		}
+			prevRevisionID := modelService.RevisionID - 1
```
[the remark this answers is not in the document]
ariefrahmansyah: This is great suggestion. Thanks Krithika. I've passed endpoint.InferenceServiceName to modelService.CurrentIsvcName. As you hinted, this make the change backward compatible: the implementation now can correctly get the current deployment scale of model service and delete it safely. For example:

```
# A model service (name: my-model, version: 23) has been deployed before we update Merlin with this revision feature:

my-model-23

# Then, we update Merlin and redeploy model version above, we will create new model version revision and delete the current one:
my-model-23-1     # new revision
my-model-23       # terminating
```

api/cluster/virtual_service.go
ariefrahmansyah: Thanks again for reviewing the edge cases. As suggested, I created 3 route destinations:

For request targeting `/v1/models/{model_name}-{model_version}:predict` URI:
1. If content-type is empty, we will set content-type to application/json
2. If content-type is already set, we do nothing

3. And for other requests, we simply forward it to model version revision.

api/service/mocks/endpoints_service.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.22.1. DO NOT EDIT.
+// Code generated by mockery v2.20.0. DO NOT EDIT.
```
deadlycoconuts, line 1: Not a big issue but is the downgrade intentional? 😅

api/cluster/resource/templater.go
```diff
@@ -118,7 +103,7 @@ func NewInferenceServiceTemplater(standardTransformerConfig config.StandardTrans
 	return &InferenceServiceTemplater{standardTransformerConfig: standardTransformerConfig}
 }
 
-func (t *InferenceServiceTemplater) CreateInferenceServiceSpec(modelService *models.Service, config *config.DeploymentConfig) (*kservev1beta1.InferenceService, error) {
+func (t *InferenceServiceTemplater) CreateInferenceServiceSpec(modelService *models.Service, config *config.DeploymentConfig, currentReplicas DeploymentScale) (*kservev1beta1.InferenceService, error) {
```
deadlycoconuts, line 106: It seems like this `currentReplicas` argument is unused 🤔

api/cluster/virtual_service.go
```diff
@@ -0,0 +1,211 @@
...
+				Match: []*istiov1beta1.HTTPMatchRequest{
+					{
+						Uri: uri,
+					},
+				},
+				Route:   routeDestinations,
+				Rewrite: rewrite,
+			},
+			{
+				Route: routeDestinations,
```
krithika369, line 170: What is this 3rd case supposed to handle (especially in comparison with the 2nd)? Can you share an example?

api/models/service.go
```diff
@@ -70,6 +74,7 @@ func NewService(model *Model, version *Version, modelOpt *ModelOption, endpoint
 		DeploymentMode:    endpoint.DeploymentMode,
 		AutoscalingPolicy: endpoint.AutoscalingPolicy,
 		Protocol:          endpoint.Protocol,
+		CurrentIsvcName:   endpoint.InferenceServiceName,
```
krithika369, line 77: I think how we set `CurrentIsvcName` should be different - it is meant to be the currently deployed endpoint's isvc name, and not the name for the new revision, right? If there is no current endpoint, the `CurrentIsvcName` should be empty (`""`). I don't see this happening in the code now.

krithika369 (commented):
Thanks, @ariefrahmansyah. Some parts of the deployment flow are unclear to me. Left some comments / questions.

deadlycoconuts (commented):
Thanks a lot for this PR @ariefrahmansyah! I initially was afraid of looking at this PR because of the ~2.5k line deletions, until I read what you did and most of it was just the removal of code that's no longer relevant 🥲 I think most of the changes look good, except that I didn't have too much time to look very carefully/think deeply about the changes in `api/cluster/virtual_service.go`. I'll look at it later again but I think most of your testing should've covered most of edge cases and bugs (if any) but I just left a couple of questions for the moment.

api/cluster/controller.go
```diff
@@ -190,51 +201,39 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
...
-		if err != nil {
-			log.Errorf("unable to create inference service %s %v", isvcName, err)
-			return nil, ErrUnableToCreateInferenceService
-		}
-	} else {
-		// Get current scale of the existing deployment
-		deploymentScale := resource.DeploymentScale{}
+	// Get current scale of the existing deployment
+	deploymentScale := resource.DeploymentScale{}
+	if modelService.RevisionID > 1 {
```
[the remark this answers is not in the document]
ariefrahmansyah: `modelService.CurrentIsvcName` should not be empty, because we get the value from `versionEndpoint.InferenceServiceName`

api/cluster/virtual_service.go
ariefrahmansyah: Since standard models are using Kserve V1 inference protocol, this additional routing is for `/v1/models` path (reference: https://kserve.github.io/website/0.11/modelserving/data_plane/v1_protocol/). However, we actually not using API other than Predict API. I'm adding this 3rd case for fallback. But now I'm having second thought, I think we don't really need this one.

api/cluster/controller.go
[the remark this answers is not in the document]
ariefrahmansyah: Ah, I see what you mean. This comment is related to your other comments ([this](https://github.com/caraml-dev/merlin/pull/471#discussion_r1358062699)). For new endpoint, `versionEndpoint.InferenceServiceName` should be empty, hence `modelService.CurrentIsvcName` also must be empty. Then, if the deployment succeed, we update versionEndpoint.InferenceServiceName which will be saved to DB. Next, if this version endpoint got redeployed, the `versionEndpoint.InferenceServiceName` will already contains the current name (that has revision id suffix). Then we can continue as you suggested. Is my understanding correct?

api/cluster/virtual_service.go
krithika369: I see.. No strong opinion against it. If we are keeping it though, let's add a comment. (Actually, thinking about it, it will be better to add an inline comment for each of the cases, to explain what scenario it handles.)

api/cluster/virtual_service.go
krithika369: The discussion is continuing [here](https://github.com/caraml-dev/merlin/pull/471#discussion_r1358057199), resolving this thread.

api/cluster/controller.go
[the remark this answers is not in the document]
krithika369: Ok, that is clearer to me now. Although, on this part:

> if this version endpoint got redeployed, the versionEndpoint.InferenceServiceName will already contains the current name

I think it may need to change depending on this discussion: https://github.com/caraml-dev/merlin/pull/471#discussion_r1357985956

api/service/version_endpoint_service.go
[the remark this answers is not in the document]
deadlycoconuts: > endpoint's revisionID and any other configs from the newEndpoint should only be updated and saved to the DB if tobeDeployedEndpoint has successfully deployed.

Just to confirm my understanding, do you mean that this logic should be moved to `Deploy` in `api/queue/work/model_service_deployment.go` instead?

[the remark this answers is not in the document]
krithika369: Potentially, yes.

api/cluster/controller.go
```diff
@@ -250,6 +249,30 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
...
+		return nil, ErrUnableToCreateVirtualService
+	}
+
+	vs, err := c.deployVirtualService(ctx, vsCfg)
+	if err != nil {
+		log.Errorf("unable to create virtual service: %v", err)
+		return nil, ErrUnableToCreateVirtualService
+	}
+
+	if len(vs.Spec.Hosts) > 0 {
```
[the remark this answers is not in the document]
ariefrahmansyah: It should not. I'm adding this validation to avoid panic.

api/cluster/resource/templater.go
ariefrahmansyah: You are right, Zi Yi. Thanks for spotting this. Updated.

api/models/service.go
ariefrahmansyah: -> t is meant to be the currently deployed endpoint's isvc name, and not the name for the new revision, right?

Correct.

-> If there is no current endpoint, the CurrentIsvcName should be empty (""). I don't see this happening in the code now.

Yes, as suggested, `Service.CurrentIsvcName` will be empty for the first deployment. Once deployment is successful, we set the `Service.Name` to `VersionEndpoint.InferenceServiceName` and save it to DB. So for the next redeployment the `Service.CurrentIsvcName` will use the value from DB.

api/cluster/virtual_service.go
ariefrahmansyah: I've removed this comment.

Previously, I faced an issue with model deployment for a model with the `transformer` word on its name. The inference service URL will replace the `transformer` name with an empty string.

For example, given a model version revision: `std-transformer-s-157-1`, the resulting URL will be `std-s-157-1-transformer.domain.com`

This is because on Kserve 0.11, they changed how to getServiceHost: https://github.com/kserve/kserve/commit/dbfcc97ea20ba990ed454d052dde9c9a7bc26eee#diff-0299d5ad4bb060d2faff81d5d3f2aedca8a9549c50254b1f0430d170822decefR82

With this PR, we prevent this issue via the new revision virtual service that routing to the resulted isvc URL. (E.g. the new revision virtual service host is `std-transformer-s-157-1.domain.com` which route to `std-s-157-1-transformer.domain.com`

api/cluster/controller.go
```diff
@@ -280,6 +305,12 @@ func (c *controller) Delete(ctx context.Context, modelService *models.Service) (
 		}
 	}
 
+	vsName := fmt.Sprintf("%s-%s-%s", modelService.ModelName, modelService.ModelVersion, models.VirtualServiceComponentType)
```
krithika369, line 308: I believe we should only attempt to delete this for the new model services with RevisionID > 1. For the older ones, deleting the isvc will handle this VS deletion and this line will return an error, right?

api/cluster/virtual_service.go
```diff
@@ -0,0 +1,228 @@
...
+	switch cfg.Protocol {
+	case protocol.UpiV1:
+		return []*istiov1beta1.HTTPRoute{
+			{
+				Route: routeDestinations,
+			},
+		}
+
+	default:
+		// Default to application/json if Content-Type header is empty
```
krithika369, line 130: Nit: I think having this comment here is a bit misleading - makes it look like we are adding json content-type header for all non-UPI protocol. But in fact, this comment is only relevant to the first route being returned. Shall we remove the comment here since there is a similar one below anyway ?

api/queue/work/model_service_deployment.go
```diff
@@ -23,7 +24,7 @@ var deploymentCounter = prometheus.NewCounterVec(
 		Namespace: "merlin_api",
 		Help:      "Number of deployment",
 	},
-	[]string{"project", "model", "status"},
+	[]string{"project", "model", "status", "redeploy"},
```
krithika369, line 27: Would be good to mention changes like this (also more details about the new deployment history view) in the PR description, for future reference.

api/queue/work/model_service_deployment.go
```diff
@@ -71,32 +73,50 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
...
 	}
 
-	model.Project = project
-	log.Infof("creating deployment for model %s version %s with endpoint id: %s", model.Name, endpoint.VersionID, endpoint.ID)
-
-	// copy endpoint to avoid race condition
+	endpoint.RevisionID++
 	endpoint.Status = models.EndpointFailed
+
+	// if inference service name is not empty, it means we are redeploying the endpoint and the endpoint is already running/serving
```
krithika369, line 88: Is this condition also true for endpoints that are currently running (the ones without the RevisionID info in the DB)? And what about terminated / failed endpoints? Agree that they should not be considered a “redeployment”. But going by the check here, will the InferenceServiceName in fact be empty for the failed / terminated endpoints ?

api/service/deployment_service.go
```diff
@@ -0,0 +1,24 @@
...
+		storage: storage,
+	}
+}
+
+type deploymentService struct {
+	storage storage.DeploymentStorage
+}
+
+func (service *deploymentService) ListDeployments(modelID, versionID, endpointUUID string) ([]*models.Deployment, error) {
+	return service.storage.ListInModelVersion(modelID, versionID, endpointUUID)
```
krithika369, line 23: Would be nice to paginate this API. Can we add a //ToDo?

ui/src/pages/version/HistoryDetails.js
```diff
@@ -0,0 +1,182 @@
...
+          )
+        );
+      },
+    },
+  ];
+
+  return (
+    <ConfigSection title="Deployment History">
+      <ConfigSectionPanel>
+        <EuiInMemoryTable
```
krithika369, line 147: Shall we enable pagination on the table? It can help if there have been too many redeployments.

krithika369 (commented):
Thanks for the changes, @ariefrahmansyah. I left some comments and questions. Please review them. If we’ve tested this PR sufficiently, I’m good with the changes.

Though, I haven’t looked at the UI changes closely. @deadlycoconuts, would you be able to review those ? Thank you!

api/queue/work/model_service_deployment.go
ariefrahmansyah: -> Is this condition also true for endpoints that are currently running (the ones without the RevisionID info in the DB)?

Yes, correct, because the existing endpoints without `revision_id` have `inference_service_name` in the DB.

-> And what about terminated / failed endpoints? Agree that they should not be considered a “redeployment”. But going by the check here, will the InferenceServiceName in fact be empty for the failed / terminated endpoints ?

Existing endpoints (the ones without the `revision_id` in DB) with `terminated` / `failed` status also have `inference_service_name` in the DB. So for the existing endpoint, it will always be marked as redeploy. But for the new endpoint, it will not have `inference_service_name` in DB if the deployment fails, because now we only set inference_service_name if there's at least one successful deployment.

ui/src/pages/version/HistoryDetails.js
ariefrahmansyah: Agree. Already added `pagination={true}` prop below.

api/cluster/controller.go
ariefrahmansyah: Updated as suggested. Thanks, Krithika!

api/cluster/controller.go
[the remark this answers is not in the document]
deadlycoconuts: Got it, thanks for the clarification!

python/sdk/test/integration_test.py
```diff
@@ -98,10 +98,7 @@ def test_sklearn(
     with merlin.new_model_version() as v:
         merlin.log_model(model_dir=model_dir)
 
-    resource_request = ResourceRequest(1, 1, "100m", "200Mi")
-    endpoint = merlin.deploy(
-        v, deployment_mode=deployment_mode, resource_request=resource_request
-    )
+    endpoint = merlin.deploy(v, deployment_mode=deployment_mode)
```
deadlycoconuts, line 101: Thanks for simplifying this! 🙏

scripts/e2e/run-e2e.sh
```diff
@@ -33,4 +33,4 @@ kubectl create namespace ${E2E_PROJECT_NAME} --dry-run=client -o yaml | kubectl
 cd ../../python/sdk
 pip install pipenv==2023.7.23
 pipenv install --dev --skip-lock --python ${PYTHON_VERSION}
-pipenv run pytest -n=8 -W=ignore --cov=merlin -m "not (gpu or feast or batch or pyfunc or local_server_test or cli or customtransformer)" --durations=0
+pipenv run pytest -n=4 -W=ignore --cov=merlin -m "not (gpu or feast or batch or pyfunc or local_server_test or cli or customtransformer)" --durations=0
```
deadlycoconuts, line 36: Why do we need to change the number of e2e test workers? :O
/

python/sdk/test/integration_test.py
ariefrahmansyah: Yes, correct. deploy() function will use the environment's default_deployment_config: https://github.com/caraml-dev/merlin/pull/471/files#diff-998d1ba4cc54befe776f4bfbc6a88d3fd43a8ea4f22a42754511592947cca06aR83-R84

scripts/e2e/run-e2e.sh
ariefrahmansyah: Related to your comment [here](https://github.com/caraml-dev/merlin/pull/471#discussion_r1363564013), where I increased default_deployment_config so reducing the number of test worker will prevent from exceeding the test cluster resource.

tiopramayudi:
just saw this PR, and it reminds me of inferencelogger because the logger get model name and model version from inferenceservice name, can you check whether it is already handled?

ui/src/pages/version/HistoryDetails.js
```diff
@@ -0,0 +1,182 @@
...
+            Error message
+          </EuiText>
+          <EuiCodeBlock isCopyable>{deployment.error}</EuiCodeBlock>
+        </>
+      );
+    }
+    setItemIdToExpandedRowMap(itemIdToExpandedRowMapValues);
+  };
+
+  const cellProps = (item, column) => {
```
deadlycoconuts, line 74: Haha this is really nice and helpful; maybe we can bring this into some utils package in the future and reuse it in other places 🙏

ui/src/pages/version/HistoryDetails.js
```diff
@@ -0,0 +1,182 @@
...
+      field: "updated_at",
+      name: "Deployment Time",
+      render: (date, deployment) => (
+        <>
+          <DateFromNow date={date} size={defaultTextSize} />
+          &nbsp;&nbsp;
+          {deployment.id === deployedRevision.id && (
+            <EuiBadge color="default">Current</EuiBadge>
+          )}
+          {/* {JSON.stringify(deployment.id)} */}
```
deadlycoconuts, line 95: Is this meant to be removed? 😅

ui/src/version/VersionListTable.js
```diff
@@ -426,7 +426,7 @@ const VersionListTable = ({
                       <EuiLink
                         onClick={() =>
                           navigate(
-                            `/merlin/projects/${activeModel.project_id}/models/${activeModel.id}/versions/${endpoint.version_id}/endpoints/${endpoint.id}`
+                            `/merlin/projects/${activeModel.project_id}/models/${activeModel.id}/versions/${endpoint.version_id}/endpoints/${endpoint.id}/details`
```
deadlycoconuts, line 429: Why does this need to change? :o

scripts/e2e/run-e2e.sh
deadlycoconuts: Ah oops, got it, thanks for the explanation!

deadlycoconuts (approved):
LGTM! 🚀 There are some pending open discussions but I think most of them are pretty minor, except for say this one here https://github.com/caraml-dev/merlin/pull/471#discussion_r1353618373 which might affect the implementation by quite a bit, and Tio's latest comment. What's your take on it? The rest of the open threads are either already resolved or aren't too crucial to the behaviour of the changes.

I took a look at the UI changes - they look good! 🙏 Thanks for adding those changes into this PR.

ui/src/version/VersionListTable.js
ariefrahmansyah: Because without `/details` the model version ui page does not automatically set the Configuration table as active even though we already set index props on /details Route ([ref](https://github.com/caraml-dev/merlin/blob/v0.33.0/ui/src/pages/version/VersionDetails.js#L201)).
Screen Shot 2023-10-18 at 17 24 46

ui/src/pages/version/HistoryDetails.js
ariefrahmansyah: Oops. I'll remove it.

ariefrahmansyah:
-> just saw this PR, and it reminds me of inferencelogger because the logger get model name and model version from inferenceservice name, can you check whether it is already handled?

Thanks @tiopramayudi for looking at this! You are correct that it was not properly handled. The logger will create new kafka topic with model version on it.

For example, given in `my-project` project, we've deployed a `my-model` model, version `2`. The pod created would be something like this: my-model-2-r1 and the Kafka topic created would be `merlin-my-project-my-model-2-inference-log`. However, the expected topic name would be `merlin-my-project-my-model-inference-log` without version number.

I have fixed it in this commit: https://github.com/caraml-dev/merlin/pull/471/commits/37856c51b0ca2223982cef2d6d297dc71654c885
Please help take a look again!

ui/src/version/VersionListTable.js
deadlycoconuts: Ahh ok got it :) Thanks!

api/service/version_endpoint_service.go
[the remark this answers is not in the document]
ariefrahmansyah: For Merlin, we still need to save the endpoint to DB before doing the deployment because of our queue-worker implementation. In ModelServiceDeployment.Deploy(), Merlin calls the DB to get the latest endpoint data to be deployed in case other workers (other Merlin replicas) already picked up the deployment process.

I understand your concern, and I have updated the ModelServiceDeployment.Deploy() implementation to not update endpoint status from running/serving to fail if the new redeployment fails.

PR #474, opened 2023-10-19
fix(api): Add additional check to allow update endpoint requests without deployment type specified

deadlycoconuts:
As a follow up to #455 and #466, this tiny change to the update endpoint _endpoint_ allows requests that do not specify a deployment type to avoid the check that checks for different deployment types between a deployed endpoint and the new endpoint configuration. What will instead happen if a deployment type isn't specified is that the existing deployed endpoint's deployment type will be used.

The main change is just the addition of another boolean condition, but there's a unit test that I added which is just an almost exact copy of another unit test case.

Fixes #

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


ariefrahmansyah (approved):
LGTM! Thanks Zi Yi for quick fix!

PR #475, opened 2023-10-20
Fix current deployment history status

ariefrahmansyah:
…, the current deployment history shall be in Deployed status

During the redeployment, the current revision deployment status should keep in Deployed status instead of Not Deployed.

### During redeployment

Screen Shot 2023-10-20 at 14 37 34

### After redeployment completed 

Screen Shot 2023-10-20 at 14 38 46

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/queue/work/model_service_deployment.go
+++ b/api/queue/work/model_service_deployment.go
@@ -107,11 +107,9 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 	}
 
 	defer func() {
-		deploymentCounter.WithLabelValues(model.Project.Name, model.Name, fmt.Sprint(endpoint.Status), fmt.Sprint(isRedeployment)).Inc()
+		deploymentCounter.WithLabelValues(model.Project.Name, model.Name, fmt.Sprint(deployment.Status), fmt.Sprint(isRedeployment)).Inc()
 
 		// record the deployment result
-		deployment.Status = endpoint.Status
-		deployment.Error = endpoint.Message
 		deployment.UpdatedAt = time.Now()
 		if _, err := depl.DeploymentStorage.Save(deployment); err != nil {
 			log.Warnf("unable to update deployment history", err)
@@ -125,6 +123,8 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 
 	modelOpt, err := depl.generateModelOptions(ctx, model, version)
 	if err != nil {
+		deployment.Status = models.EndpointFailed
+		deployment.Error = err.Error()
 		endpoint.Message = err.Error()
 		return err
 	}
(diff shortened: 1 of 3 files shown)


deadlycoconuts (approved):
Thanks a lot for this fix @ariefrahmansyah! 🚀 Sorry I didn't catch this earlier from your previous PR ><"

api/queue/work/model_service_deployment.go
```diff
@@ -107,11 +107,9 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 	}
 
 	defer func() {
-		deploymentCounter.WithLabelValues(model.Project.Name, model.Name, fmt.Sprint(endpoint.Status), fmt.Sprint(isRedeployment)).Inc()
+		deploymentCounter.WithLabelValues(model.Project.Name, model.Name, fmt.Sprint(deployment.Status), fmt.Sprint(isRedeployment)).Inc()
 
 		// record the deployment result
-		deployment.Status = endpoint.Status
-		deployment.Error = endpoint.Message
```
deadlycoconuts, line 114: If I understood correctly, this object `deployment` corresponds to the 'new' deployment that we're attempting to  deploy right? If so, how does the change above prevent a running 'old' deployment's `Status` from being changed to anything that is not pending/serving/running when a 'new' re-deployment attempt fails?

ariefrahmansyah: -> If I understood correctly, this object deployment corresponds to the 'new' deployment that we're attempting to deploy right?

Correct.

-> If so, how does the change above prevent a running 'old' deployment's Status from being changed to anything that is not pending/serving/running when a 'new' re-deployment attempt fails?

If you refer to the old deployments as the previous rows in the deployment table, we will not modify them. Instead, if you are referring to the running old version_endpoints, we will not modify the status if it's already running or serving (ref: https://github.com/caraml-dev/merlin/blob/main/api/queue/work/model_service_deployment.go#L90-L93)

deadlycoconuts: Got it, thanks for the clarification!

> If you refer to the old deployments as the previous rows in the deployment table, we will not modify them.

I'm actually referring to this. Your explanation on the running old version_endpoints makes sense. It answers why the status in (1) remains as 'running' (if I'm understanding things correctly). What about the status of the old deployment marked in (2)? I still don't get how that is fixed because it should be 'running' right? Just like in (1). 😣 Sorry if I misunderstood something in the current workflow.

Screenshot 2023-10-23 at 12 33 10 PM

deadlycoconuts:
Ouch our CICD pipeline keeps breaking for a lot of reasons outside our control: https://github.com/caraml-dev/merlin/actions/runs/6608473758/job/17947224988#step:7:17 🥲

api/queue/work/model_service_deployment.go
ariefrahmansyah: Can you elaborate on how to reproduce your issue? From my end, the old deployment status remains Deployed when the newer deployment failed:

Screen Shot 2023-10-23 at 11 43 18

ariefrahmansyah:
> Ouch our CICD pipeline has been breaking a lot recently for reasons outside our control: https://github.com/caraml-dev/merlin/actions/runs/6608473758/job/17947224988#step:7:17 🥲

Yeah, it's most likely due to uvloop v0.19.0 stopping Python 3.7 support: https://github.com/MagicStack/uvloop/releases/tag/v0.19.0

api/queue/work/model_service_deployment.go
deadlycoconuts: Okay, I think misunderstood how the implementation worked so I couldn't understand how the bug came about. In any case, I get how the whole thing works now and I see why your fix manages to address that issue. Thanks a lot for following up on this thread Arief! :D Everything's clear now!

deadlycoconuts (approved):
Thanks a lot for these fixes @ariefrahmansyah! One of the pyfunc server tests is unfortunately still failing but feel free to merge this whenever you manage to get that to pass (thanks for looking into that issue too!) 🚀🚀🚀

PR #479, opened 2023-10-25
Model Version Revision Improvement

ariefrahmansyah:

This PR is for fixing and improving model version revision:

1. For the redeployment, we are no longer update the current endpoint with new endpoint data to make sure the UI still showing the current endpoint deployed which is consistent with the actual endpoint deployed in the cluster. Instead, once the deployment is successful, we will then update the endpoint.
2. Fix Deployment History UI when deployment status is running but with error message displayed as Failed.

* Fixes deployment status UI is failed even though status is running or serving.

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


ui/src/version/VersionListTable.js
```diff
@@ -192,7 +192,7 @@ const VersionListTable = ({
                           <EuiLink
                             onClick={() =>
                               navigate(
-                                `/merlin/projects/${activeModel.project_id}/models/${activeModel.id}/versions/${version.id}/endpoints/${versionEndpoint.id}`
+                                `/merlin/projects/${activeModel.project_id}/models/${activeModel.id}/versions/${version.id}/endpoints/${versionEndpoint.id}/details`
```
leonlnj, line 195: this is to fix the config tab hiding when clicked on the deployment environment link?

ui/src/pages/version/HistoryDetails.js
```diff
@@ -39,6 +33,10 @@ const DeploymentStatus = ({
   } else if (status === "pending") {
     return <EuiHealth color="gray">Pending</EuiHealth>;
   }
+
+  if (deployment.error !== "") {
```
leonlnj, line 37: i think its more consistent to join the above if else block by checking `status === failed'

api/queue/work/model_service_deployment.go
```diff
@@ -58,17 +56,8 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 		return err
 	}
 
-	endpointArg := jobArgs.Endpoint
-	endpoint, err := depl.Storage.Get(endpointArg.ID)
```
ariefrahmansyah, line 62: We no longer read the database to get the endpoint data to be deployed. Instead, we are using the endpoint argument passed on by queue.Job. It should be safe to read from the job since the same replicas process the job (publisher and consumer for a job is in the same replicas) and we have locking when getting unprocessed job.

ui/src/pages/version/HistoryDetails.js
ariefrahmansyah: Apparently, there's a case when status is running or serving but deployment.error is not empty, that's why I put it in different condition branch.

ui/src/version/VersionListTable.js
ariefrahmansyah: Yes, correct.

ui/src/pages/version/HistoryDetails.js
leonlnj: if the status is running or serving, wont it returned `return <EuiHealth color="default">Not Deployed</EuiHealth>;` at the above instead?

ariefrahmansyah: Yes, correct. That's the reason I moved deployment.error validation after checking the status because status takes precedence over error.

api/models/version_endpoint.go
```diff
@@ -45,9 +45,9 @@ type VersionEndpoint struct {
 	// ServiceName service name
 	ServiceName string `json:"service_name" gorm:"service_name"`
 	// InferenceServiceName name of inference service
-	InferenceServiceName string `json:"-" gorm:"inference_service_name"`
+	InferenceServiceName string `json:"inference_service_name" gorm:"inference_service_name"`
 	// Namespace namespace where the version is deployed at
-	Namespace string `json:"-" gorm:"namespace"`
+	Namespace string `json:"namespace" gorm:"namespace"`
```
leonlnj, line 50: why is the marshalling now when it previously wasn't needed

api/queue/work/model_service_deployment.go
```diff
@@ -58,17 +56,8 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
...
-		log.Errorf("could not found version endpoint with id %s and error: %v", endpointArg.ID, err)
-		return err
-	}
-	if err != nil {
-		log.Errorf("could not fetch version endpoint with id %s and error: %v", endpointArg.ID, err)
-		// If error getting record from db, return err as RetryableError to enable retry
-		return queue.RetryableError{Message: err.Error()}
-	}
+	endpoint := jobArgs.Endpoint
+	previousStatus := endpoint.Status
```
leonlnj, line 60: what is the status at the point of request?

api/models/version_endpoint.go
ariefrahmansyah: Previously we fetched version endpoint value from the database, but now we will get the version endpoint from job queue which is stored in JSON and need to be unmarshal properly.

api/queue/work/model_service_deployment.go
ariefrahmansyah: Depends on whether this is the first deployment or redeployment. If it's first deployment, it would be in Pending state. However, if it's redeployment it would use the current endpoint status.

krithika369: Should we set the status to "pending" regardless of whether it's a deployment or redeployment, while the deployment is in progress? This is the current behavior before this PR, right? (Also relates to the comment below: `// if this is a redeployment, we don't want to update the endpoint status`)

HAving the "pending" status in the UI is helpful to know that a deployment is in progress. Otherwise, the only place is the new history tab which is not very obvious. Wdyt?

ariefrahmansyah: Oh, my bad. I have updated the endpoint status to be pending regardless of deployment or redeployment: https://github.com/caraml-dev/merlin/pull/479/files/680513818f33a595931250fcb7e1ef41e2dbd684#diff-1a4360ee4934995aec6c53cb8d49a852273a8b6f44cf4cda36a451436ab05b28R120

This is slightly different from the current behaviour where we update all other columns. This PR only updates the endpoint status to pending prior to the deployment/redeployment. Once the deployment is successful, we update other columns. And yes this pending status will also be displayed in the UI.

ariefrahmansyah: So the [`tobeDeployedEndpoint`](https://github.com/caraml-dev/merlin/pull/479/files/680513818f33a595931250fcb7e1ef41e2dbd684#diff-1a4360ee4934995aec6c53cb8d49a852273a8b6f44cf4cda36a451436ab05b28R133) passed into this Deploy() function has the current endpoint status. While separately we updated endpoint status to pending in DB -- but this will not affect the `tobeDeployedEndpoint`'s status.

ui/src/pages/version/HistoryDetails.js
leonlnj: so this `if (deployment.error !== "")` will only be called when `status===failed`? or there are other status? if the status is running or serving it wont call this block

deadlycoconuts: > so this if (deployment.error !== "") will only be called when status===failed? or there are other status? if the status is running or serving it wont call this block

I think the only possible status values are these: https://github.com/caraml-dev/merlin/blob/e78908cf7a1e89fb37771a1cff25cbbfd748a04d/api/models/endpoint_status.go#L20
So it's only when an endpoint is in the `failed` or `terminated` state where the code execution enters this block. That said, should we return `<EuiHealth color="danger">Terminated</EuiHealth>` if the `deployment.error === "" && status==="terminated"`? 

> if the status is running or serving it wont call this block

I don't think we're expecting the deployment to have an error if it's not in the `failed` state if I understood correctly 🤔

ui/src/version/VersionEndpointActions.js
```diff
@@ -166,7 +171,6 @@ const VersionEndpointActions = ({
 
     // Undeploy
     if (
-      versionEndpoint.status === "pending" ||
```
deadlycoconuts, line 169: Thanks for this @ariefrahmansyah !

deadlycoconuts: I was thinking of replacing `actions.undeploy` with something like this instead but I guess this isn't too important:
```js
    undeploy: (undeployable) =>
      <EuiFlexItem grow={false} key={`undeploy-${versionEndpoint.id}`}>
        <EuiButtonEmpty
          onClick={openModal(props.toggleUndeployEndpointModal)}
          color="danger"
          iconType="exportAction"
          size="xs"
          disabled={undeployable}
        >
          <EuiText size="xs">Undeploy</EuiText>
        </EuiButtonEmpty>
      </EuiFlexItem>
    ,
```

deadlycoconuts (approved):
Thanks a lot for all the investigation into the bugs and for the fixes @ariefrahmansyah ! 🚀 The changes mostly look good to me, maybe the rest will have some other opinions so I'll leave them to share their other comments.

leonlnj (approved):
i dont have any major concern regarding the change. 

I find it hard to follow what is going on cause there seems to be side effects or assumption about the data that is going into the function and im not familiar with the codebase for the various flow that can be missed 😅

PR #480, opened 2023-10-27
Fix: Set version endpoint created and updated time on initialization

ariefrahmansyah:

After using jobArgs.Endpoint data as the [reference value](https://github.com/caraml-dev/merlin/blob/v0.34.0/api/queue/work/model_service_deployment.go#L61) for the deployment, the CreatedUpdated value is set to the time default value (`0001-01-01 00:00:00 +0000 UTC`) which could lead to DB update to use this value. This PR fixes it by getting the created_at and updated_at from version_endpoints DB so that it can be updated correctly.

Fixes wrong created_at value on version_endpoints

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/queue/work/model_service_deployment.go
+++ b/api/queue/work/model_service_deployment.go
@@ -77,6 +77,10 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 		return queue.RetryableError{Message: err.Error()}
 	}
 
+	// Use CreatedUpdated value from DB so that jobArgs.Endpoint's CreatedUpdated are not default value (0001-01-01 00:00:00 +0000 UTC)
+	endpoint.CreatedAt = prevEndpoint.CreatedAt
+	endpoint.UpdatedAt = prevEndpoint.UpdatedAt
+
 	isRedeployment := false
 
 	// Need to reassign destionationURL cause it is ignored when marshalled and unmarshalled
@@ -157,7 +161,6 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 	endpoint.URL = svc.URL
 	endpoint.ServiceName = svc.ServiceName
 	endpoint.InferenceServiceName = svc.CurrentIsvcName
-	endpoint.UpdatedAt = time.Now()
 	endpoint.Message = "" // reset message
 
 	if previousStatus == models.EndpointServing {
--- a/api/queue/work/model_service_deployment.go
+++ b/api/queue/work/model_service_deployment.go
@@ -157,6 +157,7 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 	endpoint.URL = svc.URL
 	endpoint.ServiceName = svc.ServiceName
 	endpoint.InferenceServiceName = svc.CurrentIsvcName
+	endpoint.UpdatedAt = time.Now()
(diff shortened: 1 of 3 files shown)


api/queue/work/model_service_deployment.go
```diff
@@ -77,6 +77,10 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 		return queue.RetryableError{Message: err.Error()}
 	}
 
+	// Use CreatedUpdated value from DB so that jobArgs.Endpoint's CreatedUpdated are not default value (0001-01-01 00:00:00 +0000 UTC)
+	endpoint.CreatedAt = prevEndpoint.CreatedAt
```
krithika369, line 81: Do you think the problem is that the json tag is not defined for `CreatedUpdated` [here](https://github.com/caraml-dev/merlin/blob/7fc8e96b22826e6f460cdaf32fc90976ff24f279/api/models/version_endpoint.go#L73)?

ariefrahmansyah: I don't think that's the issue. CreatedUpdated is composite field and we specified json tag here: https://github.com/caraml-dev/merlin/blob/7fc8e96b22826e6f460cdaf32fc90976ff24f279/api/models/model.go#L50-L51

And our Get Version Endpoints API returns endpoint data with created_at and updated_at fields.

The issue is `endpoint` here is using the default time value for created_at and updated_at. And when we save it, we use that default time value for created_at.

krithika369: Hmm.. You're right that the JSON tag should not be necessary for this composite field. But I'm still not clear why `0001-01-01 00:00:00 +0000 UTC` is the value of created_at / updated_at.

It sounds like a problem during marshal / unmarshal. But the default implementations should handle it, right?
* https://pkg.go.dev/time#Time.MarshalJSON
* https://pkg.go.dev/time#Time.UnmarshalJSON

Or is this bug only happening for cases where the new endpoint is created, since the endpoint record is not saved to the DB (which is supposed to set created_at/updated_at) prior to creating the job?
* https://github.com/caraml-dev/merlin/blob/main/api/service/version_endpoint_service.go#L116
* https://github.com/caraml-dev/merlin/blob/main/api/models/version_endpoint.go#L77

If this is the case, I guess the better thing to do would be to save the record to the DB first rather than correct the timestamps during Deploy.

krithika369 (approved):

api/queue/work/model_service_deployment.go
ariefrahmansyah: Yes, this bug is only happening for the new endpoint or in other words, for the first deployment of the model version.

When a model version is deployed for the first time, the VersionEndpoint returned by `models.NewVersionEndpoint` didn't have CreatedAt and UpdatedAt values. Then when we save to DB for the first time, the created_at and updated_at were set by DB.

However, when we pass that VersionEndpoint to Deploy() function, the VersionEndpoint still didn't have CreatedAt and UpdatedAt values so that when we save the VersionEndpoint after the deployment is complete, the CreatedAt value is override by the empty CreatedAt value.

Your suggestion is correct and we can set the CreatedAt and UpdatedAt value upon the initialization. I have updated the code and tested it.

PR #483, opened 2023-11-02
Refactor deployments storage on redeploy and undeployment

ariefrahmansyah:

The terminated endpoint means the user has triggered Undeployment and there's no deployed endpoint in the cluster. When we redeploy this terminated endpoint, the deployment history should show all previous successful deployment logs as Not Deployed and the new deployment as Pending.

However, when redeploying the terminated endpoint, the deployment history shows the previous successful deployment as the current and deployed one:

Before redeployment of terminated endpoint:

image

Actual view from redeployment of terminated endpoint:

image

## Fixes introduced by this PR:

Before redeployment, the current label is given to the latest successful/terminated endpoint:

image

During redeployment:

image

After redeployment succeed:

image

*Note: Please ignore the time displayed as `in 7 hours` -- this is because I'm running from local and the time zone is not translated correctly.

---

Behind the scene, we refactor deployments table for given version endpoint for redeployment and unemployment:

1. Redeployment
  a. new deployment status = running/serving,
  b. old successful deployment status = terminated
2. Undeployment
  a. old successful deployment status = terminated

* Fixes redeployment of terminated endpoint
* Fixes duplicate link icon on Mlflow run

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


api/storage/deployment_storage.go
```diff
@@ -83,3 +89,84 @@ func (d *deploymentStorage) GetFirstSuccessModelVersionPerModel() (map[models.ID
...
+		if err != nil {
+			tx.Rollback()
+		} else {
+			tx.Commit()
+		}
+	}()
+
+	var deployments []*models.Deployment
+	err = tx.Where("version_model_id = ? AND version_id = ? AND version_endpoint_id = ?",
+		newDeployment.VersionModelID, newDeployment.VersionID, newDeployment.VersionEndpointID).Find(&deployments).Error
```
krithika369, line 119: We should only update previous "running" / "serving" endpoints to "terminated", right?

Can we include the filter in this query? If not, potentially, this can also be expensive as it will end up returning all previous deployment attempts. 

Once we do the ad-hoc DB migration, at any given time, there should only be a single deployment record in the DB for a given endpoint that is in the running/serving state. All others will be failed/terminated and can be ignored.

api/storage/deployment_storage.go
```diff
@@ -27,6 +29,10 @@ type DeploymentStorage interface {
 	ListInModelVersion(modelID, versionID, endpointUUID string) ([]*models.Deployment, error)
 	// Save save the deployment to underlying storage
 	Save(deployment *models.Deployment) (*models.Deployment, error)
+	// OnDeploymentSuccess updates the new deployment status to successful on DB and update all previous deployment status for that version endpoint to terminated.
+	OnDeploymentSuccess(newDeployment *models.Deployment) error
```
krithika369, line 33: I'm not sure these 2 methods should be in the storage layer. The storage layer is typically slim, providing just the data access capabilities to the DB, right? Instead of these, can we introduce:

```golang
ListDeploymentsWithStatus(modelID string, versionID string, endpointUUID string, statuses []string)
```

This method can be called with the appropriate input params (statuses will be {"running", "serving"}) - then, their status can be updated and `Save()` called from the service layer.

Wdyt?

ui/src/pages/version/ModelVersionPanelHeader.js
```diff
@@ -46,8 +46,7 @@ export const ModelVersionPanelHeader = ({ model, version }) => {
       title: "MLflow Run",
       description: (
         <EuiLink href={version.mlflow_url} target="_blank">
-          {version.mlflow_url}&nbsp;
-          <EuiIcon className="eui-alignBaseline" type="popout" size="s" />
+          {version.mlflow_url}
```
ariefrahmansyah, line 49: Fix duplicate link icon:
image

api/storage/deployment_storage.go
ariefrahmansyah: The main reason I added it here is that I want to use a DB transaction because we are updating multiple data at once. Calling Save() multiple times in the service layer might introduce inconsistent data updates (one of Save() might fail which can lead to not updated data).

We also have a similar approach with [Model Endpoint Storage's Save method](https://github.com/caraml-dev/merlin/blob/main/api/storage/model_endpoint_storage.go#L79-L132) where it updates both model and version endpoints.

api/storage/deployment_storage.go
ariefrahmansyah: We were thinking the same thing. I realized this after doing some testing and am surprised that you were suggesting the same improvement. Thanks for the suggestion.

krithika369:
Hi @ariefrahmansyah, I will have another look at this PR today. In the meantime, I was thinking it may also be good to capture the data migration steps (the SQL script and any ad-hoc action that needs to be taken for deployments since Merlin 0.34.0) in this MR as a `.md` doc or something like that, so it's easier for everyone to review. Could you help update that too? Thanks.

ariefrahmansyah:
> Hi @ariefrahmansyah, I will have another look at this PR today. In the meantime, I was thinking it may also be good to capture the data migration steps (the SQL script and any ad-hoc action that needs to be taken for deployments since Merlin 0.34.0) in this MR as a `.md` doc or something like that, so it's easier for everyone to review. Could you help update that too? Thanks.

Hi Krithika, I'm still playing on the SQL on my local machine. Will update the MR with the SQL doc. Btw, I'll create the doc with this path: `docs/dev-guide/deployments_table_migration.md`

ui/src/pages/version/HistoryDetails.js
```diff
@@ -14,48 +14,41 @@ import { useMerlinApi } from "../../hooks/useMerlinApi";
...
-      if (deployment.status === "running" || deployment.status === "serving") {
-        deployment.revision_id = idx + 1;
-        idx++;
-      }
+    const ordered = deployments.sort((a, b) => (a.id < b.id ? 1 : -1));
+    setOrderedDeployments(ordered);
+
+    const currentDeployment = ordered.find((deployment) => {
+      return (
+        (deployment.status === "running" ||
```
krithika369, line 43: Should we just look for `deployment.status` != "failed"? Also, is there a reason we also need to check for `deployment.error === ""` besides the status?

ui/src/pages/version/HistoryDetails.js
```diff
@@ -14,48 +14,41 @@ import { useMerlinApi } from "../../hooks/useMerlinApi";
...
+          deployment.status === "terminated") &&
+        deployment.error === ""
+      );
     });
 
-    orderedDeployments = orderedDeployments.sort((a, b) =>
-      a.id < b.id ? 1 : -1
-    );
-    setOrderedDeployments(orderedDeployments);
+    console.log("currentDeployment", currentDeployment);
```
krithika369, line 50: This needs to be cleaned up before merging.

ui/src/pages/version/HistoryDetails.js
```diff
@@ -14,48 +14,41 @@ import { useMerlinApi } from "../../hooks/useMerlinApi";
...
+    const ordered = deployments.sort((a, b) => (a.id < b.id ? 1 : -1));
+    setOrderedDeployments(ordered);
+
+    const currentDeployment = ordered.find((deployment) => {
+      return (
+        (deployment.status === "running" ||
+          deployment.status === "serving" ||
+          deployment.status === "terminated") &&
+        deployment.error === ""
+      );
```
krithika369, line 47: We should also handle the case where all previous deployment attempts were failed (`currentDeployment` is empty) - in this case, the latest record is simply the current.

krithika369 (commented):
@ariefrahmansyah left some comments. As discussed earlier, do capture the data migration script in this PR as well. I think it should be safe to add it as a normal SQL migration file as we don't anticipate any ad-hoc handling for the deployments status update. Thanks!

ui/src/pages/version/HistoryDetails.js
ariefrahmansyah: `deployment.status != "failed"` will include pending deployment too, however, we don't want to mark pending deployment as Current deployment.

We check `deployment.error === ""` for backward compatibility. Prior to Merlin 0.34 (model version revision introduction), deployments table might have the row with status in running and serving but with error message due to deployments data is always copying from the latest endpoint deployment's message. Ref: https://github.com/caraml-dev/merlin/blob/v0.33.0/api/queue/work/model_service_deployment.go#L95

api/storage/deployment_storage.go
```diff
@@ -83,3 +89,91 @@ func (d *deploymentStorage) GetFirstSuccessModelVersionPerModel() (map[models.ID
...
 	return d.db.Where("version_id = ? AND version_model_id = ?", versionID, modelID).Delete(models.Deployment{}).Error
 }
+
+// OnDeploymentSuccess updates the new deployment status to successful on DB and update all previous successful deployment status for that version endpoint to terminated.
+func (d *deploymentStorage) OnDeploymentSuccess(newDeployment *models.Deployment) error {
+	if newDeployment.ID == 0 {
+		return fmt.Errorf("newDeployment.ID must not be empty")
+	}
+
+	if !newDeployment.IsSuccess() {
```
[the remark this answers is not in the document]
ariefrahmansyah: Yes, I still think it's better to have this validation inside the function implementation as well. The OnDeploymentSuccess should not make an assumption on the data it receives, instead, it should have its own validation to make sure the input is correct.

ui/src/pages/version/HistoryDetails.js
```diff
@@ -14,48 +14,41 @@ import { useMerlinApi } from "../../hooks/useMerlinApi";
...
-    }
-    return <EuiHealth color="default">Not Deployed</EuiHealth>;
-  } else if (status === "pending") {
-    return <EuiHealth color="gray">Pending</EuiHealth>;
-  }
-
-  if (deployment.error !== "") {
-    return <EuiHealth color="danger">Failed</EuiHealth>;
+    case "terminated":
+      return <EuiHealth color="default">Not Deployed</EuiHealth>;
```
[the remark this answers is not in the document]
ariefrahmansyah: Updated:

image

ui/src/pages/version/HistoryDetails.js
ariefrahmansyah: Updated:

image

api/storage/deployment_storage.go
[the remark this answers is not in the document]
ariefrahmansyah: Oh, I finally remembered a phrase for this: "trust but verify" 😁

api/storage/deployment_storage.go
ariefrahmansyah: Yes, my implementation which uses transaction will only call one save. I was referring to another approach where the service layer is the one that call db.Save() multiple times and without using transaction.

api/queue/work/model_service_deployment.go
```diff
@@ -111,8 +111,15 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
 
 		// record the deployment result
 		deployment.UpdatedAt = time.Now()
-		if _, err := depl.DeploymentStorage.Save(deployment); err != nil {
-			log.Warnf("unable to update deployment history", err)
+		if deployment.IsSuccess() {
+			if err := depl.DeploymentStorage.OnDeploymentSuccess(deployment); err != nil {
```
leonlnj, line 115: im at dilemma for these, given the current state of how it is written it seems fine. So just leaving a comment and not expecting any changes for now.

I would expect the service to use the gorm.DB directly to start transaction but its abstracted away for now or the storage to expose transactional methods.

Personally I think there are many side effects and unclear segregation of responsibilities. I would expect storage layer to own "How to save to db" and the service to own "How and what to save to db". 
Its within the storage layer logic that all old deployment must be updated. It can expose a method to return all past deployment, the service layer updates the deployment and logic, then pass a list of deployment for the storage. The storage would simply persist the list.

Imagine using another storage, it would assume the storage should decide the old deployment gets updated or not.

Now it looks like `Save` is for failure and `OnDeploymentSuccess` for success case. It is also not obvious from these.

api/storage/deployment_storage.go
```diff
@@ -83,3 +89,91 @@ func (d *deploymentStorage) GetFirstSuccessModelVersionPerModel() (map[models.ID
...
+	defer func() {
+		if err != nil {
+			tx.Rollback()
+		} else {
+			tx.Commit()
+		}
+	}()
+
+	var deployments []*models.Deployment
+	err = tx.Where("version_model_id = ? AND version_id = ? AND version_endpoint_id = ? AND status IN ('pending', 'running', 'serving')",
```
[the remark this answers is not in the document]
leonlnj: this is updating all deployment under the version right? or was this comment before update 😅

docs/dev-guide/deployments_table_migration.md
```diff
@@ -0,0 +1,85 @@
+# Deployments Table Migration
+
+This doc outline the data migration steps required for [PR #483 - Refactor deployments storage on redeploy and undeployment](https://github.com/caraml-dev/merlin/pull/483).
```
leonlnj, line 3: CMIIW, the storage service ensure that ALL deployment are updated under the endpoint is correct during a deployment

this is to update everything outside of deployment as a one time effort

ariefrahmansyah: Yes, correct. The deployment storage makes sure there will be only one running/serving deployment for a given endpoint. This migration doc is just a one-time effort to make sure the old deployment data follows the same.

api/storage/deployment_storage.go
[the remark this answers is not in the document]
krithika369: Ah perhaps my phrasing wasn't clear. I was specifically referring to "pending" deployments. Currently the code is updating pending/serving/running deployments. But "pending" is a special scenario for which I am not sure what the correct handling should be.

> The API is expected to prevent further redeployment / undeployment attempts if there are current pending deployments, right? This is currently blocked from the UI but we should also block this from the API (there is a card for this in the backlog) because otherwise, the K8s resources from the "pending" deployment are not cleaned up.

> Alternatively, we should handle the clean up of the K8s resources for the pending deployment in the undeploy/redeploy endpoint flow - in this case, it makes sense to handle the pending deployment here, as you've done.

> What is the desired behavior of endpoint update / delete when a deployment is already in progress?

ui/src/pages/version/HistoryDetails.js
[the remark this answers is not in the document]
krithika369: Wait - is this the correct screenshot for this comment? There are 2 suggestions in the comment, neither captured in the screenshot. 😅

api/storage/deployment_storage.go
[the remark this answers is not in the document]
leonlnj: I would assume `pending` is only applicable to the latest/ongoing deployment, since its not a terminal state. 

If there was past hanging deployment that got stuck in pending, then it should be updated to `failed` since it never succeeded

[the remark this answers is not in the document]
krithika369: > If there was past hanging deployment that got stuck in pending, then it should be updated to failed since it never succeeded

Makes sense. So, by the time this code is reached, we shouldn't have any "pending" deployments anymore (unless something was stuck in the past and not cleaned up properly; but this shouldn't be masked by updating the pending status to terminated in this block of code). So, TL;DR: we should drop "pending" from the query here.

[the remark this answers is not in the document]
deadlycoconuts: Late to the party but yeah I agree `pending` should be dropped from this query once the we block further redeployment/undeployment of pending deployments on the API server. 

Just one more not so important question, is it overkill to be updating _all_ deployments that have the `running`/`serving` status to terminated? Or should only the existing _current_ deployment's status be updated (if it is indeed `running` or `serving`)? Granted, this doesn't change the end result and will still be super safe but was just wondering if this blanket update might potentially hide some unexpected behaviour (say for some reason we have deployments that are somehow `running` or `serving` but are not the current deployment - would updating their status in the DB would hide some potential underlying problem?)

docs/dev-guide/deployments_table_migration.md
```diff
@@ -0,0 +1,85 @@
+# Deployments Table Migration
```
krithika369, line 1: Per [this comment](https://github.com/caraml-dev/merlin/pull/483#pullrequestreview-1713955770), can we just turn this into up / down sql migration files? This will make it easier to propagate this to all envs.

Is there any concern with capturing this data migration in the `db-migrations` scripts?

api/storage/deployment_storage.go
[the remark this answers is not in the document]
krithika369: > is it overkill to be updating all deployments that have the running/serving status to terminated?

The problem is, we are not actually tracking the "current" deployment record anywhere - it's identified to be a deployment that's in:
* running / serving state
* (OR) latest failed deployment

Only the first one needs to be updated for this workflow, to terminated.

> if this blanket update might potentially hide some unexpected behaviour (say for some reason we have deployments that are somehow running or serving but are not the current deployment

This is in fact the case today. But it should go away after the DB migration captured in this PR.

ui/src/pages/version/HistoryDetails.js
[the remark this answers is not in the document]
ariefrahmansyah: Oops, my bad. Here are the updated ones:

image

image

docs/dev-guide/deployments_table_migration.md
ariefrahmansyah: Oh, we can. Updated here: https://github.com/caraml-dev/merlin/pull/483/files#diff-39f6fa7711419b318e134517b1068e6ffb81268adc7f36fa0d215a58af5ea697

Btw, the down.sql file is empty though. Because it's backward compatible after we update the deployment status. Do you have any suggestions?

deadlycoconuts: The script looks good to me! :)

api/storage/deployment_storage.go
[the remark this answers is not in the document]
krithika369: @ariefrahmansyah perhaps this comment was overlooked. Could you review / address this too ?

api/storage/deployment_storage.go
```diff
@@ -83,3 +89,87 @@ func (d *deploymentStorage) GetFirstSuccessModelVersionPerModel() (map[models.ID
...
+		newDeployment.VersionModelID, newDeployment.VersionID, newDeployment.VersionEndpointID).Find(&deployments).Error
+	if err != nil {
+		return err
+	}
+
+	for i := range deployments {
+		// Update the new deployment
+		if deployments[i].ID == newDeployment.ID {
```
krithika369, line 126: Instead of this, shall we just cover this id != newDeployment.ID in the WHERE clause of the SQL query above?

api/storage/deployment_storage.go
```diff
@@ -83,3 +89,87 @@ func (d *deploymentStorage) GetFirstSuccessModelVersionPerModel() (map[models.ID
...
+	defer func() {
+		if err != nil {
+			tx.Rollback()
+		} else {
+			tx.Commit()
+		}
+	}()
+
+	var deployments []*models.Deployment
+	err = tx.Where("version_model_id = ? AND version_id = ? AND version_endpoint_id = ? AND status IN ('pending', 'running', 'serving')",
```
krithika369, line 159: Related to this thread: https://github.com/caraml-dev/merlin/pull/483/files#r1391955411

I'm not sure "pending" should be included here either.

db-migrations/33_update_multiple_deployments.up.sql
```diff
@@ -0,0 +1,31 @@
+-- For terminated endpoint, update the status of all successful deployments to 'terminated'
```
krithika369, line 1: For both queries below, were you able to test the performance on realistic data (perhaps a dump of the staging database)?

I'm worried that, for large volumes of data, the performance of the subquery will be poor and we might be better off using appropriate GROUP BY / JOINS to achieve the desired result.

ui/src/pages/version/HistoryDetails.js
krithika369: Got it. Thanks for the clarification!

krithika369: > deployments table might have the row with status in running and serving but with error message

Hmm.. Does it matter? If the deployment ended up to be "running" / "serving" despite the error, it should still be considered the current deployment, right?

ui/src/pages/version/HistoryDetails.js
```diff
@@ -14,48 +14,50 @@ import { useMerlinApi } from "../../hooks/useMerlinApi";
...
+          deployment.status === "terminated") &&
+        deployment.error === ""
+      );
     });
 
-    orderedDeployments = orderedDeployments.sort((a, b) =>
-      a.id < b.id ? 1 : -1
-    );
-    setOrderedDeployments(orderedDeployments);
+    // If no successful deployment, use the first deployment as current
```
krithika369, line 55: What if there is an active "pending" deployment? Should we pick the first "failed" deployment here?

So, in general, we have 3 cases:
* There is a serving / running / terminated deployment - pick the first one we find as current, by descending order of deployment id and use it (note - at most one of {serving, running} may be found)
* No serving / running / terminated deployment but there is a failed deployment - the latest one should be considered current
* No serving / running / terminated / failed deployments (both conditions above not satisfied) - this means, we only have a single deployment (we already checked `if (deployments.length < 1)` above) and it's in the pending state (guess we can skip the label for this).

Does it make sense?

krithika369 (commented):
@ariefrahmansyah Left some more comments. One the whole, LGTM. Will have another look when the open comments are resolved, as some of them may require some thought. Thanks!

db-migrations/33_update_multiple_deployments.up.sql
ariefrahmansyah: I have tested both on our dev and staging environment. For staging alone which have around 40K data, the migration only take 1 second.

api/storage/deployment_storage.go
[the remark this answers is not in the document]
ariefrahmansyah: I'll remove the 'pending' state on both `OnDeploymentSuccess` and `Undeploy` methods as I agree that 'pending' state should be handled differently.

> is it overkill to be updating all deployments that have the running/serving status to terminated?

Once DB migration executed, there should be at most one running or serving deployment status for a given model version endpoint. So the query will only update at most one row.

> if this blanket update might potentially hide some unexpected behaviour (say for some reason we have deployments that are somehow running or serving but are not the current deployment - would updating their status in the DB would hide some potential underlying problem?)

This should be fine. As suggested in your question, running or serving deployments that not the current revision are safe to be updated to terminated status as it's not being use everywhere -- only on deployment history log UI.

ui/src/pages/version/HistoryDetails.js
ariefrahmansyah: Oh, you're right. We don't need to check deployment.error value anymore if the status is already running or serving.

api/queue/work/model_service_deployment.go
ariefrahmansyah: Thanks, Leon for sharing your concern and idea. I'm agree that the separation of concern between service and storage layers in Merlin codebase is not clear enough leading the confusion on how to implement both of them. I also agree that service is responsible for "what to save" and storage layer is responsible for "how to save the data".

However, this might need more time and thinking to achieve the best solution, so for now I'll create GitHub issue to track this: https://github.com/caraml-dev/merlin/issues/490 so we can revisit it later.

leonlnj: yep not expecting to be done here at all

ui/src/pages/version/HistoryDetails.js
ariefrahmansyah: Yup, it makes sense. Thanks for the input!

For only Pending deployment:
image

ariefrahmansyah: For Pending and Terminated:

image

api/storage/deployment_storage.go
ariefrahmansyah: Updated

docs/dev-guide/deployments_table_migration.md
```diff
@@ -0,0 +1,85 @@
...
+
+1. For version endpoints in terminated status, and
+2. For version endpoints in pending/running/serving status
+
+### For terminated version endpoints
+
+If endpoint status = terminated, update all succesful deployments status to terminated
+
+```
+UPDATE
```
krithika369: Why was the `db-migrations/33_update_multiple_deployments.up.sql` file removed? Can we add it back and reference it here in the doc? It's ok that the `.down.sql` file is empty. Having the migration file will make it easy to roll out the migration to all environments as well.

krithika369 (approved):
Left one comment. The rest LGTM. Thanks, @ariefrahmansyah !

docs/dev-guide/deployments_table_migration.md
ariefrahmansyah: It's not removed as I can see it in the latest commit of this PR's branch (`redeploy-update-deployments-table`): https://github.com/caraml-dev/merlin/blob/redeploy-update-deployments-table/db-migrations/33_update_multiple_deployments.up.sql

Seems like you were looking at previous commit.

image

PR #484, opened 2023-11-02
Add image builder's prometheus metrics: attempts, results, and duration

ariefrahmansyah:

For better observability across the image builder process, I added 3 Prometheus metrics: attempts, results, and durations.

The results in `/metrics` endpoint would be something like this:

```
# HELP merlin_image_builder_attempts Total of image builder attempts
# TYPE merlin_image_builder_attempts counter
merlin_image_builder_attempts{model="pyfunc-sample",model_type="pyfunc",model_version="4",project="string-theory",python_version="3.8.*"} 1

# HELP merlin_image_builder_duration_seconds Duration of image builder in seconds
# TYPE merlin_image_builder_duration_seconds histogram
merlin_image_builder_duration_seconds_bucket{model="pyfunc-sample",model_type="pyfunc",model_version="4",project="string-theory",python_version="3.8.*",result="success",le="420"} 1
...
merlin_image_builder_duration_seconds_sum{model="pyfunc-sample",model_type="pyfunc",model_version="4",project="string-theory",python_version="3.8.*",result="success"} 385.718119541
merlin_image_builder_duration_seconds_count{model="pyfunc-sample",model_type="pyfunc",model_version="4",project="string-theory",python_version="3.8.*",result="success"} 1

# HELP merlin_image_builder_results Total of image builder results
# TYPE merlin_image_builder_results counter
merlin_image_builder_results{model="pyfunc-sample",model_type="pyfunc",model_version="4",project="string-theory",python_version="3.8.*",result="success"} 1
```

NONE

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/pkg/imagebuilder/imagebuilder.go
+++ b/api/pkg/imagebuilder/imagebuilder.go
@@ -24,6 +24,8 @@ import (
 	"time"
 
 	backoff "github.com/cenkalti/backoff/v4"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -43,6 +45,27 @@ import (
...
 var maxCheckImageRetry uint64 = 2
 
+var (
+	imageBuilderAttempts = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "merlin",
+		Name:      "image_builder_attempts",
+		Help:      "Total of image builder attempts",
+	}, []string{"project", "model", "model_type", "model_version", "python_version"})
+
+	imageBuilderResults = promauto.NewCounterVec(prometheus.CounterOpts{
```
krithika369, line 55: The histogram metric below comes with a counter as well (`merlin_image_builder_duration_seconds_count`), right? Perhaps we don't need this separate results metric.

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -43,6 +45,27 @@ import (
...
+		Namespace: "merlin",
+		Name:      "image_builder_results",
+		Help:      "Total of image builder results",
+	}, []string{"project", "model", "model_type", "model_version", "python_version", "result"})
+
+	imageBuilderDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
krithika369, line 61: Do you think the attempt # can also be a label in this metric instead of a separate `image_builder_attempts` metric? No strong opinion on this though.

krithika369 (approved):
LGTM. Left some comments. Since the image builder improvements are in progress now (and may go through some changes in the near term), don't need to act on this right away. But we should propagate this change to [Turing](https://github.com/caraml-dev/turing) too, when stable. Thanks, @ariefrahmansyah !

api/pkg/imagebuilder/imagebuilder.go
ariefrahmansyah: Attempt metrics should be separate counter metrics because they should be decoupled with the results of the image building process. If image building process fails, we still have the counter of the attempts. The ideal scenario is attempts == results, so if the count of attempts.

Also, having attempt metrics as label for duration metrics will double the size of duration metrics.

api/pkg/imagebuilder/imagebuilder.go
ariefrahmansyah: This is good suggestion but I prefer to explicitly have results metrics instead of bury the result information in `merlin_image_builder_duration_seconds_count` metrics. `merlin_image_builder_results` as metrics name is easier to find and understand.

PR #488, opened 2023-11-09
feat: consumer worker for prediction log

khorshuheng:
This PR add a consumer worker, which will consume prediction logs published by Merlin models and send them to Arize AI. This is not expected to be deployed automatically yet.

Fixes #

None

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


python/observability/conf/environment/example-override.yaml
```diff
@@ -0,0 +1,23 @@
...
+    prediction_column: "accept"
+    prediction_column_type: "BOOLEAN"
+    timestamp_column: "request_timestamp"
+    feature_columns:
+      - "distance"
+      - "transaction"
+    feature_column_types:
+      - "INT32"
+      - "FLOAT64"
+arize:
```
krithika369, line 15: Although it's unlikely we will provide integrations with multiple providers any time in the near future, shall we design the configuration and classes such that an `observability_backend` is taken into account? Eg:

```yaml
observability_backend:
    type: arize
    arize_config:
        api_key: "SECRET_API_KEY"
        space_key: "SECRET_SPACE_KEY"
```

python/observation-publisher/publisher/observability_backend.py
```diff
@@ -0,0 +1,47 @@
...
+class ObservationSink(abc.ABC):
+    @abc.abstractmethod
+    def write(self, dataframe: pd.DataFrame):
+        raise NotImplementedError
+
+
+class ArizeSink(ObservationSink):
+    def __init__(self, config: ArizeConfig, model_spec: ModelSpec):
+        self._client = Client(space_key=config.space_key, api_key=config.api_key)
+        self._schema = Schema(
```
tiopramayudi, line 20: please set `prediction_id_column`

tiopramayudi, line 20: the arize schema is depending on the model type, you need to check for each model type what are the required fields. For example for ranking model we need to specify `prediction_group_id_column_name` also it needs `rank_column_name` being set. You can refer to [this](https://docs.arize.com/arize/model-types/ranking) for ranking, and you can check for other model types

python/observation-publisher/publisher/prediction_log_consumer.py
```diff
@@ -0,0 +1,91 @@
...
+            print(f"Last encountered error: {errors[-1]}")
+            raise KafkaException(errors[-1])
+        prediction_logs = [
+            parse_message_to_prediction_log(msg.value().decode("utf-8"))
+            for msg in messages
+            if (msg is not None and msg.error() is None)
+        ]
+        if len(prediction_logs) > 0:
+            df = log_to_dataframe(prediction_logs, model_cfg.schema)
+            observation_sink.write(dataframe=df)
```
tiopramayudi, line 43: this can potentially cause lagging, can we publish it async?

python/observation-publisher/conf/environment/example-override.yaml
```diff
@@ -0,0 +1,31 @@
...
+  type: "ARIZE"
+  arize_config:
+    api_key: "SECRET_API_KEY"
+    space_key: "SECRET_SPACE_KEY"
+kafka:
+  topic: "test-topic"
+  bootstrap_servers: "localhost:9092"
+  group_id: "test-group"
+  auto_offset_reset: "latest"
+  batch_size: 100
```
tiopramayudi, line 31: this is number of row?

python/observation-publisher/publisher/config.py
```diff
@@ -0,0 +1,87 @@
...
+    INT64 = 2
+    BOOLEAN = 3
+    STRING = 4
+
+
+@dataclass
+class ModelSchema:
+    prediction_columns: List[str]
+    prediction_column_types: List[ValueType]
+    prediction_score_column: str
```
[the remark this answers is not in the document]
khorshuheng: We can use prediction schema, though the prediction schema alone doesn't tell us the sequence of the fields in the array. For example, the incoming message could be of the form of [prediction_label_1, prediction_label_2, prediction_score] or [prediction_score, prediction_label_1, prediction_label_2].

We can implicitly assumes that prediction labels always follow after prediction_score, and assumes that the producer will always obey this convention. But, if this convention is violated by the producer, either mistakenly or some refactoring, then the consumer will send the wrong result but yet not throwing any error.

python/observation-publisher/conf/environment/example-override.yaml
khorshuheng: Yes, this is number of rows as opposed to size in bytes.

python/observation-publisher/publisher/prediction_log_consumer.py
```diff
@@ -0,0 +1,91 @@
...
+def poll_prediction_logs(
+    kafka_consumer: Consumer,
+    consumer_cfg: KafkaConsumerConfig,
+    observation_sink: ObservationSink,
+    model_cfg: ModelSpec,
+):
+    kafka_consumer.subscribe([consumer_cfg.topic])
+    while True:
+        messages = kafka_consumer.consume(
+            consumer_cfg.batch_size, timeout=consumer_cfg.poll_timeout_seconds
```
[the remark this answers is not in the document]
khorshuheng: Kafka consumer API doesn't allow us to poll based on batch size in bytes, only maximum number of rows. We could, however, still control the min/max bytes of data return per polling request via Kafka consumer configuration.

The larger the size per batch, the greater is the latency. We also need to be careful of session timeout. So at this point we can't say for sure if we can definitely guarantee at least 50Mb per poll, will have to tweak the configuration and check. What happens if we we send less than 50Mb?

python/observation-publisher/publisher/prediction_log_consumer.py
khorshuheng: We could publish it async, but that comes at the risk of data loss. For example, if a commit has been performed before the write is completed, all the messages will be lost forever should the write ultimately failed. Unlike producer, there are a lot more messages in a single polling request (especially if we want to fulfil the min batch size of 50Mb per poll).

python/observation-publisher/publisher/config.py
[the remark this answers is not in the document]
tiopramayudi: why do you need the order? as long as the defined columns in the schema are there in the message it should be fine right?

python/observation-publisher/publisher/prediction_log_consumer.py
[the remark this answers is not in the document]
tiopramayudi: it adds backpressure to their system, and they advice as to use bigger batch size when using batch api, if not we need to use single prediction API.

we can add a variable to holds the data until reach the minimum size though, then once minimum size is reached and published we empty the variable again

python/observation-publisher/publisher/config.py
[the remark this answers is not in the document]
khorshuheng: Ah yes, you are right. I check the sample message again, and column names are part of the message itself. I will remove the prediction columns from the schema.

python/observation-publisher/publisher/prediction_log_consumer.py
khorshuheng: Yes, it will be the same set of messages.

python/observation-publisher/publisher/prediction_log_consumer.py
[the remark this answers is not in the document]
khorshuheng: I see. Yes, we can certainly poll multiple times until a certain size threshold is met before sending them to Arize. Let's investigate whether it is possible to send a minimum of 50Mb in a single poll, failing that, we can handle this on our end.

python/observation-publisher/publisher/config.py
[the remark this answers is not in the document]
khorshuheng: Updated the code to use column types mapping as opposed to having feature and prediction names in the schema.

python/observation-publisher/conf/environment/example-override.yaml
```diff
@@ -0,0 +1,47 @@
...
+    # If not provided, it's assumed to be prediction_id.
+    prediction_id_column: "prediction_id"
+observability_backend:
+  # Supported backend types:
+  #  - ARIZE
+  type: "ARIZE"
+  arize_config:
+    api_key: "SECRET_API_KEY"
+    space_key: "SECRET_SPACE_KEY"
+consumer:
```
krithika369, line 36: Nit: Instead of `consumer`, would `observations_source` be more meaningful? The "consumer" seems more suitable for Kafka but not as much for, say BQ source.

python/observation-publisher/publisher/prediction_log_consumer.py
```diff
@@ -0,0 +1,150 @@
...
+        raise ValueError(f"Unknown consumer type: {config.type}")
+
+
+def parse_message_to_prediction_log(msg: str) -> PredictionLog:
+    log = PredictionLog()
+    log.ParseFromString(msg)
+    return log
+
+
+def convert_value(
```
krithika369, line 96: Nit: Would it be clearer to have `convert_value_to_np` or something like that to indicate what conversion is happening? Same comment about the list conversion function below.

python/observation-publisher/tests/test_prediction_log_consumer.py
```diff
@@ -0,0 +1,194 @@
+from datetime import datetime
```
krithika369, line 1: I suppose the CI for the observation-publisher would come in later. 

In the meantime, can we start adding some basic dev docs (eg: running the tests locally), in a README file for this module?

krithika369 (commented):
Left some small comments. Perhaps some of @tiopramayudi's comments are still being addressed. The rest LGTM. Thanks, @khorshuheng !

python/observation-publisher/publisher/observation_log_consumer.py
```diff
@@ -0,0 +1,149 @@
...
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def close(self):
+        raise NotImplementedError
+
+    def start_polling(self, observation_sink: ObservationSink, model_spec: ModelSpec):
+        try:
+            while True:
+                logs = self.poll_new_logs()
```
tiopramayudi, line 36: how is the batching with certain bytes size? is it possible from kafka consumer?

python/observation-publisher/publisher/config.py
```diff
@@ -0,0 +1,160 @@
...
+        ct = {self.prediction_label_column: ValueType.STRING}
+        if self.prediction_score_column is not None:
+            ct[self.prediction_score_column] = ValueType.FLOAT64
+        return ct
+
+
+@dataclass
+class MulticlassClassificationPredictionOutput:
+    prediction_label_column: str
+    prediction_score_column: Optional[str] = None
```
tiopramayudi, line 42: for multiclass the prediction_score_column must be in list type, because the scores are calculated for all possible class

python/observation-publisher/publisher/config.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @property
+    def column_types(self) -> Dict[str, ValueType]:
+        return {self.prediction_score_column: ValueType.FLOAT64}
+
+
+@dataclass
+class BinaryClassificationPredictionOutput:
+    prediction_label_column: str
+    prediction_score_column: Optional[str] = None
```
tiopramayudi, line 29: in arize the prediction_score is optional, but from model standpoint the prediction score is mandatory and prediction label is optional (can be derived from prediction score). We need to update this later on when introducing model schema

python/observation-publisher/publisher/observation_log_consumer.py
khorshuheng: No, that's not possible, unfortunately. We will have to implement it ourselves if its necessary.

python/observation-publisher/publisher/config.py
khorshuheng: Noted, will add a TODO inline comment.

python/observation-publisher/publisher/observation_log_consumer.py
tiopramayudi: do you plan to update in this PR or separate PR?

khorshuheng: We could control the number of rows in a batch, so that should have suffice. If it's not, i will implement this in a separate PR.

python/observation-publisher/publisher/observation_log_consumer.py
tiopramayudi, line 1: should this call `prediction_log_consumer` ?

tiopramayudi (approved):
LGTM, thanks @khorshuheng ! I add one small comment

python/observation-publisher/publisher/prediction_log_parser.py
```diff
@@ -0,0 +1,73 @@
...
+@dataclass
+class PredictionLogFeatureTable:
+    columns: List[str]
+    rows: List[List[Union[np.int64, np.float64, np.bool_, np.str_]]]
+
+
+@dataclass
+class PredictionLogResultsTable:
+    columns: List[str]
+    rows: List[List[List[Union[np.int64, np.float64, np.bool_, np.str_]]]]
```
tiopramayudi, line 20: the type should be `List[List[Union[np.int64, np.float64, np.bool_, np.str_]]]` right ?

python/observation-publisher/publisher/prediction_log_consumer.py
```diff
@@ -0,0 +1,137 @@
...
+def parse_message_to_prediction_log(msg: str) -> PredictionLog:
+    log = PredictionLog()
+    log.ParseFromString(msg)
+    return log
+
+
+def log_to_records(
+    log: PredictionLog, inference_schema: InferenceSchema
+) -> Tuple[List[List[np.int64 | np.float64 | np.bool_ | np.str_]], List[str]]:
+    request_timestamp = log.request_timestamp.ToDatetime()
```
tiopramayudi, line 103: is this supported in arize? should we use the timestamp in second instead? based on this https://docs.arize.com/arize/sending-data-guides/model-schema-reference#6.-timestamp it expect integer unix second

python/observation-publisher/publisher/observability_backend.py
```diff
@@ -0,0 +1,109 @@
...
+            case InferenceType.MULTICLASS_CLASSIFICATION:
+                self._schemas = [
+                    Schema(
+                        feature_column_names=inference_schema.feature_columns,
+                        prediction_label_column_name=prediction_label_column,
+                        prediction_score_column_name=prediction_score_column,
+                        prediction_id_column_name=inference_schema.prediction_id_column,
+                        timestamp_column_name=PREDICTION_LOG_TIMESTAMP_COLUMN,
+                    )
+                    for prediction_label_column, prediction_score_column in zip(
```
tiopramayudi, line 55: usually in multi class the prediction_score_column is multiple depending on the classes but the prediction_label_column must be one usually the label for the highest prediction_score, how will it work with `zip`?

tiopramayudi: since IMO the use case of multiclass is not many currently, I think we can revisit this later need to check how the DS use the multiclass in production

tiopramayudi (approved):
Adding small feedback, the rest is LGTM

python/observation-publisher/publisher/observability_backend.py
tiopramayudi: so the assumption in here that the prediction_score_column has the same length with the prediction_label_column then, the sample row will look like the on in this https://docs.arize.com/arize/model-types/multiclass-classification#case-2-supports-classification-and-auc-metrics?

khorshuheng: Yes, that's the expectation.

python/observation-publisher/publisher/prediction_log_consumer.py
khorshuheng: https://github.com/Arize-ai/client_python/blob/main/arize/pandas/validation/validator.py#L1013

Timestamp is accepted too. I have added a test case for validation.

python/observation-publisher/publisher/observability_backend.py
khorshuheng: Yes that's the assumption.

PR #489, opened 2023-11-09
 Update PyFunc and support publishing log from PyFunc

tiopramayudi:

To onboard to model observability, we need to gather features and prediction value of the model, per current condition in PyFunc model we can't get all the features needed for the model and the prediction that generated by the model, since the input and output of the PyFunc model is not necessary features and prediction value. This PR try to solve that by adding new PyFunc model that can identify which one is features and prediction value. Once the data is identified, then it will be published to kafka for later processing

Modification:
* `python/sdk/merlin/pyfunc.py`
  * Adding new `PyFuncV3Model` to differentiate features, and prediction value
  * Introducing `PyFuncOutput` as the single output for realtime Pyfunc (`PyFuncModel` and `PyFuncV3Model)
* `python/pyfunc-server/pyfuncserver/config.py` - Adding configuration for kafka publishing and sampling ratio
* `python/pyfunc-server/pyfuncserver/protocol/rest/handler.py` - Add async publishing after get pyfunc model output
*  `python/pyfunc-server/pyfuncserver/protocol/rest/server.py` 
  * Create subclass of tornado web application that will hold kafka producer instance
* `python/pyfunc-server/pyfuncserver/publisher/publisher.py` - Adding asyncio publisher code
* `python/pyfunc-server/pyfuncserver/sampler/sampler.py` - Random sampling method base on the given ration
* `python/pyfunc-server/pyfuncserver/publisher/kafka.py` - Kafka producer implementation given PyFuncOutput

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation


python/pyfunc-server/pyfuncserver/config.py
```diff
@@ -68,6 +101,33 @@ def __init__(self, model_dir: str):
         self.push_gateway = PushGateway(push_enabled,
                                         push_url,
                                         push_interval)
+        
+        # Publisher
+        sampling_ratio = float(os.getenv(*PUBLISHER_SAMPLING_RATIO))
+        publisher_enabled = str_to_bool(os.getenv(*PUBLISHER_ENABLED))
```
krithika369, line 107: Should this be the first check and then we parse / initialise all other fields based on whether this is enabled? For example, if publisher is disabled, we don't have to even bother reading the other config variables, `self.publisher` can be `None`, etc. Wdyt?

python/pyfunc-server/pyfuncserver/protocol/rest/handler.py
```diff
@@ -57,12 +57,14 @@ def post(self, full_name: str):
...
 
-        response = model.predict(request, headers=headers)
-
-        response_json = orjson.dumps(response)
+        output = model.predict(request, headers=headers)
+        response_json = orjson.dumps(output.http_response)
         self.write(response_json)
         self.set_header("Content-Type", "application/json; charset=UTF-8")
 
+        if self.publisher is not None and output.contains_prediction_log:
```
krithika369, line 65: Do we not need to invoke this method as `output.contains_prediction_log()` ? Otherwise it will always evaluate to `True`.

krithika369 (commented):
@tiopramayudi, can I request for more details in the PR description on the actual changes covered by this PR? Generally, 2 pieces of info are very important:
1. Motivation for the changes (you've included this)
2. A brief summary of the code changes - this should highlight the key files changed. This is particularly useful when there are many files changes in the PR and going through the changes sequentially would not be helpful. So, something like:
    * `python/sdk/merlin/pyfunc.py` - Introducing `PyFuncV3Model` class.
    * `python/pyfunc-server/pyfuncserver/model/model.py` - Updating predict method to return additional info for the observation
    * ...

This would make it easier to navigate the PR and preserve some context for anyone referring to this PR in the future. Thanks.

python/sdk/pyfunc.Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Copyright 2020 The Merlin Authors
```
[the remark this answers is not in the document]
tiopramayudi: ah this one is generated after running unit-test, will remove this

python/pyfunc-server/pyfuncserver/config.py
tiopramayudi: yeah we can do that, I'll update that

python/pyfunc-server/pyfuncserver/protocol/rest/handler.py
tiopramayudi: 🤦 I forgot to add `()`, will add that

python/sdk/merlin/pyfunc.py
```diff
@@ -109,7 +117,190 @@ def upiv1_infer(self, request: upi_pb2.PredictValuesRequest,
...
+        """
+        preprocess is the method that doing preprocessing before calling the ml framework for prediction.
+        This method is called when user use HTTP_JSON protocol
+        :param request: raw input to the model (dict)
+        :return: model input, this model input must already have all the features that required for model prediction
+        """
+        raise NotImplementedError("preprocess is not implemented")
+
+    @abstractmethod
+    def ml_predict(self, model_input: ModelInput) -> ModelOutput:
```
krithika369, line 262: To match the `PyFuncV2Model` class, should we call this method `infer` ?

Btw, on a related note, I don't see the [preprocess](https://github.com/caraml-dev/merlin/blob/main/python/sdk/merlin/pyfunc.py#L166) and [postprocess](https://github.com/caraml-dev/merlin/blob/main/python/sdk/merlin/pyfunc.py#L177) methods of the `PyFuncV2Model` class being called anywhere. Am I missing something?

krithika369 (approved):

krithika369:
Hi @tiopramayudi, thanks for the PR and the details in the PR description! It largely LGTM. I left some comments, mostly minor. But I'd especially like to clarify / address these before merging this PR:
* https://github.com/caraml-dev/merlin/pull/489#discussion_r1398308470
* https://github.com/caraml-dev/merlin/pull/489#discussion_r1398302161

python/sdk/merlin/pyfunc.py
tiopramayudi: honestly I don't have enough context why there is `preprocess` and postprocess` in the `PyFuncV2Model` in the comment itself stated that it is not called during batch inference. Actually the only method that is called is `infer` same like `PyFuncModel`

I'm prefer not to use `infer` method because user uses this to do all of the preprocessing, ml prediction, and postprocess in this method in `PyFuncModel`,  use `infer` method will add ambiguity IMO

python/pyfunc-server/pyfuncserver/sampler/sampler.py
```diff
@@ -0,0 +1,31 @@
+import random
+from abc import ABC, abstractmethod
+from typing import Optional
+
+class Sampler(ABC):
+    @abstractmethod
+    def should_sample(self, request_id: Optional[int] = None) -> bool:
+        pass
+
+class RatioSampling(Sampler):
```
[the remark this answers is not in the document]
tiopramayudi: I used opentelemetry as reference, you can check the implementation [here](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/sampling.py#L253) , I'm honestly don't know how the distribution looks like with the suggested above, that's why I used the one that adopted by the well known library

python/pyfunc-server/pyfuncserver/protocol/rest/server.py
```diff
@@ -22,34 +25,50 @@ async def sig_handler(server):
...
+        self.config = config
         self.workers = config.workers
+        self.model_manifest = config.model_manifest
         self.http_port = config.http_port
         self.metrics_registry = metrics_registry
         self.registered_models: dict = {}
         self.register_model(model)
 
     def create_application(self):
-        return tornado.web.Application([
```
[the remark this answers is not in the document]
tiopramayudi: tornado web application needs to holds the publisher information that will be used for each request, previously I add it in the handler part, but the handler is created per incoming request, hence put the publisher information in the application so there won't be initialization of publisher everytime new request is coming

python/pyfunc-server/pyfuncserver/protocol/upi/server.py
```diff
@@ -3,22 +3,37 @@
...
+        self._publisher: Optional[Publisher] = None
+
+    def set_publisher(self, publisher: Publisher):
+        if self._publisher is None:
+            self._publisher = publisher
 
     def PredictValues(self, request, context):
-        return self._model.upiv1_predict(request=request, context=context)
+        output = self._model.upiv1_predict(request=request, context=context)
+        if self._publisher is not None and output.contains_prediction_log:
```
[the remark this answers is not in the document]
tiopramayudi: > Is it ever possible that `output.contains_prediction_log` evaluates to `False` if `self._publisher is not None`?

it is possible, from pyfunc server standpoint, it doesn't know whether the model that is used is `PyFuncV3Model` or `PyFuncModel. If the model `PyFuncModel` the output won't contains prediction log.

python/pyfunc-server/pyfuncserver/sampler/sampler.py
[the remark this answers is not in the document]
krithika369: Got it. Could you leave an inline comment with the link for reference? It's better to use the latest version [v1.21.0](https://github.com/open-telemetry/opentelemetry-python/blob/v1.21.0/opentelemetry-sdk/src/opentelemetry/sdk/trace/sampling.py#L253) rather than the main branch.

python/sdk/merlin/pyfunc.py
krithika369: > honestly I don't have enough context why there is preprocess and postprocessin thePyFuncV2Modelin the comment itself stated that it is not called during batch inference. Actually the only method that is called isinfersame likePyFuncModel`

@ariefrahmansyah, any clue about this?

krithika369: > I'm prefer not to use infer method because user uses this to do all of the preprocessing, ml prediction, and postprocess in this method in PyFuncModel, use infer method will add ambiguity IMO

This is a valid concern. But I still think `infer` is better, for consistency. I also feel that `ml_predict` can be a bit restrictive (what if the model is non-ML based?). There is also no harm if the user does the pre and post process inside `infer`, unless they are enabling model monitoring, right? At the end of the day, if the user choses to use the V3 model version, I think the inline comments on each of the methods can clearly capture the purpose; for the `infer` method, we can mention that pre/post processing is recommended to be done in the other respective methods.

python/pyfunc-server/pyfuncserver/protocol/upi/server.py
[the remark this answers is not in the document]
krithika369: Ahh makes sense. Thanks for the clarification. Could you also add this rationale as an inline comment, for future reference?

python/sdk/merlin/pyfunc.py
ariefrahmansyah: -> Why there is a preprocess and postprocess in the PyFuncV2?

Because when we first design batch prediction case in Merlin, we envisioned PyFuncV2 to be able to be implemented in both online and batch case.

For example, case 1: User implements preprocess, infer, and postprocess methods
- This model implementation is reusable for both online prediction and batch prediction
- If it’s deployed as online prediction the request will be served using preprocess->infer->postprocess sequence.
- If it’s deployed as batch prediction the infer method will be used to calculate the inference result. 
This model implementation is intended for reusability, but sacrifices performance slightly for online prediction.

image

Case 2: User implements raw_infer, preprocess, infer, and postprocess methods
- This model implementation is also reusable for both online prediction and batch prediction
- Compared to case 1, the online prediction will only use the raw_infer method to serve incoming requests. This is intended so that users shouldn’t need to convert the incoming request into pandas Dataframe which might incur additional latency. 
- This model implementation is intended for both performance and reusability, but with extra duplication (since the user will have to implement model inference in both raw_infer and infer methods)
- 
image

ariefrahmansyah: However, preprocess, postprocess, and raw_infer are not called by the current pyfunc server. So to implement it, users need to call preprocess and postprocess from their infer method. And for raw_infer, there's no way for users to call it so it's actually not implemented by our users.

python/pyfunc-server/test/test_http.py
```diff
@@ -5,16 +5,38 @@
...
+                data=[["order1", "1111"], ["order1", "2222"], ["order1", "3333"]]
+            ),
+            session_id="randomSessionID"
+        )
+sample_model_output = ModelOutput(
+            prediction_ids=["prediction_1", "prediction_2", "prediction3"],
+            predictions=Values(
+                columns=["prediction_score", "prediction_label"],
+                data=[[0.9, "fraud"], [0.2, "not fraud"], [0.6, "fraud"]]
+            )
```
khorshuheng, line 37: Can we add row_id here as well?

khorshuheng: Nvm, i just realized that row id is not part of the Values type.

python/sdk/merlin/pyfunc.py
```diff
@@ -109,7 +114,190 @@ def upiv1_infer(self, request: upi_pb2.PredictValuesRequest,
...
+class ModelInput:
+    # unique identifier of each prediction
+    prediction_ids: List[str]
+    # features data for model prediction, the length of data of features must be the same as `prediction_ids` length
+    features: Values
+    # entities data is additional data that are not used for prediction, but this data is used to retrieved another features.
+    # The length of data of entities must be the same as `prediction_ids` length
+    entities: Optional[Values] = None
+    # session id is identifier for the request
+    session_id: str = ""
```
khorshuheng, line 140: How is this session_id going to be used?

python/sdk/merlin/pyfunc.py
krithika369: Got it. Thanks for the detailed explanation, @ariefrahmansyah ! Given that the current state of the code for PyFuncV2 is only partially implementing these, do you see any issues removing these abstract methods (`preprocess`, `postprocess` and `raw_infer`) from PyFuncV2 entirely? That will prevent any confusion about their existence. If, by any chance, the user has implemented these in their concrete class, it should still continue to work.

ariefrahmansyah: -> do you see any issues removing these abstract methods (preprocess, postprocess and raw_infer) from PyFuncV2 entirely?

No, I don't foresee any issue. It should be safe to remove them.

krithika369: Thanks for the confirmation, @ariefrahmansyah! @tiopramayudi, if you don't have any objections either, please go ahead and remove those methods from the PyFuncV2 class. Thanks!

PR #492, opened 2023-11-17
Model deployment support for model observability

tiopramayudi:

In the previous [PR](https://github.com/caraml-dev/merlin/pull/489) we introduce `PyFuncV3Model` to support model observability use case. What missing from previous PR is deployment process of `PyFuncV3Model`, hence this PR is raised.

Modification:
* `api/config/config.go` - Adding `PyFuncPublisherConfig` new configuration for pyfunc publisher
* `api/config/environment.go` - Update parsing `DeploymentConfig` to add relevant config to the deployment config
* `db-migrations/34_version_endpoints_enable_observability.*.sql` - Adding model observability enable flag to version endpoint
* `api/models/model.go` - Adding `pyfunc_v3` model type
* `api/cluster/resource/templater.go` 
  * All deployment config will be part of `deploymentConfig` including standard transformer configuration
  * Add required env vars to be passed for `PyFuncV3Model`
* `api/api/version_endpoints_api.go` - Adding `PyFuncV3Model` as supported model type for UPI
* `python/sdk/test/pyfunc_integration_test.py` - Adding integration test for `PyFuncV3Model`

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/cluster/resource/templater.go
```diff
@@ -79,6 +79,13 @@ const (
 	defaultGRPCPort                  = 9000
 	defaultPredictorPort             = 80
 
+	envPublisherKafkaTopic    = "PUBLISHER_KAFKA_TOPIC"
+	envPublisherKafkaBrokers  = "PUBLISHER_KAFKA_BROKERS"
+	envPublisherEnabled       = "PUBLISHER_ENABLED"
+	envPublisherKafkaLinger   = "PUBLISHER_KAFKA_LINGER_MS"
+	envPublisherKafkaAck      = "PUBLISHER_KAFKA_ACKS"
+	envPublisherSamplingRatio = "PUBLISHER_SAMPLING_RATIO"
+
```
krithika369, line 88: Do we also want to expose the [PUBLISHER_KAFKA_CONFIG](https://github.com/caraml-dev/merlin/blob/main/python/pyfunc-server/pyfuncserver/config.py#L31) env var via the Merlin control plane configs?

api/cluster/resource/templater_test.go
```diff
@@ -714,6 +726,127 @@ func TestCreateInferenceServiceSpec(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "pyfunc spec with model observability enabled; there is no effect ",
```
krithika369, line 730: If a user tries to enable model observability with `ModelTypePyFunc` or even the std model types for now, it would be better to let the API return an error, right? Shall we update `createPredictorSpec` function to return an error and propagate that?

api/cmd/api/setup.go
```diff
@@ -143,7 +143,7 @@ func initImageBuilder(cfg *config.Config) (webserviceBuilder imagebuilder.ImageB
 	ctl, err := cluster.NewController(
 		clusterCfg,
 		config.DeploymentConfig{}, // We don't need deployment config here because we're going to retrieve the log not deploy model.
```
krithika369, line 145: Not related to this PR but this comment looks wrong 🤔

api/models/service.go
```diff
@@ -73,19 +74,24 @@ func NewService(model *Model, version *Version, modelOpt *ModelOption, endpoint
...
+		CurrentIsvcName:           endpoint.InferenceServiceName,
+		EnabledModelObservability: endpoint.EnableModelObservability,
 	}
 }
 
 func (svc *Service) GetPredictionLogTopic() string {
 	return fmt.Sprintf("caraml-%s-%s-prediction-log", svc.Namespace, svc.ModelName)
 }
 
+func (svc *Service) GetPredictionLogTopicPerVersion() string {
```
krithika369, line 91: Nit: Can we call this `GetPredictionLogTopicForVersion` instead of `GetPredictionLogTopicPerVersion` to indicate that it is for the specific version? "Per" version makes it sound like we will return 1 topic per existing version of the model in the same API call.

krithika369 (approved):
Thanks for this PR and the detailed description, @tiopramayudi! Left some small comments - the rest LGTM.

api/config/config.go
```diff
@@ -336,6 +337,13 @@ type KafkaConfig struct {
 	MaxMessageSizeBytes int    `validate:"required" default:"1048588"`
 	ConnectTimeoutMS    int    `validate:"required" default:"1000"`
 	SerializationFmt    string `validate:"required" default:"protobuf"`
+	LingerMS            int    `validate:"required" default:"0"`
+	Acks                int    `validate:"required" default:"-1"`
```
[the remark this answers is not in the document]
khorshuheng: 0 Acks will give you the best performance at the expense of slight possibility of losing data when there are connectivity issue between producer <-> Kafka broker. Since losing observation log is less impactful that latency increase, we will mostly use 0. We will also use linger ms > 0 under high RPS scenario so that the Kafka cluster does not get overwhelmed.

So, it depends on whether the expectation is for the Merlin administrators/maintainers to always override the default config for the different environment, or if we should set the default config such that it caters for high RPS use case.

api/cluster/resource/templater_test.go
tiopramayudi: yes, it better to throw error, I made change to validation happen in the API layer so it can return early.

python/sdk/test/pyfunc_integration_test.py
```diff
@@ -119,6 +151,36 @@ def test_pyfunc_env_vars(integration_test_url, project_name, use_google_oauth, r
 
     merlin.undeploy(v)
 
+@pytest.mark.pyfunc
+@pytest.mark.integration
+@pytest.mark.dependency()
+def test_pyfunc_model_observability(integration_test_url, project_name, use_google_oauth, requests):
```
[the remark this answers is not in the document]
tiopramayudi: we have separate test file for UPI, for now I'll just update to that file

api/cluster/resource/templater.go
```diff
@@ -281,10 +287,20 @@ func createPredictorSpec(modelService *models.Service, config *config.Deployment
...
+			envVars = append(envVars, models.EnvVar{Name: envGRPCOptions, Value: t.deploymentConfig.PyfuncGRPCOptions})
+		}
+		if modelService.EnabledModelObservability && modelService.Type == models.ModelTypePyFuncV3 {
+			pyfuncPublisherCfg := t.deploymentConfig.PyFuncPublisher
+			envVars = append(envVars, models.EnvVar{Name: envPublisherEnabled, Value: strconv.FormatBool(modelService.EnabledModelObservability)})
+			envVars = append(envVars, models.EnvVar{Name: envPublisherKafkaTopic, Value: modelService.GetPredictionLogTopicForVersion()})
+			envVars = append(envVars, models.EnvVar{Name: envPublisherKafkaBrokers, Value: pyfuncPublisherCfg.Kafka.Brokers})
+			envVars = append(envVars, models.EnvVar{Name: envPublisherKafkaLinger, Value: fmt.Sprintf("%d", pyfuncPublisherCfg.Kafka.LingerMS)})
+			envVars = append(envVars, models.EnvVar{Name: envPublisherKafkaAck, Value: fmt.Sprintf("%d", pyfuncPublisherCfg.Kafka.Acks)})
+			envVars = append(envVars, models.EnvVar{Name: envPublisherSamplingRatio, Value: fmt.Sprintf("%f", pyfuncPublisherCfg.SamplingRatioRate)})
```
khorshuheng, line 302: We might need to add configuration for sampling strategy at a later time. For now, we can keep this as it is.

tiopramayudi:
> The e2e test failed because of:
> 
> ```
> Merlin log
>   {"level":"panic","ts":1701181719.603602,"caller":"api/main.go:101","msg":"Failed validating config: Key: 'Config.StandardTransformerConfig.Kafka.Acks' Error:Field validation for 'Acks' failed on the 'required' tag\nKey: 'Config.PyFuncPublisherConfig.Kafka.Brokers' Error:Field validation for 'Brokers' failed on the 'required' tag\nKey: 'Config.PyFuncPublisherConfig.Kafka.Acks' Error:Field validation for 'Acks' failed on the 'required' tag","stacktrace":"main.main\n\t/src/api/cmd/api/main.go:101\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
>   panic: Failed validating config: Key: 'Config.StandardTransformerConfig.Kafka.Acks' Error:Field validation for 'Acks' failed on the 'required' tag
>   Key: 'Config.PyFuncPublisherConfig.Kafka.Brokers' Error:Field validation for 'Brokers' failed on the 'required' tag
>   Key: 'Config.PyFuncPublisherConfig.Kafka.Acks' Error:Field validation for 'Acks' failed on the 'required' tag
> ```

Need to merge this [PR](https://github.com/caraml-dev/helm-charts/pull/379) first

PR #493, opened 2023-11-20
Configure img builder resource API

tkpd-hafizhan:

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -474,24 +473,35 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
...
-				c.config.Resources.Limits.Memory,
-			),
-		},
-	}
+	if resource != nil {
+		// Get resourceRequirements from version endpoint configuration
+		resourceRequirements := RequestLimitResources{
+			Requests: Resource{
+				CPU:    resource.MustParse(resource.CPURequest),
+				Memory: resource.MustParse(resource.MemoryRequest),
```
leonlnj, line 481: what if user only wish to overwrite one of them? like providing mem override only

@ariefrahmansyah , this should be supported right?

api/queue/work/model_service_deployment.go
```diff
@@ -185,7 +185,7 @@ func (depl *ModelServiceDeployment) generateModelOptions(ctx context.Context, mo
 	modelOpt := &models.ModelOption{}
 	switch model.Type {
 	case models.ModelTypePyFunc:
-		imageRef, err := depl.ImageBuilder.BuildImage(ctx, model.Project, model, version)
+		imageRef, err := depl.ImageBuilder.BuildImage(ctx, model.Project, model, version, version.ImageBuilderResourceRequest)
```
leonlnj, line 188: the `buildimage` method isn't updated to pass the `ImageBuilderResourceRequest`.

however is this required? version is already passed

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -474,24 +473,48 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
...
-		Limit: Resource{
-			CPU: resource.MustParse(
-				c.config.Resources.Limits.CPU,
-			),
-			Memory: resource.MustParse(
-				c.config.Resources.Limits.Memory,
-			),
-		},
-	}
+	if resource != nil {
```
leonlnj, line 476: There are four values here CPU/Mem request+limit. The end behaviour is as long as user provide either CPU/Mem request override, it will be use, for the one not passed, it will use default.



We can keep the original block then check if `if resource != nil` (can we rename to imageBuilderResource or imageBuilderResourceRequest? resource is too vague)

for each field if the imageBuilderResourceRequest is not null, override the default. 

The last part to watch is if the request > default limits. In this case we can set limits==user requested resource. This part will fail at k8 creation if request > limit.

api/cmd/api/main.go
```diff
@@ -141,9 +141,9 @@ func main() {
 	registerQueueJob(dispatcher, dependencies.modelDeployment, dependencies.batchDeployment)
 	dispatcher.Start()
 
-	if err := initCronJob(dependencies, db); err != nil {
-		log.Panicf("Failed to initialize cron jobs: %s", err)
-	}
+	// if err := initCronJob(dependencies, db); err != nil {
+	// 	log.Panicf("Failed to initialize cron jobs: %s", err)
+	// }
```
ariefrahmansyah, line 146: Zan, can you help uncomment this again now that local dev/testing is done?

api/cmd/api/main.go
```diff
@@ -141,10 +141,6 @@ func main() {
 	registerQueueJob(dispatcher, dependencies.modelDeployment, dependencies.batchDeployment)
 	dispatcher.Start()
 
-	if err := initCronJob(dependencies, db); err != nil {
-		log.Panicf("Failed to initialize cron jobs: %s", err)
-	}
-
```
ariefrahmansyah, line 147: These lines should not be removed

api/config/environment.go
```diff
@@ -49,10 +49,11 @@ type EnvironmentConfig struct {
 
 	QueueResourcePercentage string `yaml:"queue_resource_percentage"`
 
-	DefaultPredictionJobConfig *PredictionJobResourceRequestConfig `yaml:"default_prediction_job_config"`
-	DefaultDeploymentConfig    *ResourceRequestConfig              `yaml:"default_deployment_config"`
-	DefaultTransformerConfig   *ResourceRequestConfig              `yaml:"default_transformer_config"`
-	K8sConfig                  *mlpcluster.K8sConfig               `validate:"required" yaml:"k8s_config"`
+	DefaultPredictionJobConfig  *PredictionJobResourceRequestConfig `yaml:"default_prediction_job_config"`
+	DefaultDeploymentConfig     *ResourceRequestConfig              `yaml:"default_deployment_config"`
+	DefaultImageBuilderResource *ResourceRequestConfig              `yaml:"default_image_builder_resource"`
```
ariefrahmansyah, line 54: This field is not being used anywhere. Should be safe to be removed.

python/sdk/client/api_client.py
```diff
@@ -50,7 +50,7 @@ class ApiClient(object):
     PRIMITIVE_TYPES = (float, bool, bytes, six.text_type) + six.integer_types
     NATIVE_TYPES_MAPPING = {
         'int': int,
-        'long': int if six.PY3 else int,  # noqa: F821
+        'long': int if six.PY3 else long,  # noqa: F821
```
ariefrahmansyah, line 53: This should be modified to int to fix lint error

python/sdk/merlin/model.py
```diff
@@ -1138,6 +1129,14 @@ def deploy(
                             )
                             break
 
+        if image_builder_resource_request is not None:
+            target_image_builder_resource_request = client.ResourceRequest(
+                0,
```
leonlnj, line 1134: can we just set client.ResourceRequest(cpu_request=...,memory_request=....) ?

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -474,22 +479,27 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
...
+			Memory: memoryRequest,
 		},
 		Limit: Resource{
-			CPU: resource.MustParse(
-				c.config.Resources.Limits.CPU,
-			),
-			Memory: resource.MustParse(
-				c.config.Resources.Limits.Memory,
-			),
+			CPU:    cpuRequest,
```
leonlnj, line 501: can we keep the previous platform behaviour?
```
cpuRequest := resource.MustParse(c.config.Resources.Requests.CPU)
	memoryRequest := 
cpuRequest := resource.MustParse(c.config.Resources.Requests.Memory)
	memoryRequest := 
cpuLimit := resource.MustParse(c.config.Resources.Limits.CPU)
	memoryRequest := 
MemoryLimit:= resource.MustParse(c.config.Resources.Limits.Memory)
```
only if `if resourceRequest != nil`, we set request == limit

api/config/config.go
```diff
@@ -213,7 +213,7 @@ type ImageBuilderConfig struct {
 	BuildTimeout                string           `validate:"required" default:"10m"`
 	KanikoImage                 string           `validate:"required" default:"gcr.io/kaniko-project/executor:v1.6.0"`
 	KanikoServiceAccount        string
-	Resources                   ResourceRequestsLimits `validate:"required"`
+	DefaultResource             ResourceRequestsLimits `validate:"required"`
```
ariefrahmansyah, line 216: nit: Should we rename it to `DefaultResources` instead of `DefaultResource`? Because the struct contains two fields: resource requests and limits.

api/pkg/imagebuilder/config.go
```diff
@@ -38,7 +38,7 @@ type Config struct {
 	// Kaniko kubernetes service account
 	KanikoServiceAccount string
 	// Kubernetes resource request and limits for kaniko
-	Resources cfg.ResourceRequestsLimits
+	DefaultResource cfg.ResourceRequestsLimits
```
leonlnj, line 41: @ariefrahmansyah  just to confirm, this can be freely changed (as seen in the e2e) because the helm will simply use the whole yaml block as payload -> https://github.com/caraml-dev/helm-charts/blob/main/charts/merlin/templates/_helpers.tpl#L274

ariefrahmansyah: Yes, correct.

ariefrahmansyah (approved):
Overall, LGTM with nit comments. Thanks @tkpd-hafizhan @leonlnj

leonlnj (approved):
lgtm, thanks zan and arief!

PR #494, opened 2023-11-21
feat: add data struct for inference schema

khorshuheng:
This PR provides the data types required to support ML observability within Merlin. The new data types are expected to be used for https://github.com/caraml-dev/merlin/pull/488 , flyte workflows, and also in the future when we store an optional inference schema within Merlin.

dataclasses_json dependency is introduced here as it is required if we want to pass dataclass as an input to Flyte.

Fixes #

```release-note
NONE
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/sdk/merlin/observability/inference.py
+++ b/python/sdk/merlin/observability/inference.py
@@ -1,0 +1,111 @@
+from dataclasses import dataclass
+from enum import unique, Enum
+from typing import Dict, Optional, List
+
+from dataclasses_json import dataclass_json
+
+
+@unique
+class ValueType(Enum):
+    FLOAT64 = 1
+    INT64 = 2
+    BOOLEAN = 3
+    STRING = 4
+
+
+@dataclass_json
+@dataclass
+class RegressionOutput:
+    prediction_score_column: str
+
+    @property
+    def column_types(self) -> Dict[str, ValueType]:
+        return {self.prediction_score_column: ValueType.FLOAT64}
+
+
+@dataclass_json
+@dataclass
+class BinaryClassificationOutput:
+    prediction_label_column: str
+    prediction_score_column: Optional[str] = None
+
+    @property
+    def column_types(self) -> Dict[str, ValueType]:
+        column_types_mapping = {self.prediction_label_column: ValueType.STRING}
+        if self.prediction_score_column is not None:
+            column_types_mapping[self.prediction_score_column] = ValueType.FLOAT64
+        return column_types_mapping
+
+
+@dataclass_json
+@dataclass
+class MulticlassClassificationOutput:
+    prediction_label_columns: List[str]
+    prediction_score_columns: Optional[List[str]] = None
+
+    @property
+    def column_types(self) -> Dict[str, ValueType]:
+        column_types_mapping = {
+            label_column: ValueType.STRING
+            for label_column in self.prediction_label_columns
+        }
+        if self.prediction_score_columns is not None:
+            for column_name in self.prediction_score_columns:
+                column_types_mapping[column_name] = ValueType.FLOAT64
+        return column_types_mapping
+
+
+@dataclass_json
+@dataclass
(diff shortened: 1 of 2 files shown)


krithika369 (approved):
LGTM. Left some small comments. Thanks, @khorshuheng !

python/sdk/setup.py
```diff
@@ -28,6 +28,7 @@
     "Click>=7.0,<8.1.4",
     "cloudpickle==2.0.0",  # used by mlflow
     "cookiecutter>=1.7.2",
+    "dataclasses-json>=0.6.0",
```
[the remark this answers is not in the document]
khorshuheng: I can downgrade this to 0.5.2

python/sdk/merlin/observability/inference.py
```diff
@@ -0,0 +1,111 @@
...
+class InferenceType(Enum):
+    BINARY_CLASSIFICATION = 1
+    MULTICLASS_CLASSIFICATION = 2
+    REGRESSION = 3
+    RANKING = 4
+
+
+@dataclass
+@dataclass_json
+class InferenceSchema:
```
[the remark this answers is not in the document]
khorshuheng: Originally, i plan to move ModelSpec from #488 to here as it is without any change. However, I realized that we might want to include the schema as part of the Merlin Model class attributes. If we do that, then it doesn't make sense to include model id and model version. I also renamed the class from ModelSpec to InferenceSchema.

python/sdk/merlin/observability/inference.py
```diff
@@ -0,0 +1,111 @@
...
+    @property
+    def column_types(self) -> Dict[str, ValueType]:
+        return {self.prediction_score_column: ValueType.FLOAT64}
+
+
+@dataclass_json
+@dataclass
+class BinaryClassificationOutput:
+    prediction_label_column: str
+    prediction_score_column: Optional[str] = None
```
tiopramayudi, line 30: can we make it mandatory and `prediction_label_column` is optional?

python/sdk/merlin/observability/inference.py
```diff
@@ -0,0 +1,111 @@
...
+        if self.prediction_score_column is not None:
+            column_types_mapping[self.prediction_score_column] = ValueType.FLOAT64
+        return column_types_mapping
+
+
+@dataclass_json
+@dataclass
+class MulticlassClassificationOutput:
+    prediction_label_columns: List[str]
+    prediction_score_columns: Optional[List[str]] = None
```
tiopramayudi, line 44: can we make it mandatory and `prediction_label_columns` is optional?

python/sdk/merlin/observability/inference.py
```diff
@@ -0,0 +1,111 @@
...
+            for column_name in self.prediction_score_columns:
+                column_types_mapping[column_name] = ValueType.FLOAT64
+        return column_types_mapping
+
+
+@dataclass_json
+@dataclass
+class RankingOutput:
+    rank_column: str
+    prediction_group_id_column: str
```
tiopramayudi, line 62: you can add new field `prediction_score_column` as optional field

tiopramayudi (commented):
should we add conversion to arize schema here?

python/sdk/merlin/observability/inference.py
khorshuheng: Since this is merged already, i will make the changes in #488 instead.

khorshuheng: As for conversion to Arize schema - i am not sure if we should add dependency to Arize on the Merlin SDK level. My preference would be to keep the dependency to Arize on the consumer / flyte workflow instead.

tiopramayudi: one drawback is easy to overlook if there is something change on the merlin part

PR #496, opened 2023-11-27
feat: add ground truth columns to prediction output

khorshuheng:

The current inference schema does not include columns for ground truth, which is required for batch workflow.

Fixes #

```release-note
NONE
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/observation-publisher/publisher/config.py
+++ b/python/observation-publisher/publisher/config.py
@@ -1,9 +1,8 @@
 from dataclasses import dataclass
-from enum import Enum, unique
+from enum import Enum, auto
 from typing import Optional
 
 from hydra.core.config_store import ConfigStore
-from merlin.observability.inference import InferenceSchema
 
 
 @dataclass
@@ -12,9 +11,8 @@ class ArizeConfig:
     space_key: str
 
 
-@unique
 class ObservabilityBackendType(Enum):
-    ARIZE = 1
+    ARIZE = "arize"
 
 
 @dataclass
@@ -22,10 +20,15 @@ class ObservabilityBackend:
     type: ObservabilityBackendType
     arize_config: Optional[ArizeConfig] = None
 
+    def __post_init__(self):
+        if self.type == ObservabilityBackendType.ARIZE:
+            assert (
+                self.arize_config is not None
+            ), "Arize config must be set for Arize observability backend"
 
-@unique
+
 class ObservationSource(Enum):
-    KAFKA = 1
+    KAFKA = "kafka"
 
 
 @dataclass
@@ -43,12 +46,18 @@ class ObservationSourceConfig:
     type: ObservationSource
     kafka_config: Optional[KafkaConsumerConfig] = None
 
+    def __post_init__(self):
+        if self.type == ObservationSource.KAFKA:
+            assert (
+                self.kafka_config is not None
+            ), "Kafka config must be set for Kafka observation source"
+
 
 @dataclass
 class Environment:
     model_id: str
     model_version: str
-    inference_schema: InferenceSchema
+    inference_schema: dict
     observability_backend: ObservabilityBackend
     observation_source: ObservationSourceConfig
 
--- a/python/observation-publisher/publisher/observability_backend.py
+++ b/python/observation-publisher/publisher/observability_backend.py
@@ -1,15 +1,19 @@
 import abc
-from typing import List
+from typing import Tuple
 
 import pandas as pd
 from arize.pandas.logger import Client
 from arize.pandas.logger import Schema as ArizeSchema
 from arize.utils.types import Environments
 from arize.utils.types import ModelTypes as ArizeModelType
-from merlin.observability.inference import InferenceSchema, InferenceType
+from merlin.observability.inference import (
+    InferenceSchema,
+    RegressionOutput,
+    BinaryClassificationOutput,
+    RankingOutput,
+)
 
-from publisher.config import (ArizeConfig, ObservabilityBackend,
-                              ObservabilityBackendType)
+from publisher.config import ArizeConfig, ObservabilityBackend, ObservabilityBackendType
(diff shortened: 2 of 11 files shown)


python/sdk/merlin/observability/inference.py
```diff
@@ -17,41 +17,59 @@ class ValueType(Enum):
...
         return column_types_mapping
 
 
 @dataclass_json
 @dataclass
 class MulticlassClassificationOutput:
     prediction_score_columns: List[str]
+    actual_score_columns: List[str]
     prediction_label_columns: Optional[List[str]] = None
+    actual_label_column: Optional[str] = None
```
krithika369, line 57: For multi-class classification, there will be multiple score columns but only one label per row, correct? If so, `actual_label_column: Optional[str]` makes sense to me. But  should` prediction_label_columns: Optional[List[str]] = None` be a single string instead of List?

python/observation-publisher/publisher/observability_backend.py
```diff
@@ -80,16 +31,111 @@ def __init__(
...
+            id_vars=id_vars,
+            value_vars=output_schema.prediction_score_columns,
+            var_name=prediction_tag_column,
+            value_name=prediction_score_column,
+        )
+        melted_df[prediction_tag_column] = melted_df[prediction_tag_column].apply(
+            output_schema.prediction_tag
+        )
+        melted_df[prediction_id_column] = melted_df.apply(
+            lambda row: f"{row['prediction_id']}_{row[prediction_tag_column]}",
```
tiopramayudi, line 80: how can you create this prediction_id_column during ground truth publishing?

python/sdk/merlin/observability/inference.py
```diff
@@ -1,111 +1,190 @@
...
+        regression: Regression model prediction output schema.
+        ranking: Ranking model prediction output schema.
+        prediction_id_column: Name of the column containing the prediction id, so that the ground truth can be joined
+            to the predictions.
+        tag_columns: List of column names containing tags for grouping the predictions.
+    """
+
     feature_types: Dict[str, ValueType]
     type: InferenceType
     binary_classification: Optional[BinaryClassificationOutput] = None
```
tiopramayudi, line 171: instead of defining all fields for each model type can we make abstract class for each model_type_output and have preprocess as one of abstract method? it will help us later on when we add new model type

python/observation-publisher/publisher/observability_backend.py
```diff
@@ -80,16 +31,111 @@ def __init__(
...
+    ) -> Tuple[pd.DataFrame, ArizeSchema]:
+        output_schema = self._inference_schema.regression
+        schema_attributes = self.common_arize_schema_attributes() | dict(
+            prediction_score_column_name=output_schema.prediction_score_column,
+        )
+        return df, ArizeSchema(**schema_attributes)
+
+    def _preprocess_ranking(self, df: pd.DataFrame) -> Tuple[pd.DataFrame, ArizeSchema]:
+        output_schema = self._inference_schema.ranking
+        schema_attributes = self.common_arize_schema_attributes() | dict(
```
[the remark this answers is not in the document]
khorshuheng: I was originally assuming that it's more common for models to return rank than prediction scores. We can certainly make prediction scores mandatory, and derive the ranking instead, but will there be situation where ranks are available but not prediction scores?

python/observation-publisher/publisher/observability_backend.py
tiopramayudi: the problem is most of our users doesn't have prediction scores stored in the BQ, they need to modify their data pipelines to support this. Also if they have all the required data `features` + `prediction_scores` + `ground_truth` they won't need streaming publishing , they can use offline publishing.

python/sdk/merlin/observability/inference.py
khorshuheng: One issue with using an abstract class for model_type_ouput, instead of a combination of inference type and fields for all model types, is that for the former, it is tricky to serialize to json then deserialize the json back to the concrete implementation of the abstract class. This will make it difficult to pass InferenceSchema / Model Type Output as a configuration to Flyte Workflows, or as a structure config (via OmegaConf library) in the observation publisher. We might have to implement a custom json serializer / deserializer for the class.

The other issue, is whether we want to expose preprocess method to the Merlin SDK users, consider that the method is intended to be used by the observation publishers (batch and online).

python/observation-publisher/publisher/observability_backend.py
tiopramayudi: I agree to drop support for multiclass for now, since we don't have usecase, also arize plan to revamp their multiclass implementation. Maybe we can support later

python/observation-publisher/publisher/observability_backend.py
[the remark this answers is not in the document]
tiopramayudi: > ranks are available but not prediction scores
honestly I don't know , maybe we can check first whether the rank column is not there if not check whether the prediction scores is available or not, if not available we will skip the publishing with error message?

krithika369 (commented):
@khorshuheng could you also capture a summary of the key changes in the PR description? Similar to https://github.com/caraml-dev/merlin/pull/492. This will also help document any important rationale behind some of the changes like why the `prediction_label_column` is now removed from the BinaryClassification schema or why PredictionLogResultsTable is now using float for all columns.

python/observation-publisher/publisher/config.py
```diff
@@ -12,20 +12,24 @@ class ArizeConfig:
     space_key: str
 
 
-@unique
 class ObservabilityBackendType(Enum):
```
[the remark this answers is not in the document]
khorshuheng: As much as i love StrEnum, it is only on Python 3.11. I do think we should use string instead of int, so that the value doesn't change when the order change.

khorshuheng:
Closing this in favor of https://github.com/caraml-dev/merlin/pull/504

PR #499, opened 2023-11-29
Use SDK version that matches PyFunc servers

krithika369:

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


python/batch-predictor/docker/app.Dockerfile
```diff
@@ -25,6 +25,6 @@ RUN if [[ ! -z "$GOOGLE_APPLICATION_CREDENTIALS" ]]; then gcloud auth activate-s
 RUN gsutil -m cp -r ${MODEL_URL} .
 RUN /bin/bash -c ". activate ${CONDA_ENVIRONMENT} && \
     sed -i 's/mlflow[><=]\{0,1\}.*$/mlflow==1.26.1/g' ${HOME}/model/conda.yaml && \
-    sed -i '/merlin-sdk/d' ${HOME}/model/conda.yaml && \
```
krithika369, line 28: Since this will not handle "nested" dependencies on Merlin SDK introduced by users' conda env, removing this.

ariefrahmansyah (approved):

python/batch-predictor/docker/app.Dockerfile
leonlnj: i think there are still benefit - if user specific a sdk version that is incompatible with mlflow

leonlnj (approved):

PR #500, opened 2023-11-30
Image Builder Improvements

ariefrahmansyah:
- refactor pyfunc base image

Image building is a mandatory process in the deployment of PyFunc-based Merlin components. The process abstracts the complexity of building and publishing a ready-to-be-used Docker image that contains the user's model source code, artifacts, and dependencies. However, there are some issues with the current image-building approach that lead to poor development experience and inefficient resource utilization. This PR addresses them by:

1. Refactoring Kaniko job specification generated by Merlin so that:
   a. it is more configurable -- we can set Kaniko's build arguments flags in the config and they will be propagated to the kaniko job spec
   b. it supports caching
2. Merlin API now generates a separate user's model dependencies files and upload it to GCS. This dependencies then will be installed separately during the image building (cause we use new Dockerfile)
   a. To achieve it, we introduce a new gsutil package to interface with GCS
3. Refactor base and application Dockerfiles used to build Docker images for pyfunc-server and batch prediction jobs
   a. Previously, the base Dockerfile had a step to install the pyfunc-server or batch prediction jobs dependencies. This step is now executed by the application Dockerfile
      i. Subsequently, this allows us to have only one base image instead of building base images for each Python version.
   b. New base Dockerfile will only install the platform components such as conda, spark, grpc library, but not the component itself
   c. New application Dockerfile will first download and install user's model dependencies; install component's dependencies; download model source code and artifacts; and finally dry-run the model

Improves image building performance

```release-note
NONE
```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


krithika369:
@ariefrahmansyah, not sure if this is WIP but could you please summarize the changes in the PR description? Like in https://github.com/caraml-dev/merlin/pull/483.

It will make it a lot easier to review. (Doesn't need to be elaborate, a summary of the motivation and the key changes will do.) Thanks!

api/pkg/imagebuilder/config.go
```diff
@@ -26,6 +26,8 @@ type Config struct {
 	// Path to sub folder which is intended to build instead of using root folder
 	ContextSubPath string
 	// Dockerfile Path within the build context
+	BaseImage cfg.BaseImageConfig
```
krithika369, line 29: Oh why do we have this config here and in `api/config.go` ? 🤔

krithika369 (commented):
@ariefrahmansyah I'm still reviewing the PR (will get back to it shortly). Publishing the few comments I had first. 🙏🏽

.github/workflows/merlin.yml
```diff
@@ -258,47 +260,46 @@ jobs:
           path: merlin.${{ needs.create-version.outputs.version }}.tar
           retention-days: ${{ env.ARTIFACT_RETENTION_DAYS }}
 
-  build-batch-predictor:
+  build-push-batch-predictor-base:
```
krithika369, line 263: What's the motivation for combining the build and publish steps? With this, we will be publishing the base images for every single commit to the PR / main, unlike previously, when the images are only pushed if the e2e succeeds. Shall we keep them separate as before?

.github/workflows/merlin.yml
```diff
@@ -439,11 +440,9 @@ jobs:
     uses: ./.github/workflows/release.yml
     needs:
       - create-version
-      - e2e-test
+      # - e2e-test
```
krithika369, line 443: Is this intentional?

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -129,6 +136,31 @@ func (c *imageBuilder) GetMainAppPath(version *models.Version) (string, error) {
...
+	condaEnv, err := c.gsutil.ReadFile(ctx, condaEnvUrl)
+	if err != nil {
+		return "", err
+	}
+
+	hash := sha256.New()
+	hash.Write([]byte(condaEnv))
+	hashEnv := hash.Sum(nil)
+
+	dependenciesUrl := fmt.Sprintf("gs://%s/merlin/dependencies/%x", artifactURL.Bucket, hashEnv)
```
leonlnj, line 156: `merlin/dependencies` sounds like it's pip freeze, `modeldependencies`?

api/pkg/gsutil/gsutil.go
```diff
@@ -0,0 +1,125 @@
...
+	"strings"
+
+	"cloud.google.com/go/storage"
+	"github.com/bboughton/gcp-helpers/gsurl"
+	"github.com/pkg/errors"
+	"google.golang.org/api/option"
+)
+
+// parseURL exists to allow us to use 'url' as a request param
+var parseURL = url.Parse
```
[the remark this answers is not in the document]
ariefrahmansyah: Because func (g *gsutil) ParseURL(url string) also uses `url` with `string` data type as input. If we try to directly use url.Parse() in ParseURL(), we will get this error: `url.Parse undefined (type string has no field or method Parse)` because `url` here is the input variable `url`, not `net/url` package.

Other options are:

1. use alias import:

```
import (
    gourl "net/url"
)

...

func (g *gsutil) ParseURL(url string) (*URL, error) {
	u, err := gourl.Parse(url)
```

Or, 2. rename the input variable:

```
func (g *gsutil) ParseURL(gsURL string) (*URL, error) {
	u, err := url.Parse(gsURL)
```

I'll go with second option.

api/pkg/gsutil/gsutil.go
```diff
@@ -0,0 +1,125 @@
...
+	// Object is the name and or path of the object stored in the bucket. It
+	// should not start with a foward slash.
+	Object string
+}
+
+// Parse parses a Google Cloud Storage string into a URL struct. The expected
+// format of the string is gs://[bucket-name]/[object-path]. If the provided
+// URL is formatted incorrectly an error will be returned.
+func (g *gsutil) ParseURL(url string) (*URL, error) {
+	u, err := parseURL(url)
```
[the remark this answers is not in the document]
ariefrahmansyah: Discussed it here: https://github.com/caraml-dev/merlin/pull/500#discussion_r1418450859

api/pkg/imagebuilder/config.go
ariefrahmansyah: In cmd/setup.go we transform the config.ImageBuilderConfig to pkg.imagebuilder.Config: https://github.com/caraml-dev/merlin/blob/main/api/cmd/api/setup.go#L101-L141

When initializing imageBuilder object in pkg.imagebuilder (e.g. [NewModelServiceImageBuilder](https://github.com/caraml-dev/merlin/blob/main/api/pkg/imagebuilder/model_service_imagebuilder.go#L27)), we should be able to initialize it using its own internally-available Config struct, not external config.

python/batch-predictor/docker/app.Dockerfile
```diff
@@ -13,19 +13,32 @@
...
+    else echo "GOOGLE_APPLICATION_CREDENTIALS is not set. Skipping gcloud auth command." > gcloud-auth.txt; \
+    fi
+
+# Download and install user model dependencies
+ARG MODEL_DEPENDENCIES_URL
+RUN gsutil cp ${MODEL_DEPENDENCIES_URL} conda.yaml
+RUN conda env create --name merlin-model --file conda.yaml
+
+# Copy and install batch predictor dependencies
+COPY --chown=${UID}:${GID} batch-predictor ${HOME}/merlin-spark-app
```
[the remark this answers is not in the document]
ariefrahmansyah: 1. Will this command reuse the cache if the Kaniko build context doesn't change (i.e., 2 image building jobs performed within the same release using the same user conda env file)?
Yes. It will reuse the cache if no change in the build context.

2. If the build context changes (eg: 0.36.0 -> 0.37.0), is it command sure to get the new code (0.37.0) instead of reusing cache?
I haven't tested it, but in theory, it should get the new code and invalidate the cache. Will get back to you once I tested it.

.github/workflows/merlin.yml
ariefrahmansyah: -> What's the motivation for combining the build and publish steps?
No strong opinion here. It was to fasten my development iteration. I'll separate it again.

ariefrahmansyah:
-> Why is the python/sdk/test/pyfunc/model_2.joblib file removed in this PR? Isn't it needed by the integration tests?

When running pyfunc_integration_test.py from local machine, it will re-train the model hence updating the model. I deleted it  but forgot to restore the old one. I've restored it know.

api/pkg/imagebuilder/imagebuilder.go
ariefrahmansyah: updated to model_dependencies

api/pkg/gsutil/gsutil.go
```diff
@@ -0,0 +1,125 @@
...
+	bytes, err := io.ReadAll(reader)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read the content of %s", url)
+	}
+	return bytes, nil
+}
+
+// URL contains the information needed to identify the location of an object
+// located in Google Cloud Storage.
+type URL struct {
```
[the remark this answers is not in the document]
ariefrahmansyah: Updated.

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -422,25 +461,23 @@ func (c *imageBuilder) createKanikoJobSpec(project mlp.Project, model *models.Mo
...
-	if !ok {
-		return nil, fmt.Errorf("No matching base image for tag %s", version.PythonVersion)
-	}
+	baseImageTag := c.config.BaseImage
 
 	kanikoArgs := []string{
 		fmt.Sprintf("--dockerfile=%s", baseImageTag.DockerfilePath),
 		fmt.Sprintf("--context=%s", baseImageTag.BuildContextURI),
-		fmt.Sprintf("--build-arg=MODEL_URL=%s/model", version.ArtifactURI),
 		fmt.Sprintf("--build-arg=BASE_IMAGE=%s", baseImageTag.ImageName),
```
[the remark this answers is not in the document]
ariefrahmansyah: Good idea. Updated!

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -145,6 +177,13 @@ func (c *imageBuilder) BuildImage(ctx context.Context, project mlp.Project, mode
 	startTime := time.Now()
 	result := "failed"
 
+	// TODO(arief): create or get model dependency
```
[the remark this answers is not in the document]
ariefrahmansyah: -> we just need to check for the hashed file, whether it exists and create it if not, right?

The current implementation does not check whether the hashed file exists in GCS or not. It naively overwrites the hashed dependencies URL.

The comparison is:

Option 1: With checking will have at least 2 read operation + optional 1 write operation

1. Read user's raw model dependencies from GCS
2. Check hashed model dependencies URL exists in GCS or not
3. If exists, return hashed model dependencies URL
4. If not exists, write to hashed model dependencies URL in GCS

Option 2: Without checking will always have 1 read operation + 1 write operation

1. Read user's raw model dependencies from GCS
2. Write to hashed model dependencies URL in GCS

Option 2 is simpler but relatively more expensive (read operation in GCS is cheaper than write). I'm putting TODO to update it to Option 1 as originally discussed.

api/config/config.go
```diff
@@ -203,15 +204,18 @@ type ImageBuilderConfig struct {
 	GcpProject                  string
 	ContextSubPath              string
 	DockerfilePath              string           `validate:"required" default:"./Dockerfile"`
+	BaseImage                   BaseImageConfig  `validate:"required"`
```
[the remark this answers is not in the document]
ariefrahmansyah: Yup, I've removed it.

api/pkg/imagebuilder/imagebuilder.go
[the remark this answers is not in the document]
ariefrahmansyah: Yup, updated to Option 1: https://github.com/caraml-dev/merlin/blob/image-builder-improvements/api/pkg/imagebuilder/imagebuilder.go#L139

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -100,35 +104,73 @@ const (
...
 		kubeClient:    kubeClient,
 		config:        config,
 		nameGenerator: nameGenerator,
+		gsutil:        gsutil,
 	}
 }
 
 // GetMainAppPath Returns the path to run the main.py of batch predictor, as configured via env var
 func (c *imageBuilder) GetMainAppPath(version *models.Version) (string, error) {
-	baseImageTag, ok := c.config.BaseImages[version.PythonVersion]
```
krithika369, line 120: I don't think we should remove this check entirely. Can we add a `SupportedPythonVersions []string` field under `ImageBuilderConfig` and use that to check? Ultimately, the pyfunc real time and batch server have some Python version constraints as does the SDK ([example](https://github.com/caraml-dev/merlin/blob/main/python/pyfunc-server/setup.py#L48)). We should capture this in the config and check here to prevent unnecessarily triggering image building for unsupported Python versions.

api/config/config.go
```diff
@@ -202,17 +203,20 @@ type ClusterConfig struct {
 type ImageBuilderConfig struct {
 	ClusterName                 string `validate:"required"`
 	GcpProject                  string
+	GcsAuthenticationType       string
+	GcsCredentialsJson          string
```
krithika369, line 207: How is this value expected to be supplied during deployment time? If using Workload Identity, do we really need this?

Btw I just remembered we have this implementation, also used by Merlin as `MlflowDeleteService`, for model version delete:
* https://github.com/caraml-dev/mlp/blob/main/api/pkg/artifact/artifact.go#L7
* https://github.com/caraml-dev/mlp/blob/main/api/pkg/client/mlflow/mlflow.go

Can we make use of this instead of `gsutil.go` ? We will need to add the `ReadArtifact` and `WriteArtifact` APIs to the service [here](https://github.com/caraml-dev/mlp/blob/main/api/pkg/artifact/artifact.go#L11).

python/pyfunc-server/docker/Dockerfile
```diff
@@ -15,19 +15,29 @@
...
-    conda env update --name merlin-model --file model/conda.yaml && \
-    pip install ${SDK_PATH} && \
-    python -m pyfuncserver --model_dir model --dry_run"
+# Download and dry-run user model artifacts and code
+ARG MODEL_ARTIFACTS_URL
+RUN gsutil -m cp -r ${MODEL_ARTIFACTS_URL} .
+RUN /bin/bash -c ". activate merlin-model && python -m pyfuncserver --model_dir model --dry_run"
 
-CMD ["/bin/bash", "./run.sh"]
+CMD ["/bin/bash", "/pyfunc-server/run.sh"]
```
krithika369, line 43: Since set `WORKDIR /pyfunc-server` above, is this change really needed?

python/batch-predictor/docker/app.Dockerfile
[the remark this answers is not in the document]
ariefrahmansyah: Just tested it and the hypothesis is correct. The new code invalidates the cache.

api/pkg/imagebuilder/imagebuilder.go
ariefrahmansyah: Oops, I forgot about SupportedPythonVersions. I've updated it

api/config/config.go
ariefrahmansyah: I'm not aware of mlp.artifact package. Prior to writing this new gsutil package, I only see this gsutil being used by standard transformer -- which we should refactor too later. Turned out, this mlp.artifact is indirect dependencies of mlflow service.
I've updated the mlp.artifact package (thanks for the quick review there) and update merlin implementation to use it.

python/pyfunc-server/docker/Dockerfile
ariefrahmansyah: It's not needed. During testing, I was playing around with and without workdir. And even though we have workdir there, I'm keeping /pyfunc-server in path to make it explicit.

.github/workflows/merlin.yml
ariefrahmansyah: Uncommented it.

api/config/config.go
```diff
@@ -99,8 +100,10 @@ type BaseImageConfig struct {
 	DockerfilePath string `validate:"required" json:"dockerfilePath"`
 	// GCS URL Containing build context
 	BuildContextURI string `validate:"required" json:"buildContextURI"`
+	// Path to sub folder which is intended to build instead of using root folder
+	BuildContextSubPath string `json:"buildContextSubPath"`
 	// path to main file to run application
-	MainAppPath string `validate:"required" json:"mainAppPath"`
+	MainAppPath string `json:"mainAppPath"`
```
krithika369, line 106: Why has the `validate:"required"` part been removed now? We're not somehow setting the default value elsewhere, so my understanding is we still need it. 🤔

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -100,35 +104,91 @@ const (
...
+		kubeClient:      kubeClient,
+		config:          config,
+		nameGenerator:   nameGenerator,
+		artifactService: artifactService,
 	}
 }
 
+func (c *imageBuilder) validatePythonVersion(version *models.Version) error {
+	if version.PythonVersion == "" {
+		return fmt.Errorf("python version is not set model version %s", version.ID)
```
krithika369, line 131: Nit: `python version is not set for model version %s` ? The "for" is missing.

krithika369 (approved):
Left a couple of small comments. The rest LGTM. Thanks for this PR and the CI steps refactoring, @ariefrahmansyah !

PR #501, opened 2023-12-07
Add imagebuilder override and stackdriver url

leonlnj:
This PR adds two feature - stackdriver panel and imagebuilder override option.

For image building failure, the logs are unavailable, this PR adds a naive approach to show the stackdriver url for image builder. Also the stackdriver url link was broken for image builder. Fixing the url alone, doesn't solve the problem when the active image builder container died and no logs are available hence adding of a persistent panel.

There are no smart ways to tell when to show the panel (failure due to deployment or image building) hence it is shown for all case 

The imagebuilder ui is a follow up from https://github.com/caraml-dev/merlin/pull/493

### Stackdriver panel
- stackdriver url hidden for active container that are image building
- fix generated stackdriver url for imagebuilder
- required image builder information, passed by the merlin api from imagebuilder config

This PR adds a rather naive approach to show the imagebuilder url at all phase. 

When there is no active container 
image

Batch

Real time
image

Logs query - updated to query from start time == model.updated_time
image

### Imagebuilder override
Real time 
image

Reflected on builder
image

Inaccessibility to logs for image builder during failure

Fixes #

```release-note
NONE
```

**Checklist**

- [X] Tested locally
- [ ] Updated documentation


api/cmd/api/ui_handler.go
```diff
@@ -22,6 +22,10 @@ type uiEnvHandler struct {
 	AlertEnabled bool `json:"REACT_APP_ALERT_ENABLED"`
 
 	ModelDeletionEnabled bool `json:"REACT_APP_MODEL_DELETION_ENABLED"`
+
+	ImageBuilderCluster   string `json:"REACT_APP_IMAGE_BUILDER_CLUSTER"`
+	ImageBuilderProject   string `json:"REACT_APP_IMAGE_BUILDER_PROJECT"`
+	ImageBuilderNamespace string `json:"REACT_APP_IMAGE_BUILDER_NAMESPACE"`
```
leonlnj, line 28: im not too sure to use an `ImageBuilderConfig` instead of string , it has way too many config than required

ui/src/components/logs/ContainerLogsView.js
```diff
@@ -121,45 +126,57 @@ export const ContainerLogsView = ({
...
-          const newLogUrl = config.MERLIN_API + "/logs?" + logParams;
-          if (newLogUrl !== logUrl) {
-            setLogUrl(newLogUrl);
-          }
+      if (projectLoaded) {
+        // set image builder url
+        let stackdriverQuery = {
+          job_name: project.name + "-" + model.name + "-" + versionId
+        };
+        setImagebuilderStackdriverUrl(createStackdriverUrl(stackdriverQuery, "image_builder"));
```
leonlnj, line 138: this will always be created, even when there are no active components. 
The query combined with the image builder information by config, will allow this to work when there are no active components

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -485,18 +485,16 @@ func (c *imageBuilder) createKanikoJobSpec(
...
-	// User defined resource request and limit will override the default value
+	// User defined resource request and limit will override the default value and set request==limit
 	if resourceRequest != nil {
 		if !resourceRequest.CPURequest.IsZero() {
 			cpuRequest = resourceRequest.CPURequest
+			cpuLimit = resourceRequest.CPURequest
 		}
 		if !resourceRequest.MemoryRequest.IsZero() {
 			memoryRequest = resourceRequest.MemoryRequest
+			memoryLimit = resourceRequest.MemoryRequest
```
ariefrahmansyah, line 496: Wow. Thanks for fixing this!

ui/src/components/logs/ContainerLogsView.js
```diff
@@ -175,12 +192,29 @@ export const ContainerLogsView = ({
...
+          >
+          <EuiDescriptionListTitle style={{ width: "30%", fontSize: '14px'}} >Stackdriver Logs</EuiDescriptionListTitle>
+          <EuiDescriptionListDescription>
+            <EuiLink href={imagebuilderStackdriverUrl} target="_blank" external>Image Builder</EuiLink>
+          </EuiDescriptionListDescription>
+          </EuiDescriptionList>
+        </EuiCard>
+        )
+      }
+      <EuiSpacer size="s" />
```
ariefrahmansyah, line 211: This EuiSpacer should be inside the conditional block above:

```
{
        imagebuilderStackdriverUrl && (
          <>
            <EuiCard>...</EuiCard>
            <EuiSpacer size="s" />
          </>
        )
}
```

leonlnj: i knew but didnt want to add a `<Fragment>` for it so i left it like this 🤣 , added the fragment and spacer anyway, thanks!

ui/src/pages/version/components/forms/steps/ModelStep.js
```diff
@@ -12,12 +12,22 @@ import { DeploymentConfigPanel } from "../components/DeploymentConfigPanel";
...
 import { ResourcesPanel } from "../components/ResourcesPanel";
+import { ImageBuilderSection } from "../components/ImageBuilderSection";
 
 export const ModelStep = ({ version, isEnvironmentDisabled = false }) => {
   const { data, onChangeHandler } = useContext(FormContext);
   const { onChange } = useOnChangeHandler(onChangeHandler);
   const { errors } = useContext(FormValidationContext);
 
+  // isEnvironmentDisabled is true for redeployment and false for deploy 
+  const imagebuilderSection = !isEnvironmentDisabled ? (
```
krithika369, line 23: Hmmm.. If the first/previous deployment attempt failed, we will still go through the "redeployment" flow. In this case, `isEnvironmentDisabled` will be true but we may still want to show the image builder section to let the user tweak those values, in case that's where the prior deployment attempt failed.

So I'm thinking we can just show this section always.

ui/src/pages/version/components/forms/components/ImageBuilderSection.js
```diff
@@ -0,0 +1,71 @@
...
+  }) => {
+
+    const { onChange } = useOnChangeHandler(onChangeHandler);
+
+    return (
+    <EuiAccordion 
+    id="imagebuilderOption"
+    buttonContent="Advanced configurations">
+    <EuiSpacer size="s" />
+    <EuiFlexGroup direction="row">
```
krithika369, line 25: Nit: I think up until this flex group (the "Advanced configurations" accordion) can just move to `ui/src/pages/version/components/forms/steps/ModelStep.js` and `ui/src/job/form/JobFormOthers.js`, respectively. That should be a part of the parent component while `ImageBuilderSection` only deals with the image builder related components.

krithika369 (approved):
Left some comments, a number of them minor. Besides these the PR LGTM. Thanks for this, @leonlnj ! Please feel free to merge once you've reviewed / addressed the comments.

ui/src/pages/version/components/forms/validation/schema.js
```diff
@@ -3,6 +3,9 @@ import { appConfig } from "../../../../../config";
 
 const cpuRequestRegex = /^(\d{1,3}(\.\d{1,3})?)$|^(\d{2,5}m)$/,
   memRequestRegex = /^\d+(Ei?|Pi?|Ti?|Gi?|Mi?|Ki?)?$/,
+  // these are similar to the cpu request, just that it allows empty fields
+  imageBuilderCpuRequestRegex = /^(\d{1,3}(\.\d{1,3})?)?$|^(\d{2,5}m)$/,
```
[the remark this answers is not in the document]
leonlnj: was trying out some of such but missed this, exactly what fit the case, thanks!

ui/src/pages/version/components/forms/steps/ModelStep.js
leonlnj: ` we will still go through the "redeployment" flow` oh right, i was actually under the impression user have to create multiple new version to tweak this config. Just occur to me that they can call deploy multiple times, as long as the image isn't built, it will go through this flow

ui/src/job/form/context.js
```diff
@@ -209,6 +210,26 @@ export const JobFormContextProvider = ({ job: initJob, ...props }) => {
     },
     [setJob]
   );
+  
+  // This port partial of the FormContextProvider that most of the forms are using,
+  // so that the similiar/future section using onChangeHandler can be reused and 
+  // does not requires adding new callbacks per field
+  var handleChanges = React.useCallback(function (paths, value) {
```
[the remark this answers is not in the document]
leonlnj: i totally get that why we -should- do this, just that I was concern over refactoring the other 10 callbacks created in this file as a result of not using the `FormContextProvider` and this was my half attempt at using the `FormContextProvider` implementation just for this new field instead of adding another callback 😅 

i'll take a stab at this tomorrow or in a separate PR.

ui/src/components/logs/ContainerLogsView.js
```diff
@@ -174,13 +187,35 @@ export const ContainerLogsView = ({
...
+            <EuiSpacer size="s"/>
+            <EuiPanel>
+              <EuiFlexGroup direction="row" alignItems="center">
+                <EuiFlexItem style={{marginTop:0, marginBottom:0}} grow={false}>
+                  <EuiText  style={{ fontSize: '14px', fontWeight:"bold"}}>Stackdriver Logs</EuiText>
+                </EuiFlexItem>
+                {Object.entries(stackdriverUrls).map(([component,url])=> (
+                  <EuiFlexItem style={{marginTop:0, marginBottom:0, paddingLeft:"10px"}} key={component} grow={false}>
+                    <EuiText size="xs" >
+                      <EuiLink href={url} target="_blank" external>{component.replace(new RegExp("_", "g"), " ")}</EuiLink>
```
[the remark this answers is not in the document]
leonlnj: done 
image

ui/src/job/form/context.js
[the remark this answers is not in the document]
krithika369: That's fair. Perhaps we can wrap the `JobForm` with the `FormContextProvider` (instead of defining the generic callback handlers here) but only take advantage of it for the new fields first and add a TODO in this file for refactoring the existing ones.

Anyway, if this proves to be troublesome, we can just add the TODO, keep things as is and revisit this later too.

PR #502, opened 2023-12-08
feat: Support model observability to pyfunc model

tiopramayudi:

We introduce `PyFuncV3Model` as new pyfunc class that support model observability. It is working fine for new model but it is not the case for existing model that use `PyFuncModel` since merlin model can't be changed once it is logged or deployed. Implication of this, users need to create new model that use `pyfunc_v3` type and the endpoint will be different hence there will be modification on the upstream caller. This PR will allow `PyFuncModel` to user model observability as long as the output of the model is `PyFuncOutput`.

This PR also allow `features`, `entities`, and `prediction` in `PyFuncOutput` to be pandas DataFrame, previously the type is strictly `Values` where in reality most of our users use pandas DataFrame and numpy ndarray

Modification:
* api/api/validator.go -> Relax the validation to allow PyFuncModel also
* api/cluster/resource/templater.go -> Supply necessary env vars for PyFunc and PyFuncV3 if model observability enabled
* python/sdk/merlin/pyfunc.py -> Supporting used of pandas DataFrame and numpy ndarray for model input and output

Fixes #

```release-note

```

**Checklist**

- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/pyfunc-server/test/test_http.py
+++ b/python/pyfunc-server/test/test_http.py
@@ -1,4 +1,5 @@
 import json
+import pytest
 import os
 import pathlib
 import re
@@ -6,11 +7,13 @@ import shutil
 import signal
 import subprocess
 import random
+import pandas as pd
+import numpy as np
 
 import mlflow
 import requests
 from merlin.model import PyFuncModel, PyFuncV3Model
-from merlin.pyfunc import ModelInput, ModelOutput, Values
+from merlin.pyfunc import ModelInput, ModelOutput, Values, PyFuncOutput
 from prometheus_client.metrics import Counter, Gauge
 
 from pyfuncserver.config import HTTP_PORT, WORKERS, MODEL_NAME, MODEL_VERSION, MODEL_FULL_NAME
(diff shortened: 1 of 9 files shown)


khorshuheng (approved):
lgtm!

krithika369:
@tiopramayudi @khorshuheng as discussed internally, let's explore moving the model type to the model version instead to solve the problem:

> for existing model that use PyFuncModel since merlin model can't be changed once it is logged or deployed

python/pyfunc-server/test/test_http.py
```diff
@@ -59,45 +95,32 @@ def infer(self, model_input: ModelInput) -> ModelOutput:
...
-            ),
-            prediction_ids = model_input.prediction_ids
-        )
+        return sample_model_output_df
     def postprocess(self, model_output: ModelOutput, request: dict) -> dict:
         return model_output.predictions_dict()
 
+
+class SamplePyFuncWithModelObservability(PyFuncModel):
+    def infer(self, request, **kwargs) -> ModelOutput:
```
krithika369, line 117: I think we should also update the signature of the `infer` method for the `PyFuncModel` to better reflect this.

```python
def infer(self, model_input: Union[dict, ModelInput]) -> Union[dict, ModelOutput]:
```

This will also take us a step closer to unifying the 2 PyFunc types later on. Wdyt?

krithika369 (approved):
Left one comment. The rest LGTM. Thanks, @tiopramayudi !

python/pyfunc-server/test/test_http.py
tiopramayudi: For now I'll leave it as it is, will modify it in the another PR to unifying the pyfuncv3 and pyfunc

PR #504, opened 2023-12-12
feat: add ground truth columns to prediction output

khorshuheng:

Changes introduced:
- Using string enum type as opposed to int, so that the json output for the enum field will be of the string value as opposed to int, which would make it easier to understand the json content. This is important for the use case of providing the enum values as input to flyte tasks.
- Instead of having all possible classes as separate attributes within the inference schema class, they are now combined within a single field, model_prediction_output. In addition, the prediction output is now expected to implement the preprocess method, which will be used to preprocess the dataframe before sending the result to Arize.
- Due to the existence of models that might have non-float prediction value (such as ranking model), the subclasses of PredictionOutput are expected to provide prediction types as well.
- Changes required for observation publisher and batch publisher, due to the changes made to the Merlin SDK, will be on a separate pull request.

Fixes #

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation

### DIFF
--- a/python/sdk/merlin/observability/inference.py
+++ b/python/sdk/merlin/observability/inference.py
@@ -1,111 +1,252 @@
-from dataclasses import dataclass
-from enum import unique, Enum
-from typing import Dict, Optional, List
+import abc
+from dataclasses import dataclass, field
+from enum import Enum, unique
+from typing import Dict, List, Optional
 
-from dataclasses_json import dataclass_json
+import numpy as np
+import pandas as pd
+from dataclasses_json import DataClassJsonMixin, config, dataclass_json
+
+
+class ObservationType(Enum):
+    """
+    Supported observation types.
+    """
+
+    FEATURE = "feature"
+    PREDICTION = "prediction"
+    GROUND_TRUTH = "ground_truth"
 
 
 @unique
 class ValueType(Enum):
-    FLOAT64 = 1
-    INT64 = 2
-    BOOLEAN = 3
-    STRING = 4
+    """
+    Supported feature value types.
+    """
+
+    FLOAT64 = "float64"
+    INT64 = "int64"
+    BOOLEAN = "boolean"
+    STRING = "string"
 
 
-@dataclass_json
-@dataclass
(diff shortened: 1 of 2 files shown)


python/sdk/merlin/observability/inference.py
```diff
@@ -1,111 +1,251 @@
...
-    rank_column: str
+class RankingOutput(PredictionOutput):
+    rank_score_column: str
     prediction_group_id_column: str
+    relevance_score_column: str
+    """
+    Ranking model prediction output schema.
+    Attributes:
+        rank_score_column: Name of the column containing the ranking score of the prediction.
+        prediction_group_id_column: Name of the column containing the prediction group id.
```
krithika369, line 212: Could you also document the `relevance_score_column` attribute?

python/sdk/merlin/observability/inference.py
```diff
@@ -1,111 +1,251 @@
...
-    def column_types(self) -> Dict[str, ValueType]:
-        column_types_mapping = {self.prediction_score_column: ValueType.FLOAT64}
-        if self.prediction_label_column is not None:
-            column_types_mapping[self.prediction_label_column] = ValueType.STRING
-        return column_types_mapping
-
-
-@dataclass_json
-@dataclass
-class MulticlassClassificationOutput:
```
krithika369, line 42: Why has `MulticlassClassificationOutput` been removed in this PR? Do we plan to address it later on only when there is a need?

krithika369 (approved):
Left some small comments. LGTM. Thanks for this refactoring, @khorshuheng !

python/sdk/merlin/observability/inference.py
khorshuheng: Multiclass model takes significantly more effort to implement compared to the other models, in terms of the preprocessing required for the input dataframe. And this effort is complicated by the fact that we do not really know how our users specify the input for multiclass models, both on online prediction and batch training data scenario. Therefore, we should address this later, when we have the clarity.

PR #505, opened 2023-12-13
Fix: image build process on pyfunc v3 model

tiopramayudi:

In v0.37.0 the image building is triggered when deploying pyfunc_v3, but in the recent release the image building process is skip for pyfunc_v3 hence the deployment will fail

Trigger image building process for pyfunc_v3 deployment

Fixes #

```release-note

```

**Checklist**

- [ ] Tested locally
- [ ] Updated documentation


ariefrahmansyah (approved):
Thanks for fixing this Tio! 🙏🏻

PR #509, opened 2023-12-15
feat: update observation publisher to use newer sdk

khorshuheng:

This is a follow up to https://github.com/caraml-dev/merlin/pull/504 , which update the observation publisher to use the latest Merlin SDK. Also introduce the use of piptools to pin the dependencies.

Another important changes is using dict instead of the InferenceSchema dataclass. This is because OmegaConf (and by extension, Hydra) does not have a way to instantiate the correct subclass to an abstract class field, so we have to perform our own deserialization instead.

Fixes #

NONE

```release-note
NONE
```

**Checklist**

- [x] Tested locally
- [x] Updated documentation


python/observation-publisher/Makefile
```diff
@@ -4,6 +4,12 @@ ENVIRONMENT_CONFIG = "example-override"
 setup:
 	@echo "Setting up environment..."
 	@pip install -r requirements.txt --use-pep517
```
tiopramayudi, line 6: do we still need this?

python/observation-publisher/Makefile
```diff
@@ -4,6 +4,12 @@ ENVIRONMENT_CONFIG = "example-override"
...
+	@pip install -r requirements-dev.txt
+
+.PHONY: pip-compile
+pip-compile:
+	@echo "Compiling requirements..."
+	@python -m piptools compile
```
tiopramayudi, line 12: just curious, what is the purpose of this compiling action?

python/observation-publisher/publisher/__main__.py
```diff
@@ -12,11 +13,18 @@ def start_consumer(cfg: PublisherConfig) -> None:
     if missing_keys:
         raise RuntimeError(f"Got missing keys in config:\n{missing_keys}")
     prediction_log_consumer = new_consumer(cfg.environment.observation_source)
+    inference_schema = InferenceSchema.from_dict(
```
tiopramayudi, line 16: where this `from_dict` method is declared? is it part of dataclass method?

python/observation-publisher/tests/test_observability_backend.py
```diff
@@ -1,70 +1,155 @@
...
+    try:
+        assert len(errors) == 0
+    except AssertionError:
+        for error in errors:
+            print(error.error_message())
+        raise
 
-def test_arize_schema_mapping():
+
+def assert_arize_schema_validity(
```
tiopramayudi, line 33: should we use [no op arize client](https://github.com/Arize-ai/client_python/blob/main/tests/pandas/test_pandas_logger.py#L55C1-L55C1) to check whole log functionality?

python/observation-publisher/Makefile
khorshuheng: `--use-pep517` is used mainly to suppress the warning.

python/observation-publisher/Makefile
khorshuheng: Let say we are writing our own requirements.txt, and pin all the dependencies we require on specific versions. Unless we pin the transitive dependencies as well, the dependencies installed may differ every time we build the image / install the dependencies.

We could do a pip freeze and capture all the dependencies, but it would be harder to update the dependencies at a later time. You also won't have good visibility on which dependencies are core dependencies, and which are transitive dependencies.

What piptools does is that, it takes in an input file, where the dependencies version may or may not be fixed, then generate a requirements.txt that has the exact dependency version, along with comments to illustrate the dependency hierarchy.  One advantage of doing so, is that, let say you update a dependency, such as google-cloud-bigquery, you can run pip-compile to know what are the dependencies that could be upgraded / are incompatible.

This approach is certainly not suitable for libraries, but for standalone application, it's a fairly common practice.

python/observation-publisher/publisher/__main__.py
khorshuheng: This comes from the `@dataclass_json` annotation. The decorator modifies the original class, add additional methods, then register the original class as a virtual subclass to DataclassJsonMixIn.

python/observation-publisher/tests/test_observability_backend.py
khorshuheng: Yup, that's a good idea.

krithika369 (approved):
LGTM. Left one small comment. Thanks for the refactoring, @khorshuheng !

python/observation-publisher/Dockerfile
```diff
@@ -1,14 +1,12 @@
 FROM python:3.10
 
 WORKDIR /mlobs
-
+COPY sdk ./sdk
+WORKDIR /mlobs/observation-publisher
```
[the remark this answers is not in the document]
khorshuheng: The reason why we are copying the requirements.txt, is to take advantage of docker caching. If requirements.txt didn't change, there is no need to reinstall the dependencies. In the simplified step above, regardless of whether requirement changes, we still need to spend time installing the dependencies.

As for why we remove the requirement file, the intention is to make sure that only necessary files required to run the application is on docker. Although, there's no issue if we keep it either.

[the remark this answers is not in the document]
krithika369: Got it. May be helpful to have an inline comment in the Dockerfile about copying the requirements first and the caching.

PR #511, opened 2023-12-22
fix: Set grpc as the container name when raw deployment mode is used

krithika369:
# Description

UPI models (using gRPC) were unable to deploy using Raw deployment mode. It is related to this issue: https://github.com/kserve/kserve/issues/2728

The fix is to set the container name as "grpc" rather than "h2c" as the latter is only applicable for Serverless deployments ("h2c" also works for Serverless deployments using HPA based autoscaling metrics).

# Modifications

- `api/cluster/resource/templater.go` - Make `createContainerPorts` take the deployment mode into account

# Tests

- [ ] Deploying new PyFunc models using UPI + Raw Deployment

# Checklist
- [x] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
Fix UPI models unable to deploy using raw deployment mode
```


ariefrahmansyah (approved):

tiopramayudi (approved):

tiopramayudi:
LGTM, Thanks @krithika369

PR #513, opened 2024-01-02
feat: Add support for multiple sinks

khorshuheng:
# Description

Allow the observation publisher to publish to multiple sinks. Supported sinks are: Arize, BigQuery

# Modifications

- configuration format has been modified to support multiple sinks
- update python requirement
- fix mypy linter errors
- Bigquery sink implementation

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```


python/observation-publisher/publisher/observation_sink.py
```diff
@@ -0,0 +1,223 @@
...
+            print(f"Failed to log to Arize: {error_message}")
+            raise e
+        except Exception as e:
+            print(f"Failed to log to Arize: {e}")
+            raise e
+
+
+@dataclass_json
+@dataclass
+class BigQueryConfig:
```
tiopramayudi, line 118: can you add doctring for this class

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @property
+    def write_location(self) -> str:
+        table_name = f"prediction_log_{self._model_id}_{self._model_version}".replace(
+            "-", "_"
+        ).replace(".", "_")
+        return f"{self._project}.{self._dataset}.{table_name}"
+
+    def write(self, dataframe: pd.DataFrame):
+        self._client.insert_rows_from_dataframe(dataframe=dataframe, table=self._table)
```
tiopramayudi, line 191: is this using streaming insert API?

python/sdk/merlin/observability/inference.py
```diff
@@ -250,3 +269,11 @@ class InferenceSchema:
     @property
     def feature_columns(self) -> List[str]:
         return list(self.feature_types.keys())
+
+    @property
+    def prediction_columns(self) -> List[str]:
```
tiopramayudi, line 274: who call this property?

python/sdk/merlin/observability/inference.py
```diff
@@ -250,3 +269,11 @@ class InferenceSchema:
...
+        return list(self.model_prediction_output.prediction_types().keys())
+
+    @property
+    def ground_truth_columns(self) -> List[str]:
```
tiopramayudi, line 278: who call this property?

python/observation-publisher/publisher/observation_sink.py
khorshuheng: Yes, it's streaming insert API.

tiopramayudi: is it easy if we use [BigQuery Storage Write API](https://cloud.google.com/bigquery/docs/write-api#api_flow) instead ?

python/sdk/merlin/observability/inference.py
khorshuheng: Not in this PR, but will be used by flyte workflow that publish ground truth eventually.

python/observation-publisher/publisher/prediction_log_consumer.py
```diff
@@ -34,22 +39,68 @@ def close(self):
...
+                if (
+                    len(buffered_logs) < self.buffer_capacity
+                    and buffered_duration < buffered_max_duration_seconds
+                ):
+                    continue
+                df = log_batch_to_dataframe(buffered_logs, inference_schema)
+                most_recent_prediction_timestamp = df[
+                    PREDICTION_LOG_TIMESTAMP_COLUMN
+                ].max()
+                MetricWriter().update_last_processed_timestamp(
```
tiopramayudi, line 65: what is the purpose of updating this?

python/observation-publisher/publisher/prediction_log_consumer.py
```diff
@@ -34,22 +39,68 @@ def close(self):
...
+    batch_size: int = 100
+    poll_timeout_seconds: float = 1.0
+    additional_consumer_config: Optional[dict] = None
+
+
 class KafkaPredictionLogConsumer(PredictionLogConsumer):
-    def __init__(self, config: KafkaConsumerConfig):
+    def __init__(
+        self,
+        buffer_capacity: int,
```
tiopramayudi, line 96: seems the buffer capacity is number of records that is being hold, why not size of the data that it holds?

python/observation-publisher/publisher/metric.py
```diff
@@ -0,0 +1,47 @@
+from pandas import Timestamp
+from prometheus_client import Gauge
+
+
+class MetricWriter(object):
```
tiopramayudi, line 5: Should we add metrics number of records that has been published to arize? So we can see the discrepancy between data from arize and ours

tiopramayudi (approved):
Adding small comments. The rest is LGTM

python/observation-publisher/publisher/prediction_log_consumer.py
khorshuheng: This is mainly to track whether the consumer is lagging. If we see that the last processed timestamp is far earlier than the current time, then it's an indicator that the processing speed is not fast enough.

PR #517, opened 2024-01-15
fix(api): Fix failing recovery of deployment attempts interrupted by API server pod termination

deadlycoconuts:
# Description
When a deployment attempt of a certain Merlin _model version_ (this happens within an API server pod) gets interrupted midway by the termination of the said API server pod, the attempt gets retried by one of the other remaining API server pods or by a new one that gets spun up to take its place. However this redeployment attempt (of the model version) will currently fail:

Screenshot 2024-01-11 at 6 00 26 PM

Not only will the deployment attempt (shown as an entry in the deployments table) show a status that will perpetually be stuck in 'pending', another deployment entry gets 'created' in the deployments table that has the 'failed' state with an error message that claims that the model version in question has already been deployed. 

Trying to manually trigger a redeployment of the model version subsequently would not work either and would instead only lead to the same error message mentioned above to be returned. (The model version in question has in reality been deployed but not known to the API server).

This PR thus introduces a fix to address the incorrect handling of interrupted redeployments, so as 1) to stop the creation of an additional deployment entry in the deployments table as well as resolve occurrences of deployments stuck in the 'pending' state, and 2) to stop the recovery/redeployment attempt from failing because an inference service corresponding to the model version in question has already been deployed.

# Context
The scenario described above can roughly be represented more accurately by the following steps which take place in the Merlin API server (note that _pod_ refers to the Merlin API server pod):
1. Deployment job gets picked up by pod 1 
2. Pod 1 creates a new entry in the deployments table
3. Pod 1 starts to deploy an inference service specified by the deployment job
4. While waiting for the deployment to complete, pod 1 gets terminated
5. Deployment job (the same one) gets picked up by pod 2
6. **_Pod 2 creates an entry in the deployments table_**
7. **_Pod 2 starts to deploy an inference service specified by the deployment job_**
8. The deployment fails because by this time the inference service deployment started in step 3 is completed
9. The 2nd deployment entry in the deployments table is saved as failed
10. The deployment job gets saved as completed in the jobs table

The first fix to be introduced adds a check at step 6, to verify that the last deployment entry in the DB for the model version is not stuck in the ‘pending’ state (which would be unexpected). If so, we’ll assume that the previous deployment attempt had failed and we would not create a new entry in the DB. If there is no deployment entry found (the model version has not been deployed before) or if the latest deployment entry is not in the ‘pending’ state, we will create a new deployment entry that corresponds to the new deployment attempt.

The next fix introduces an additional check in step 7, just before the controller is used to create an inference service, to verify if there’s already one that exists in the cluster where the model should be deployed. If one already exists, we let the deployment workflow continue as usual without creating a new one. 

# Modifications
- `api/cluster/controller.go` - Add inference service checks before new entities are created
- `api/queue/work/model_service_deployment.go` - Add deployment checks before new entries get created in the DB
- `api/storage/deployment_storage.go` - Add new helper method `GetLatestDeployment` which returns the latest Merlin model deployment entry from the deployments table in the DB

# Tests
- Existing behaviour of redeployments should remain unchanged
- Added unit tests to the refactored controller, model service deployment and deployment storage files

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
NONE
```


api/queue/work/model_service_deployment_test.go
```diff
@@ -488,11 +601,16 @@ func TestExecuteDeployment(t *testing.T) {
 				assert.Equal(t, env.DefaultResourceRequest, savedEndpoint.ResourceRequest)
 			}
 
+			mockDeploymentStorage.AssertNumberOfCalls(t, "GetLatestDeployment", 1)
 			if tt.deployErr != nil {
 				mockDeploymentStorage.AssertNumberOfCalls(t, "Save", 2)
 				assert.Equal(t, models.EndpointFailed, savedEndpoint.Status)
 			} else {
-				mockDeploymentStorage.AssertNumberOfCalls(t, "Save", 1)
+				if tt.endpoint.Status == models.EndpointPending {
```
deadlycoconuts, line 609: This condition checks if an endpoint is originally in the 'pending' state - if it is, we will expect the deployment attempt to not bother creating a new deployment entry in the deployments table, hence the 0 calls expected for the number of db save operations.

api/cluster/controller.go
```diff
@@ -223,10 +223,19 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 		return nil, errors.Wrapf(err, fmt.Sprintf("%v (%s)", ErrUnableToCreateInferenceService, isvcName))
 	}
 
-	s, err := c.kserveClient.InferenceServices(modelService.Namespace).Create(spec)
+	// check the cluster to see if the inference service has already been deployed
+	s, err := c.kserveClient.InferenceServices(modelService.Namespace).Get(modelService.Name, metav1.GetOptions{})
```
leonlnj, line 227: to confirm `modelService.Name` is unique per deployment/version/revision right?

api/cluster/controller.go
```diff
@@ -223,10 +223,19 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
...
 	if err != nil {
-		log.Errorf("unable to create inference service %s: %v", isvcName, err)
-		return nil, errors.Wrapf(err, fmt.Sprintf("%v (%s)", ErrUnableToCreateInferenceService, isvcName))
+		if !kerrors.IsNotFound(err) {
```
leonlnj, line 229: why is there an error if its found? anyway can it be rephrase so that there are no double negatives?

```
if kerrors.IsNotFound(err) {
   s, err = c.kserveClient.InferenceServices(modelService.Namespace).Create(spec)...
} else {
   ... errors.Wrapf(err, "unable to check status of inference service: %s")...
}
```

api/queue/work/model_service_deployment.go
```diff
@@ -92,18 +92,35 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
...
-	deployment, err := depl.DeploymentStorage.Save(&models.Deployment{
-		ProjectID:         model.ProjectID,
-		VersionModelID:    model.ID,
-		VersionID:         endpoint.VersionID,
-		VersionEndpointID: endpoint.ID,
-		Status:            models.EndpointPending,
-	})
-	if err != nil {
-		log.Warnf("unable to create deployment history", err)
+	// create a new entry if the there is no deployment entry found or if it is in the 'pending' state
```
leonlnj, line 102: typo `if the there is no`

api/queue/work/model_service_deployment.go
```diff
@@ -92,18 +92,35 @@ func (depl *ModelServiceDeployment) Deploy(job *queue.Job) error {
...
+	if deployment == nil || deployment.Status != models.EndpointPending {
```
leonlnj, line 103: what is the `deployment.Status != models.EndpointPending` for?
Are there any case where deployment != nil and `deployment.Status != models.EndpointPending` by the fact it got into the job queue?

leonlnj: maybe its easier to read if its just swap around (i know its the same) 😅 

When the last deployment is pending
```
if deployment != nil && deployment.Status == models.EndpointPending {
```

api/queue/work/model_service_deployment_test.go
```diff
@@ -128,6 +234,7 @@ func TestExecuteDeployment(t *testing.T) {
 			},
 			deploymentStorage: func() *mocks.DeploymentStorage {
 				mockStorage := &mocks.DeploymentStorage{}
+				mockStorage.On("GetLatestDeployment", mock.Anything, mock.Anything).Return(nil, gorm.ErrRecordNotFound)
```
leonlnj, line 237: instead of adding this to all, perhaps create a function that returns a default behaviour used by most and construct from there and override if needed? i.e

```
func createMockDeploymentStorage() *mocks.VersionEndpointStorage {
	mockStorage := &mocks.VersionEndpointStorage{}
	mockStorage.On("GetLatestDeployment", mock.Anything, mock.Anything).Return(nil, gorm.ErrRecordNotFound)
	mockStorage.On("Save", mock.Anything).Return(nil)
	mockStorage.On("OnDeploymentSuccess", mock.Anything).Return(nil)
	return mockStorage
}

```

api/cluster/controller.go
ariefrahmansyah: Yes, it is. Ref: https://github.com/caraml-dev/merlin/blob/904a8940dc8911c0948bf6452c6670ec9aef0942/api/models/service.go#L60C20-L60C46

api/queue/work/model_service_deployment.go
ariefrahmansyah: +1 on swapping the conditional

api/cluster/controller.go
deadlycoconuts: Thanks Arief for sharing the link! 🙏

api/cluster/controller.go
deadlycoconuts: > why is there an error if its found?

If it is found there won't be an error, the else block below, will be executed instead:
```go
} else {
	log.Infof("found existing inference service %s; skipping its creation", isvcName)
}
```

> anyway can it be rephrase so that there are no double negatives? 

Ah actually yeah it can 😅 I was initially avoiding it to use one less else-block but I agree it's probably better to avoid the double negative.

api/queue/work/model_service_deployment.go
deadlycoconuts: Thanks for catching this!

api/queue/work/model_service_deployment.go
deadlycoconuts: > what is the `deployment.Status != models.EndpointPending` for?

It's to check if the last entry in the deployments table has the deployment status set as pending (the deployment was aborted) or otherwise (in some terminal state like failed or deployed). In this case, if the status is a terminal state then we can safely create a new deployment entry in the table.

> Are there any case where `deployment != nil and deployment.Status != models.EndpointPending` by the fact it got into the job queue?

Uh yes, technically the deployment isn't what that gets into the job queue; a job gets into the job queue 😅 So if a job gets into a job queue, and if the last deployment entry fulfils the following condition, `deployment != nil and deployment.Status != models.EndpointPending`, that means that the last deployment is in a terminal state (deployed or failed).

> Are there any possibility that last deployment is in pending but there is no deployment ongoing? How will it be handled

Isn't this is supposed to be handled by the changes in this PR? Maybe I don't get what you're trying to ask 😅

> maybe its easier to read if its just swap around (i know its the same) 😅
When the last deployment is pending
`if deployment != nil && deployment.Status == models.EndpointPending {`

Yes I think the confusion might've stemmed from the fact that I chose to express the conditional in the inverse instead of the much more intuitive and natural alternative `if deployment != nil && deployment.Status == models.EndpointPending` (sign that I haven't coded in a long time perhaps 😅). Just so if it wasn't clear, `deployment == nil` happens when the model version is deployed for the first time.

api/queue/work/model_service_deployment_test.go
deadlycoconuts: Hmm that's true, that'll make things cleaner. Let me try refactoring this.

api/queue/work/model_service_deployment_test.go
```diff
@@ -869,3 +1005,10 @@ func TestExecuteRedeployment(t *testing.T) {
 		})
 	}
 }
+
+func createDefaultMockDeploymentStorage() *mocks.DeploymentStorage {
```
deadlycoconuts, line 1009: Added this helper method here to build a default mocked deployment storage component.

api/queue/work/model_service_deployment.go
leonlnj: the confusion aside, just for my conceptual understanding
1. Merlin API Starts up, the jobqueue get populated with incompleted jobs
2. Incomplete jobs are processed by workers and `deployed` by this service
3. Hence`deployment.Status == models.EndpointPending` will always be true, since they are incompleted?

deadlycoconuts: Ahaha no worries, but yes, that'll always be true because the previous worker that was deploying the model wouldn't have updated the DB's deployments table to change the deployment's status from `pending` to `failed`/`deployed`.

leonlnj (approved):
thanks, lgtm

deadlycoconuts:
Thanks for the comments everyone, merging this now! :D

PR #518, opened 2024-01-16
feat: Model schemas API and migrate client to OpenAPI 3 

tiopramayudi:
# Description

A model to onboard to model monitoring must define the schema that indicate the contract of the data that is required to do prediction also what is the actual or ground truth for that model. During the implementation it needs to upgrade to OpenAPI 3 due to lack of `oneOf` in OpenAPI 2.x

# Modifications

* Regenerate python client
* More strict check in SDK
* Model schema

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```


khorshuheng:
@tiopramayudi Seems like we need to fix the dependency requirement, pydantic is now required

ariefrahmansyah:
Can we remove `swagger.yaml` entirely?

khorshuheng (approved):

tiopramayudi:
> Can we remove `swagger.yaml` entirely?

For now I replace the content of `swagger.yaml` with OpenAPI 3 specs, the intention is to reduce number of changes, since we also have deployment of merlin swagger that still refer to swagger.yaml

PR #519, opened 2024-01-17
feat(api): Introduce CPU and memory limit request factors to API server

deadlycoconuts:
# Description
In order to allow the resource limits of any pod deployed by the Merlin API server to be scaled automatically as a factor of the resource request value initially decided by the user, this PR introduces two new config values, `UserContainerCPULimitRequestFactor` and `UserContainerMemoryLimitRequestFactor`, which will be used to determine the resource limit when multiplied by the corresponding resource request value.

In addition, this PR also allows the CPU resource limit values to be set as `0`, which would represent no resource limits being applied to pods deployed by the Merlin API server, when these new config values as set as `0`. 

**_This PR contains changes similar to what is implemented for Turing in https://github.com/caraml-dev/turing/pull/367._** 

🚨 **EDIT**: KServe automatically sets default resource request and limit values when they are not specified in the KService inference service specification. However, these default (e.g. CPU) values cannot be configured manually or removed (they are actually [hardcoded](https://github.com/kserve/kserve/blob/c254e704c3719d3310391ee994123d5b49588e5d/pkg/apis/serving/v1beta1/inference_service_defaults.go#L35) in the controller code itself), and as a (hopefully temporary) workaround to this, **this PR introduces another new configuration value** `UserContainerCPUDefaultLimit` **that specifies the default CPU limit of any pods deployed by the Merlin API server, when** `UserContainerCPULimitRequestFactor` **is set to** `0`. This ensures that KServe does not impose its own default CPU limit values on them.

# Modifications
- Introduction of the `UserContainerCPULimitRequestFactor` and `UserContainerMemoryLimitRequestFactor` config values
- Introduction of conditional blocks to not apply CPU or memory limits if any of the corresponding values mentioned above are set as `0`

# Tests

- Ensure existing models redeployed have their resource limit values correctly set as a factor of their resource request values

# Checklist
- [x] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
NONE
```


api/cluster/resource/templater.go
```diff
@@ -187,21 +188,25 @@ func (t *InferenceServiceTemplater) CreateInferenceServiceSpec(modelService *mod
...
 	envVars := modelService.EnvVars
 
-	// Set cpu limit and memory limit to be 2x of the requests
-	cpuLimit := modelService.ResourceRequest.CPURequest.DeepCopy()
-	cpuLimit.Add(modelService.ResourceRequest.CPURequest)
-	memoryLimit := modelService.ResourceRequest.MemoryRequest.DeepCopy()
-	memoryLimit.Add(modelService.ResourceRequest.MemoryRequest)
+	// Set resource limits to request * userContainerCPULimitRequestFactor or UserContainerMemoryLimitRequestFactor
+	limits := map[corev1.ResourceName]resource.Quantity{}
+	if t.deploymentConfig.UserContainerCPULimitRequestFactor != 0 {
```
leonlnj, line 193: possible null?

leonlnj:
LGTM ,thanks!

ariefrahmansyah (approved):

api/cluster/resource/templater.go
deadlycoconuts: No it wouldn't be null:

https://github.com/caraml-dev/merlin/blob/main/api/cluster/resource/templater.go#L115

api/config/config.go
```diff
@@ -351,6 +352,14 @@ type PyFuncPublisherConfig struct {
 	SamplingRatioRate float64 `validate:"required" default:"0.01"`
 }
 
+// InferenceServiceDefaults captures some of the configurable defaults specific to
+// KServe inference services
+type InferenceServiceDefaults struct {
+	UserContainerCPULDefaultLimit         string  `json:"userContainerCPULDefaultLimit" default:"8"`
```
deadlycoconuts, line 358: This is the new config that we can use to specify the default CPU limit we can specify on pods deployed by Merlin. Note that this only happens when the `UserContainerCPULimitRequestFactor` is specified as `0`.

ariefrahmansyah: Is it typo for default limit: `UserContainerCPULDefaultLimit`, should it be `UserContainerCPUDefaultLimit`?

deadlycoconuts: Whoa ahaha yes it's a typo! Thanks for catching it!

leonlnj (approved):

deadlycoconuts:
Thanks for all the comments, merging this now!

PR #521, opened 2024-01-24
fix: move git initialization to function body

khorshuheng:
# Description

Import git module will fail when the execution environment does not contain git executable. As Merlin SDK does not strictly rely on git, we should move the import to the function body where it is needed instead.

# Modifications

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
Fix a bug in which pyfunc related functions will fail when the execution environment does not have git executable.
```


ariefrahmansyah (approved):
LGTM! Thanks Shu Heng!

PR #523, opened 2024-01-25
fix: make sure transformer property on version endpoint is not undefined

khorshuheng:
# Description

The "transformer" property could be undefined in the event where transformer is None, because the private property, _transformer is not set in that case.

# Modifications

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
Fix a bug in which transformer property is undefined for model version when there is no transformer found for the end point.
```


ariefrahmansyah (approved):
lgtm. thanks, Shu Heng!

PR #524, opened 2024-01-26
Propagate image building and model deployment errors

ariefrahmansyah:
# Description

Returns more information when failures happen on image building and model deployment process by propagating related kubernetes resources.

For image building, returns:
1. Kubernetes job conditions
2. Related pod's container status (in this case pyfunc-image-builder container)
3. Related pod's last termination message

For model deployment, returns:
1. Kserve inference service conditions
2. Related pod's container status
3. Related pod's last termination message

## Screenshots

- When image building is OOMKilled:

image

- When docker image for custom model not exist:
   
image

- When deployment timeout because no node available:

image

# Modifications

1. Add ParsePodContainerStatuses utility functions
2. Add related parsing function to parse inference service and kubernetes job.

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```


api/cluster/controller.go
```diff
@@ -325,28 +328,107 @@ func (c *controller) deleteInferenceService(serviceName string, namespace string
...
+			isvc, ok := isvcEvent.Object.(*kservev1beta1.InferenceService)
+			if !ok {
+				return nil, errors.New("unable to cast isvc object")
 			}
+			log.Debugf("isvc event received [%s]", isvc.Name)
 
-			if s.Status.IsReady() {
-				// Inference service is completely ready
-				return s, nil
+			if isvc.Status.Status.Conditions != nil {
```
leonlnj, line 384: for `<-isvcWatcher.ResultChan():` and `<-podWatcher.ResultChan():`,  are the "live" status of the pod/isvc being stream in?

I see `podContainerTable` and `isvcConditionTable` being constantly updated. Once the condition is ready (i assume `isvc.Status.Status.Conditions == nil`,  wont the `isvcConditionTable` of the previous state be included? Some comments around what is being returns and the condition would be helpful in the future.

api/cluster/controller.go
```diff
@@ -325,28 +328,107 @@ func (c *controller) deleteInferenceService(serviceName string, namespace string
...
+				isvcConditionTable, err = parseInferenceServiceConditions(isvc.Status.Status.Conditions)
```
leonlnj, line 385: would it be better to keep track of the last state of  `isvc, ok := isvcEvent.Object.(*kservev1beta1.InferenceService)` and `pod, ok := podEvent.Object.(*corev1.Pod)`, then do the parsing once in the defer instead of every message

api/cluster/controller.go
```diff
@@ -325,28 +328,107 @@ func (c *controller) deleteInferenceService(serviceName string, namespace string
...
+			}
+
+		case podEvent := <-podWatcher.ResultChan():
+			pod, ok := podEvent.Object.(*corev1.Pod)
+			if !ok {
+				return nil, errors.New("unable to cast pod object")
+			}
+			log.Debugf("pod event received [%s]", pod.Name)
+
+			if len(pod.Status.ContainerStatuses) > 0 {
```
leonlnj, line 399: any reason why not `pod.Status.Conditions`like isvc? It is to shows latest current individual container by status?

api/cluster/controller.go
leonlnj: oh I think everytime there is an update this is involved but the conditions of each update are being appended to the resource. CMIIW

api/utils/kubernetes.go
```diff
@@ -0,0 +1,48 @@
+package utils
+
+import (
+	"errors"
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+func ParsePodContainerStatuses(containerStatuses []v1.ContainerStatus) (string, string, error) {
```
leonlnj, line 10: Probably some comment here would be good or rather the error is the error while executing this function instead of the implicit meaning, as its normally expected 
`"", "". err` or `"something", "something", nil`

leonlnj (approved):
Thanks, LGTM! Left some minor comments, but overall lgtm

api/cluster/controller.go
ariefrahmansyah: -> for <-isvcWatcher.ResultChan(): and <-podWatcher.ResultChan():, are the "live" status of the pod/isvc being stream in?
Yes, correct. `ResultChan` returns a chan which will receive all the events. Then we use `select` to await both of these values simultaneously.


-> wont the isvcConditionTable of the previous state be included?
No, we only get the latest isvc's condition. I'll add comments there.

api/cluster/controller.go
ariefrahmansyah: In my opinion, in terms of performance, your suggestion would be more efficient since we don't need to parse the k8s objects on every event. But I prefer to have `defer` function as short as possible, hence no need to have complex logic here and just return the error message if exist.

PR #525, opened 2024-01-29
fix: Reduce memory usage when publishing prediction log to kafka

tiopramayudi:
# Description

We've seen gradual increase of memory usage when model observability is enabled for a model.
 
First hypothesis why this happened is due to using asyncio, because we pass prediction input and output to the async function. We try to prove it by reducing sampling rate to 0, since the async function that being called need to publish to kafka, so we need to isolate this only for asyncio overhead. After set sampling rate to 0 the memory usage is stable there is no gradual increase

Since first hypothesis is not correct, we have new hypothesis that this is due to publishing the data to kafka, and we did memory profiler to the model

PS: We use memray as profiler https://github.com/bloomberg/memray
 
Screen Shot 2024-01-29 at 09 47 21

Screen Shot 2024-01-29 at 09 49 11

Screen Shot 2024-01-29 at 09 48 52

We see that the memory usage is keep increasing and producing the message to kafka contribute to this.

# Modifications
To solve this problem, kafka producer must call `poll` after publish the message, this is necessary so `ack` buffer from producer will be drained and the memory usage won't gradually increase, ref: [1](https://github.com/openedx/event-bus-kafka/issues/10) , [2](https://github.com/confluentinc/librdkafka/wiki/FAQ#when-and-how-should-i-call-rd_kafka_poll)

After the changes
Screen Shot 2024-01-29 at 09 49 02

Screen Shot 2024-01-29 at 09 49 19

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/python/pyfunc-server/pyfuncserver/publisher/kafka.py
+++ b/python/pyfunc-server/pyfuncserver/publisher/kafka.py
@@ -23,3 +23,4 @@ class KafkaProducer(Producer):
         prediction_log = build_prediction_log(pyfunc_output=data, model_manifest=self.model_manifest)
         serialized_data = prediction_log.SerializeToString() 
         self.producer.produce(topic=self.topic, value=serialized_data)
+        self.producer.poll(0)


khorshuheng (approved):
lgtm, thanks!

ariefrahmansyah:
On a separate note, what tool are you using for memory profiling? @tiopramayudi

tiopramayudi:
> On a separate note, what tool are you using for memory profiling? @tiopramayudi

I use memray as profiler https://github.com/bloomberg/memray

python/pyfunc-server/pyfuncserver/publisher/kafka.py
```diff
@@ -23,3 +23,4 @@ def produce(self, data: PyFuncOutput):
         prediction_log = build_prediction_log(pyfunc_output=data, model_manifest=self.model_manifest)
         serialized_data = prediction_log.SerializeToString() 
         self.producer.produce(topic=self.topic, value=serialized_data)
+        self.producer.poll(0)
```
leonlnj, line 26: Why is this not required in the [merlin logger](https://github.com/caraml-dev/merlin/blob/9f3b161527caea636e5a5e3fed77b4b1d68cab99/api/pkg/kafka/kafka.go#L84) 

Is this because of the `"acks": publisher_config.kafka.acks,` above?

tiopramayudi: I don't think so. I think this is due to golang client will create poll go routine when [initialize producer](https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/producer.go#L594) hence no need to call poll and no `poll` function from golang producer

PR #527, opened 2024-01-30
fix: Retraining model for PyFunc E2E test

tiopramayudi:
# Description

Currently PyFunc E2E test is using pre-built model artifact, this will running fine when the dependency used for the model artifact is the same with the server. In scenario where dependency is different between artifact and the server, the artifact load will be failed in the PyFunc server, hence we need to retraining the model and using the output of that as artifact for PyFunc server in E2E script

# Modifications

Retraining xgboost and sklearn model 

# Tests

# Checklist
- [x] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/python/sdk/test/pyfunc_integration_test.py
+++ b/python/sdk/test/pyfunc_integration_test.py
@@ -17,14 +17,16 @@ import warnings
 from test.utils import undeploy_all_version
 
 import joblib
-import merlin
 import numpy as np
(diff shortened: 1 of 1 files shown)


ariefrahmansyah (approved):
lgtm. thanks, tio!

PR #531, opened 2024-02-05
feat: reuse the same bigquery table for multiple model versions for Arize BQ Sink

khorshuheng:
# Description

The current sink create a new bigquery table per model version. This makes it harder to implement Arize ground truth ingestion, because the ground truth provided by the users are typically model version agnostic.

# Modifications

- A single table will be used per model id, rather than model version
- A new column, model_version, is added.
- session id and row id are used in favor of prediction id as the concept of prediction id in Arize differs from Merlin

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/python/observation-publisher/publisher/observation_sink.py
+++ b/python/observation-publisher/publisher/observation_sink.py
@@ -1,6 +1,6 @@
 import abc
+import time
 from dataclasses import dataclass
-from datetime import datetime, timedelta
 from typing import List, Tuple
 
 import pandas as pd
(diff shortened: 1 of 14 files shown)


python/observation-publisher/publisher/observation_sink.py
```diff
@@ -207,13 +287,31 @@ def schema_fields(self) -> List[SchemaField]:
 
     @property
     def write_location(self) -> str:
-        table_name = f"prediction_log_{self._model_id}_{self._model_version}".replace(
-            "-", "_"
-        ).replace(".", "_")
-        return f"{self._project}.{self._dataset}.{table_name}"
+        table_name = f"prediction_log_{self._model_id}".replace("-", "_").replace(
```
tiopramayudi, line 290: thinking out loud, should we make it as configuration?

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -207,13 +287,31 @@ def schema_fields(self) -> List[SchemaField]:
...
+            try:
+                response = self._client.insert_rows_from_dataframe(
+                    dataframe=dataframe, table=self._table
+                )
+                errors = [error for error_chunk in response for error in error_chunk]
+                if len(errors) > 0:
+                    print(
+                        f"Errors when inserting rows to BigQuery, retrying attempt {i}/{self.retry.retry_attempts}"
+                    )
+                    time.sleep(self.retry.retry_interval_seconds)
```
tiopramayudi, line 306: will it create build up on the unprocessed kafka messages?

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -207,13 +287,31 @@ def schema_fields(self) -> List[SchemaField]:
...
+                else:
+                    return
+            except NotFound as e:
+                print(
+                    f"Table not found: {e}, retrying attempt {i}/{self.retry.retry_attempts}"
+                )
+                time.sleep(self.retry.retry_interval_seconds)
```
tiopramayudi, line 313: why do we need to retry for not found table?

python/observation-publisher/publisher/observation_sink.py
khorshuheng: Yes, it does, so retry is only useful in the event where the incoming message rate is low enough.

python/observation-publisher/publisher/observation_sink.py
khorshuheng: When the Bigquery table is first created, it does take ~10 seconds for the table to be ready.

tiopramayudi (approved):
LGTM ! Thanks @khorshuheng

PR #534, opened 2024-02-09
fix(cicd): Reduce memory footprint of e2e tests

deadlycoconuts:
# Description
Recently the e2e test job has been [failing](https://github.com/caraml-dev/merlin/actions/runs/7880641930/job/21503099773) due to errors in deploying a model version, which would messages such as:
```
An error occurred (XMinioStorageFull) when calling the PutObject operation: Storage backend has reached its minimum free disk threshold. Please delete a few objects to proceed.
```
Upon digging further, there appears to be a lack of ephemeral storage on the k3d node we set up in the e2e tests, which causes a deployment's pods (the transformer usually) to be evicted:
```
The node was low on resource: ephemeral-storage. Threshold quantity: 778512530, available: 748868Ki.
```
Apparently we’ve been encountering this problem for some time already (@ariefrahmansyah tried to debug and address this in a previous PR #446), and we already tried some of the workarounds suggested here:
- https://k3d.io/v5.4.6/faq/faq/#pods-evicted-due-to-lack-of-disk-space
- https://github.com/k3d-io/k3d/issues/133

As a quick fix to workaround these issues given that some of the workarounds have already been implemented, this PR introduces some changes to the e2e test setup:
- Deleting the model created after each e2e test
- Making the e2e tests (run using pytest) serialised (not [parallel](https://github.com/caraml-dev/merlin/blob/e110a96763e96333518b53ef6b9c18785fd08fa0/scripts/e2e/run-e2e.sh#L36))
- Using docker system prune instead of just docker image prune [before](https://github.com/caraml-dev/merlin/blob/e110a96763e96333518b53ef6b9c18785fd08fa0/.github/workflows/merlin.yml#L474) running the e2e tests
- Lowering of kubelet eviction thresholds even further from 1% to 0.5%

# Modifications
- `.github/workflows/merlin.yml` - Reduction of kubelet eviction thresholds and replacement of docker image pruning with a system-level pruning
- `python/sdk/test/integration_test.py` - Addition of steps to delete model versions created after each test run
- `scripts/e2e/run-e2e.sh` - Reduction of pytest workers from 4 to 1 to remove parallelism

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
NONE
```

### DIFF
--- a/.github/workflows/merlin.yml
+++ b/.github/workflows/merlin.yml
@@ -470,8 +470,8 @@ jobs:
       - name: Deploy merlin and mlp
         working-directory: merlin/scripts/e2e
         run: ./deploy-merlin.sh ${{ env.INGRESS_HOST }} ${{ env.LOCAL_REGISTRY }}:${{ env.LOCAL_REGISTRY_PORT }} ${{ needs.create-version.outputs.version }} ${{ github.ref }} ${{ env.MERLIN_CHART_VERSION }}
-      - name: Prune docker image to make some space
-        run: docker image prune --all --force
+      - name: Prune docker system to make some space
+        run: docker system prune --all --force
       - name: Print space consumption
         run: sudo df -h
       - name: Run E2E Test
--- a/.github/workflows/merlin.yml
+++ b/.github/workflows/merlin.yml
@@ -441,10 +441,10 @@ jobs:
           k3d registry create $LOCAL_REGISTRY --port $LOCAL_REGISTRY_PORT
           k3d cluster create $K3D_CLUSTER --image rancher/k3s:${K3S_VERSION} --port 80:80@loadbalancer \
             --k3s-arg '--disable=traefik,metrics-server@server:*' \
-            --k3s-arg '--kubelet-arg=eviction-hard=imagefs.available<1%,nodefs.available<1%@server:0' \
-            --k3s-arg '--kubelet-arg=eviction-minimum-reclaim=imagefs.available=1%,nodefs.available=1%@server:0' \
-            --k3s-arg '--kubelet-arg=eviction-hard=imagefs.available<1%,nodefs.available<1%@agent:*' \
-            --k3s-arg '--kubelet-arg=eviction-minimum-reclaim=imagefs.available=1%,nodefs.available=1%@agent:*'
+            --k3s-arg '--kubelet-arg=eviction-hard=imagefs.available<0.5%,nodefs.available<0.5%@server:0' \
+            --k3s-arg '--kubelet-arg=eviction-minimum-reclaim=imagefs.available=0.5%,nodefs.available=0.5%@server:0' \
+            --k3s-arg '--kubelet-arg=eviction-hard=imagefs.available<0.5%,nodefs.available<0.5%@agent:*' \
+            --k3s-arg '--kubelet-arg=eviction-minimum-reclaim=imagefs.available=0.5%,nodefs.available=0.5%@agent:*'
       - name: Setup cluster
         working-directory: merlin/scripts/e2e
         run: ./setup-cluster.sh merlin-cluster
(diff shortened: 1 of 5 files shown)


python/sdk/test/conftest.py
```diff
@@ -82,7 +82,12 @@ def model(project, mlflow_url, api_client):
 @pytest.fixture
 def version(project, model, mlflow_url, api_client):
     mlflow.set_tracking_uri(mlflow_url)
-    r = mlflow.start_run()
+    run_id = None
+    if not mlflow.get_experiment_by_name("unit_test_experiment"):
+        run_id = mlflow.create_experiment(
+            name="unit_test_experiment"
+        )
+    r = mlflow.start_run(experiment_id=run_id)
```
deadlycoconuts, line 90: With the removal of pytest parallelism, e2e tests that get run before this test would've already created (and deleted with the changes in this PR) Mlflow experiments, leaving behind a default experiment that does not exist. This causes errors like [these](https://github.com/caraml-dev/merlin/actions/runs/7855799983/job/21437927704#step:15:169) when this test tries to start a new run:
```python3
>           raise MlflowException(
                "Could not find experiment with ID %s" % experiment_id,
                databricks_pb2.RESOURCE_DOES_NOT_EXIST,
            )
E           mlflow.exceptions.MlflowException: Could not find experiment with ID 25
``` 

This happens because `start_run` will "look for valid experiment in the following order: activated using set_experiment, MLFLOW_EXPERIMENT_NAME environment variable, MLFLOW_EXPERIMENT_ID environment variable, or the default experiment as defined by the tracking server" when the `experiment_id` kwarg is not specified (see [this](https://mlflow.org/docs/1.30.0/python_api/mlflow.html#mlflow.start_run) for more details).

To ensure this doesn't happen, we simply force the creation of a new experiment solely for this fixture to create its own Mlflow run.

.github/workflows/merlin.yml
```diff
@@ -470,8 +470,10 @@ jobs:
       - name: Deploy merlin and mlp
         working-directory: merlin/scripts/e2e
         run: ./deploy-merlin.sh ${{ env.INGRESS_HOST }} ${{ env.LOCAL_REGISTRY }}:${{ env.LOCAL_REGISTRY_PORT }} ${{ needs.create-version.outputs.version }} ${{ github.ref }} ${{ env.MERLIN_CHART_VERSION }}
-      - name: Prune docker image to make some space
-        run: docker image prune --all --force
+      - name: Prune docker system to make some space
+        run: docker system prune --all --force
+      - name: Print space consumption
+        run: sudo df -h
```
deadlycoconuts, line 476: I don't see much point in keeping these here but I'm okay with leaving them too.

leonlnj: leave it because its highly likely we need to shift the runner and these will be useful then

.github/workflows/merlin.yml
```diff
@@ -441,10 +441,10 @@ jobs:
           k3d registry create $LOCAL_REGISTRY --port $LOCAL_REGISTRY_PORT
           k3d cluster create $K3D_CLUSTER --image rancher/k3s:${K3S_VERSION} --port 80:80@loadbalancer \
             --k3s-arg '--disable=traefik,metrics-server@server:*' \
-            --k3s-arg '--kubelet-arg=eviction-hard=imagefs.available<1%,nodefs.available<1%@server:0' \
-            --k3s-arg '--kubelet-arg=eviction-minimum-reclaim=imagefs.available=1%,nodefs.available=1%@server:0' \
-            --k3s-arg '--kubelet-arg=eviction-hard=imagefs.available<1%,nodefs.available<1%@agent:*' \
-            --k3s-arg '--kubelet-arg=eviction-minimum-reclaim=imagefs.available=1%,nodefs.available=1%@agent:*'
+            --k3s-arg '--kubelet-arg=eviction-hard=imagefs.available<0.5%,nodefs.available<0.5%@server:0' \
```
leonlnj, line 444: what does these do?

leonlnj (approved):
LGTM thanks!

.github/workflows/merlin.yml
deadlycoconuts: The changes? It just changes the kubelet's threshold for the NodeHasDiskPressure status (of space left on the filesystem). So the lower the less likely we get this status, and the less likely pods get evicted.

PR #539, opened 2024-02-16
feat(api): Expose tolerations and node selectors for batch jobs

deadlycoconuts:
# Description
This PR introduces changes to expose tolerations and node selectors that can be configured for batch jobs. Previously, these fields were hardcoded and there was no way to change them other than by making changes to the Merlin API server. With the new changes, these fields can be configured via the Merlin API server config values file.

For this to be done, a new `batchJobTemplater` attribute of a new `BatchJobTemplater` type is added to the batch controller, just like how the real-time Merlin model controller [has](https://github.com/caraml-dev/merlin/blob/683884464ced1da2565b9a1194bc370a17cb317b/api/cluster/controller.go#L94) a `kfServingResourceTemplater` that propagates default configuration values for real-time models to the model specs.

This new `BatchJobTemplater` struct only carries the new configs corresponding to the tolerations and node selectors that are set on the Merlin API server level, and 'adopts' the functions `CreateSparkApplicationResource`, `createSpec`, `createDriverSpec` and `createExecutorSpec` that were previously merely functions (i.e. these are now made methods of the `BatchJobTemplater` struct).

🚨 This PR also adds a new GitHub action to e2e tests to clean up disk storage on the GitHub runner to allow the e2e tests to pass.

# Modifications
- `api/batch/controller.go` - Addition of the `batchJobTemplater` attribute to the batch controller
- `api/batch/resource.go` - Addition of the new `BatchJobTemplater` struct and refactoring of its associated methods
- `api/config/batch.go` - Addition of a new `BatchConfig` struct to contain the tolerations and node selectors to be added to batch jobs
- `.github/workflows/merlin.yml` - Adds new GitHub action to e2e tests to clean up disk storage on the GitHub runner to allow the e2e tests to pass

# Tests
- [x] Running new batch jobs with custom tolerations/node selectors

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
None
```

### DIFF
--- a/.github/workflows/merlin.yml
+++ b/.github/workflows/merlin.yml
@@ -472,6 +472,13 @@ jobs:
         run: ./deploy-merlin.sh ${{ env.INGRESS_HOST }} ${{ env.LOCAL_REGISTRY }}:${{ env.LOCAL_REGISTRY_PORT }} ${{ needs.create-version.outputs.version }} ${{ github.ref }} ${{ env.MERLIN_CHART_VERSION }}
       - name: Prune docker system to make some space
         run: docker system prune --all --force
+        # This step is needed to free up around 24GiB of storage on the GitHub runner to allow the e2e tests to pass
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@main
+        with:
+          # this will remove tools that are actually needed (like Python, which is needed),
+          # if set to "true" but frees about 6 GB
+          tool-cache: false
       - name: Print space consumption
         run: sudo df -h
       - name: Run E2E Test
--- a/api/batch/resource.go
+++ b/api/batch/resource.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta2"
+	"github.com/caraml-dev/merlin/config"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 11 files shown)


api/batch/resource.go
```diff
@@ -71,25 +72,22 @@ var (
...
-	}
-
 	defaultHadoopConf = map[string]string{
 		hadoopConfEnableServiceAccountKey: hadoopConfEnableServiceAccount,
 		haddopConfServiceAccountPathKey:   haddopConfServiceAccountPath,
 	}
-
-	defaultNodeSelector = map[string]string{
-		"node-workload-type": "batch",
-	}
```
deadlycoconuts, line 88: These effectively removes the default tolerations and node selectors. I'm not sure if we'd like to keep this, and if we do, maybe we can consider moving them to the CaraML chart?

leonlnj: should be defaulted as part of chart

scripts/e2e/values-e2e.yaml
```diff
@@ -15,6 +15,14 @@ config:
     EncryptionKey: password
   StandardTransformerConfig:
     ImageName: ghcr.io/caraml-dev/merlin-transformer:0.31.3
+  BatchConfig:
```
leonlnj, line 18: just to confirm this is already supported by the current [chart](https://github.com/caraml-dev/helm-charts/blob/main/charts/merlin/templates/merlin-config-secret.yaml#L10)

leonlnj (approved):
LGTM, thanks

scripts/e2e/values-e2e.yaml
deadlycoconuts: Yep, the chart just dumps everything in this `config` field (which is where the `BatchConfig` will be) into the merlin configs. So technically the chart doesn't need to be changed in any way (we might just want to add the defaults in that's all)

ariefrahmansyah (approved):

scripts/e2e/run-e2e.sh
```diff
@@ -33,4 +33,4 @@ kubectl create namespace ${E2E_PROJECT_NAME} --dry-run=client -o yaml | kubectl
 cd ../../python/sdk
 pip install pipenv==2023.7.23
 pipenv install --dev --skip-lock --python ${PYTHON_VERSION}
-pipenv run pytest -n=1 -W=ignore --cov=merlin -m "not (gpu or feast or batch or pyfunc or local_server_test or cli or customtransformer)" --durations=0
+pipenv run pytest -n=4 -W=ignore --cov=merlin -m "not (gpu or feast or batch or pyfunc or local_server_test or cli or customtransformer)" --durations=0
```
deadlycoconuts, line 36: Reverted this to 4 runners since the disk storage issue no longer affects us.

.github/workflows/merlin.yml
```diff
@@ -472,6 +472,13 @@ jobs:
         run: ./deploy-merlin.sh ${{ env.INGRESS_HOST }} ${{ env.LOCAL_REGISTRY }}:${{ env.LOCAL_REGISTRY_PORT }} ${{ needs.create-version.outputs.version }} ${{ github.ref }} ${{ env.MERLIN_CHART_VERSION }}
       - name: Prune docker system to make some space
         run: docker system prune --all --force
+        # This step is needed to free up around 24GiB of storage on the GitHub runner to allow the e2e tests to pass
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@main
+        with:
+          # this will remove tools that are actually needed (like Python, which is needed),
+          # if set to "true" but frees about 6 GB
+          tool-cache: false
```
deadlycoconuts, line 481: This new step allows to us to clean up around 24GiB of storage on the GitHub runner.

PR #540, opened 2024-02-20
pin mlflow

leonlnj:
# Description

# Modifications

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```


khorshuheng:
I don't need mlflow 1.30.0 on the observation publisher. I can update the dependencies.

khorshuheng:
Alternatively, instead of pinning mlflow version on the SDK, we can consider pinning the version in https://github.com/caraml-dev/merlin/blob/main/python/batch-predictor/requirements.txt instead.

khorshuheng (approved):

PR #543, opened 2024-02-21
Refactor inference schema

khorshuheng:
# Description

# Modifications

- Add session_id and row_id column to the InferenceSchema class.
- Use actual_score instead of actual_label for binary classification
- Remove prediction group id attribute from ModelPredictionOutput, since the prediction group id is equivalent to the session id.
- Remove prediction id column since Arize doesn't require the column name to be the same for both prediction and ground truth ingestion.
- Add optional feature orders column to support standard model use case

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```


tiopramayudi (approved):
LGTM!

tiopramayudi:
can you update the documentation also?

PR #544, opened 2024-02-26
feat: Support HTTP predictor for UPI

tiopramayudi:
# Description

We faced issue with scalability of PyFunc model in gRPC, compare to PyFunc http, the gRPC  PyFunc needs more pods to serve the same amount of traffic. To solve this issue we add support for gRPC UPI to call predictor that use HTTP 1 protocol. User can enables it by specifying transformer env variable `PREDICTOR_UPI_HTTP_ENABLED` and set it to true
# Modifications

* `api/pkg/transformer/server/grpc/server.go` -> Model predictor http support
* `api/cluster/resource/templater.go` 
  * Change the predictor protocol to `HTTP_JSON`, remove h2c container port and change liveness probe
# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/pkg/transformer/server/grpc/predictor.go
+++ b/api/pkg/transformer/server/grpc/predictor.go
@@ -1,0 +1,158 @@
+package grpc
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/afex/hystrix-go/hystrix"
+	hystrixGo "github.com/afex/hystrix-go/hystrix"
+	hystrixpkg "github.com/caraml-dev/merlin/pkg/hystrix"
+	"github.com/caraml-dev/merlin/pkg/transformer/server/config"
+	upiv1 "github.com/caraml-dev/universal-prediction-interface/gen/go/grpc/caraml/upi/v1"
+	"github.com/go-coldbrew/grpcpool"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+type upiPredictorClient interface {
+	predict(ctx context.Context, payload *upiv1.PredictValuesRequest) (*upiv1.PredictValuesResponse, error)
+	close() error
+}
+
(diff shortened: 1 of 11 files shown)


api/pkg/transformer/server/grpc/server.go
```diff
@@ -53,6 +57,7 @@ type UPIServer struct {
 	opts                  *config.Options
 	modelClient           upiv1.UniversalPredictionServiceClient
 	conn                  grpcpool.ConnPool
+	httpClient            hystrixHttpClient
```
khorshuheng, line 60: Instead of having both httpClient and modelClient, would it be possible to replace both of them with a single interface? We can instantiate different implementation of the interface depending on whether UPI on http is enabled or not.

tiopramayudi: that's a good idea. I already updated this

leonlnj:
I took a brief look, will look deeper again. One question, how would user get the grpc endpoint of the transformer? 

I image user create a HTTP model via UI or SDK, the model endpoint is easily available via the UI, how would they get the exposed grpc endpoint?

tiopramayudi:
> I took a brief look, will look deeper again. One question, how would user get the grpc endpoint of the transformer?
> 
> I image user create a HTTP model via UI or SDK, the model endpoint is easily available via the UI, how would they get the exposed grpc endpoint?

The SDK and UI will generate url for gRPC endpoint no behaviour changes on that

api/pkg/transformer/server/grpc/predictor.go
```diff
@@ -0,0 +1,158 @@
...
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+type upiPredictorClient interface {
+	predict(ctx context.Context, payload *upiv1.PredictValuesRequest) (*upiv1.PredictValuesResponse, error)
+	close() error
+}
+
+func newUPIPredictorClient(opts *config.Options) (upiPredictorClient, error) {
+	if opts.PredictorUPIHTTPEnabled {
```
leonlnj, line 30: may need to check if transformer enabled as well here?

tiopramayudi: it should be checked before deployment

leonlnj (approved):
did some testing, looks fine to me, thanks!

PR #546, opened 2024-03-01
feat(api,ui): Add new max allowed replicas field to env configs and relevant checks

deadlycoconuts:
# Description
Previously the Merlin API server does not have any constraints on the max number of replicas users are able to set for their deployments. This constraint however, appears on the UI via the variable `REACT_APP_MAX_ALLOWED_REPLICA` that can be set in the configs. In order to make the max allowed replica constraint consistent across the API server (as well as the SDK) and the UI, this PR introduces a new configuration value under the `Environment` struct. This struct currently already contains other environment (cluster) related configs such as the max allowable CPU and memory a user is allowed to set.

Doing so thus requires a very simple DB update in order to introduce a new column corresponding to the max allowed replica field, and also allows the UI to retrieve this max allowed replica value directly from the environments API endpoint (just as it does for other environment-related configs such as GPU information) instead of via a React app configuration value.

https://github.com/caraml-dev/merlin/assets/36802364/7c37be61-d540-495e-b12c-4efa92ebae0f

## Additional bug fixes
Removal of the requirement for the `endpoint` and `version` props to be set in the `DeploymentPanelHeader` component, which causes a warning to occur show when no endpoint is selected for a specific version:
Screenshot 2024-03-02 at 12 58 48 PM

Resetting of the `resource_request` schema after the `None` transformer type is selected AFTER any other transformer type is selected and configured with invalid `resource_request` values rejected by yup validation (this happens because the invalid resource request values do not get removed even after the transformer has been 'removed'; note that a `None` type transformer still gets certain default transformer configs saved though they are technically redundant - it might be worth changing this in the future):

https://github.com/caraml-dev/merlin/assets/36802364/062694a1-cca6-4fed-a869-d01ab0958ce8

# Modifications
- `api/cluster/controller.go` - Addition of new max replica check on the API server for **Merlin models** during deployment
- `api/config/config.go` - Removal of `MaxAllowedReplica` from the `ReactAppConfig` struct
- `api/config/environment.go` - Addition of `MaxAllowedReplica` to the `EnvironmentConfig` struct
- `ui/src/config.js` - Removal of `REACT_APP_MAX_ALLOWED_REPLICA` from React configs
- `ui/src/pages/version/components/forms/DeployModelVersionForm.js` - Addition of a new state and state setter to set the max allowed replica value based on the environment selected in the form context
- `ui/src/pages/version/components/forms/components/DeploymentConfigPanel.js` - Addition of a new effect update to update the max allowed replica value when the environment in a deployment form is updated
- `ui/src/pages/version/components/forms/components/ResourcesPanel.js` - Removal of ticks and the addition of min/max labels to indicate variable min/max replica values users may set for a deployment
- `ui/src/pages/version/components/forms/validation/schema.js` - Update validation schema to use a variable max allowed replica value to determine the struct validation outcomes

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
This update potentially allows users to set a larger max replica value than usual for their deployments and may also inadvertently constrain the number of max replicas they can set, depending on the max allowed replica value set by the Merlin API server operator.
```

### DIFF
--- a/ui/src/pages/version/components/forms/components/DeploymentConfigPanel.js
+++ b/ui/src/pages/version/components/forms/components/DeploymentConfigPanel.js
@@ -45,6 +45,7 @@ const DEFAULT_AUTOSCALING_POLICY = {
  * @param {*} onChange. Callback to be made when configuration is changed
  * @param {*} errors. Why pass errors?
  * @param {*} isEnvironmentDisabled. Disable deployment to the environment if the flag is true.
+ * @param {*} setMaxAllowedReplica State setter to update MaxAllowedReplica constraint
  */
 export const DeploymentConfigPanel = ({
   environment,
@@ -53,6 +54,7 @@ export const DeploymentConfigPanel = ({
   onChange,
   errors = {},
   isEnvironmentDisabled = false,
+  setMaxAllowedReplica,
 }) => {
   const environments = useContext(EnvironmentsContext);
 
@@ -85,6 +87,12 @@ export const DeploymentConfigPanel = ({
         onChange(key)(versionEndpoint[key]);
       });
     }
+
+    environments.forEach((env) => {
+      if (env.name === value) {
+        setMaxAllowedReplica(env.max_allowed_replica);
+      }
+    });
   };
 
   const onDeploymentModeChange = (deploymentMode) => {
--- a/ui/src/pages/version/components/forms/components/ResourcesPanel.js
+++ b/ui/src/pages/version/components/forms/components/ResourcesPanel.js
@@ -21,7 +21,6 @@ import EnvironmentsContext from "../../../../../providers/environments/context";
 import { calculateCost } from "../../../../../utils/costEstimation";
 import { Panel } from "./Panel";
 
-const maxTicks = 20;
 
 export const ResourcesPanel = ({
   environment: initEnvironment,
@@ -69,7 +68,16 @@ export const ResourcesPanel = ({
     } else {
       setGpuValueOptions([{ value: "None", inputDisplay: "None" }]);
     }
-  }, [resourcesConfig, resourcesConfig.gpu_name, gpus]);
+
+    if (resourcesConfig.min_replica > maxAllowedReplica) {
+      onChange("min_replica")(maxAllowedReplica - 1)
+    }
+
+    if (resourcesConfig.max_replica > maxAllowedReplica) {
+      onChange("max_replica")(maxAllowedReplica)
+    }
+
+  }, [resourcesConfig, resourcesConfig.gpu_name, gpus, maxAllowedReplica, onChange]);
 
   const replicasError = useMemo(
     () => [...(errors.min_replica || []), ...(errors.max_replica || [])],
@@ -235,10 +243,8 @@ export const ResourcesPanel = ({
             fullWidth
             min={0}
             max={maxAllowedReplica}
-            // This component only allows up to 20 ticks to be displayed at the slider
-            tickInterval={Math.ceil((maxAllowedReplica + 1) / maxTicks)}
             showInput
-            showTicks
+            showLabels
             value={[
               resourcesConfig.min_replica || 0,
               resourcesConfig.max_replica || 0,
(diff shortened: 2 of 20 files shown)


ui/src/pages/version/components/forms/components/ResourcesPanel.js
```diff
@@ -69,7 +68,16 @@ export const ResourcesPanel = ({
     } else {
       setGpuValueOptions([{ value: "None", inputDisplay: "None" }]);
     }
-  }, [resourcesConfig, resourcesConfig.gpu_name, gpus]);
+
+    if (resourcesConfig.min_replica > maxAllowedReplica) {
+      onChange("min_replica")(maxAllowedReplica - 1)
```
leonlnj, line 73: it should be allowed to set min-max as a fixed number of `maxAllowedReplica`?

ui/src/pages/version/components/forms/components/DeploymentConfigPanel.js
```diff
@@ -53,6 +54,7 @@ export const DeploymentConfigPanel = ({
   onChange,
   errors = {},
   isEnvironmentDisabled = false,
+  setMaxAllowedReplica,
```
leonlnj, line 57: can this be retrieved via environment context instead of passing as input, since they are in environment api now (if im not mistaken)

ui/src/pages/version/components/forms/components/SelectTransformerPanel.js
```diff
@@ -25,6 +25,12 @@ export const SelectTransformerPanel = ({
       onChange("enabled")(false);
       onChange("transformer_type")(undefined);
       onChange("type_on_ui")("");
+      onChange("resource_request")({
+        min_replica: process.env.REACT_APP_ENVIRONMENT === "production" ? 2 : 0,
+        max_replica: process.env.REACT_APP_ENVIRONMENT === "production" ? 4 : 2,
+        cpu_request: "500m",
+        memory_request: "512Mi"
+      })
```
deadlycoconuts, line 33: These lines set default resource request values, which gets set [here](https://github.com/caraml-dev/merlin/blob/main/ui/src/services/transformer/Transformer.js#L21) when a version endpoint object gets [created](https://github.com/caraml-dev/merlin/blob/main/ui/src/services/version_endpoint/VersionEndpoint.js#L38), which itself gets created when a new form is created when trying to deploy a model version.

leonlnj: what if its not disabled? does the `resource_request` has to be set in the `else` block as well?

ui/src/pages/version/components/forms/components/ResourcesPanel.js
deadlycoconuts: We aren't stopping the user from setting min_replica=max_replica. This condition only sets min_replica=`maxAllowedReplica - 1` as a default when a user changes the environment with the min_replica value **set higher than** the max allowed value for the new environment.

ui/src/pages/version/components/forms/components/DeploymentConfigPanel.js
deadlycoconuts: Mmm do you mean like instead of using a setter to set the state, make all other components that are dependent on this `MaxAllowedReplica` value use the environments API to retrieve that value directly?

If so, the environments context contains all the configs for all environments and we still need to filter the correct environment that is selected in this component.

ui/src/pages/version/components/forms/components/SelectTransformerPanel.js
deadlycoconuts: If it's not set to `None` (i.e. if a user subsequently chooses the standard, custom or feast transformer type), we can still keep the same `resource_request` values, since these values are common across all these transformer types. I feel like there probably isn't too much value in resetting the values if this is the case.

https://github.com/caraml-dev/merlin/assets/36802364/07b2aac9-5413-41fa-aac2-2548c3d10aa0

leonlnj (approved):
thanks!

PR #548, opened 2024-03-05
fix: Fix UPI integration test 

tiopramayudi:
# Description

In recent [PR](https://github.com/caraml-dev/merlin/pull/544) accidentally update env variable for transformer that fail the test as the infer method is not correct for upi http predictor. This PR fix the infer method and parameterize the test whether it is using http predictor or not
# Modifications

Update pyfunc_upi_integration_test.py
# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/python/sdk/test/pyfunc_upi_integration_test.py
+++ b/python/sdk/test/pyfunc_upi_integration_test.py
@@ -10,6 +10,7 @@ import pytest
 import xgboost as xgb
 from caraml.upi.utils import df_to_table, table_to_df
 from caraml.upi.v1 import type_pb2, upi_pb2, upi_pb2_grpc, variable_pb2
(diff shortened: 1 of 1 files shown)


deadlycoconuts (approved):
Thanks a lot for the fix @tiopramayudi ! Just curious, why is this test not run within the GitHub CICD pipeline though?

tiopramayudi:
> Thanks a lot for the fix @tiopramayudi ! Just curious, why is this test not run within the GitHub CICD pipeline though?

because it is  pyfunc predictor so require image building. We exclude pyfunc in github cicd due to resource consumption to run the test and some of the test using feast which complicate the setup

PR #554, opened 2024-03-20
Add support for mlobs sink on logger

khorshuheng:
# Description
This MR update the merlin logger to support writing prediction log instead of inference log to Kafka, via a new type of sink named MLObs sink, which will apply if the log url prefix is "mlobs". Merlin API service itself has not been updated to utilize this new sink. That will be implemented in another PR.

# Modifications
- Add new sink type (mlobs)
- Update upi package version
- getTopicName as part of KafkaSink / MLObs Sink methods as opposed to function in main package

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```


api/pkg/inference-logger/logger/mlobs_sink.go
```diff
@@ -0,0 +1,154 @@
...
+	modelVersion string
+}
+
+type StandardModelRequest struct {
+	Instances [][]*float64 `json:"instances"`
+	SessionId string       `json:"session_id"`
+	RowIds    []string     `json:"row_ids"`
+}
+
+type StandardModelResponse struct {
```
ariefrahmansyah, line 29: nit: rename both of them to just `ModelRequest` and `ModelResponse` without `Standard` prefix.

api/pkg/inference-logger/logger/mlobs_sink.go
```diff
@@ -0,0 +1,154 @@
...
+		},
+		Value: logBytes,
+	}, nil
+}
+
+func (m *MLObsSink) Sink(rawLogEntries []*LogEntry) error {
+	for _, rawLogEntry := range rawLogEntries {
+		predictionLog, err := m.newPredictionLog(rawLogEntry)
+		if err != nil {
+			m.logger.Errorf("unable to convert log entry: %v", err)
```
ariefrahmansyah, line 131: Should we return err here and cancel the sinking for invalid entry? Same as below error scenarios.

api/pkg/inference-logger/logger/mlobs_sink.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return predictionLog, nil
+}
+
+func (m *MLObsSink) buildNewKafkaMessage(predictionLog *upiv1.PredictionLog) (*kafka.Message, error) {
+	logBytes, err := proto.Marshal(predictionLog)
+	if err != nil {
+		return nil, err
+	}
+	topic := fmt.Sprintf("caraml-%s-%s-%s-prediction-log", m.projectName, m.modelName, m.modelVersion)
```
ariefrahmansyah, line 118: @khorshuheng, can you remind us again why we need one topic per model version instead of one topic per model?

api/pkg/inference-logger/logger/mlobs_sink.go
khorshuheng: The problem is, this request response format is only valid for standard model and not pyfunc. In fact, it is specific only to non tensorflow standard model. So we will likely still need some prefix, if not `Standard`.

api/pkg/inference-logger/logger/mlobs_sink.go
khorshuheng: I thought about this, but that will mean that the model itself will goes down if there are any issue with logging, which i am not sure if it is a desirable behaviour. The current inference log logger also print out the error log as opposed to returning error.

api/pkg/inference-logger/logger/mlobs_sink.go
khorshuheng: Currently, we are deploying one Kafka consumer per model version. Since a deployment is unique to a model version, Merlin API is able to provide the model schema (eg. feature types, prediction schema) to the deployment as a configuration.

If we use one topic per model, instead of version, then the consumer / worker needs to handle the scenario where multiple versions of prediction log are present in the same Kafka topic. In this scenario, the deployment will need to know the model schema for all the model versions, which is not going to be easy to maintain in the form of deployment configuration. Realistically, the deployment will need to call Merlin API instead to fetch and cache the schema for a specific model version.

ariefrahmansyah: Got it. Thanks for clear explanation!

ariefrahmansyah (approved):
lgtm. thanks shu heng!

PR #555, opened 2024-03-21
fix(image-building): Update process_conda_env.sh

ariefrahmansyah:
# Description

When users specify the version of mlflow in their pyfunc-based's conda.yaml, the image building might return an error like this:

```
The conflict is caused by:
    The user requested mlflow==1.23.0
    merlin-batch-predictor 0.41.0rc3 depends on mlflow==1.26.1
...
```

It's because of the version of their mlflow on their conda.yaml does not match the version set by merlin-sdk.

# Modifications

Update process_conda_env.sh to remove the mlflow version constraint.

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
NONE
```


leonlnj (approved):
thanks, lgtm!

PR #562, opened 2024-03-28
add support for observability for xgboost standard model

khorshuheng:
# Description

Currently, observability publisher cannot be deployed automatically for xgboost standard model because the current validator assumes that the support is only enabled on pyfuncv3 model. Since we update the Merlin logger to send prediction log, it is now possible to support xgboost standard model.

We will still need to check whether this will work for other types of standard model as well. But for the time being, we will enable this only for xgboost.

# Modifications

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/api/validator.go
+++ b/api/api/validator.go
@@ -2,7 +2,10 @@ package api
 
 import (
 	"context"
+	"errors"
 	"fmt"
+
+	"golang.org/x/exp/slices"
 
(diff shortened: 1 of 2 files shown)


ariefrahmansyah (approved):
lgtm!

PR #563, opened 2024-04-01
fix: add Merlin project as config to observation publisher

khorshuheng:
# Description

This PR addresses two bugs found in the observation publisher:
1. The model id config field in observation publisher is wrongly assumed to be in the form of (project)-(model id), instead of just (model id) . This affect the bigquery table write location for BigQuerySink, and arize model id for ArizeSink.
2. Prediction log from standard model does not contain the "columns" field, but this is not handled by the observation publisher.

# Modifications

- API service will now supply model project as an input to the deployer.
- Observation publisher will have a new config field, project. This is going to be passed to the observation sink as well.
- Observation publisher will now use the feature_orders field in inference schema when trying to construct the pandas dataframe from the prediction log, if the field value is non null

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```


python/observation-publisher/publisher/observation_sink.py
```diff
@@ -65,18 +67,20 @@ class ArizeSink(ObservationSink):
...
     def __init__(
         self,
+        project: str,
         inference_schema: InferenceSchema,
         model_id: str,
         model_version: str,
         arize_client: ArizeClient,
     ):
         """
+        :param project: Merlin project
```
ariefrahmansyah, line 77: nit: CaraML project

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -275,10 +278,10 @@ def schema_fields(self) -> List[SchemaField]:
 
     @property
     def write_location(self) -> str:
```
ariefrahmansyah, line 280: nit: I think it would be nice to have Python docstring here explaining the generated write_location. At first, it might be confusing when user try to differentiate between CaraML project and BQ project.

python/observation-publisher/publisher/prediction_log_parser.py
```diff
@@ -20,8 +21,11 @@ class PredictionLogFeatureTable:
     def from_struct(
         cls, table_struct: Struct, inference_schema: InferenceSchema
     ) -> Self:
-        assert isinstance(table_struct["columns"], ListValue)
-        columns = list_value_as_string_list(table_struct["columns"])
+        if inference_schema.feature_orders is not None:
+            columns = inference_schema.feature_orders
+        else:
+            assert isinstance(table_struct["columns"], ListValue)
+            columns = list_value_as_string_list(table_struct["columns"])
```
ariefrahmansyah, line 28: Is there any scenario when both conditions true (inference_schema.feature_orders is not None and table_struct["columns"] is ListValue instance)? And when this happens, inference_schema.feature_orders will take precedence?

ariefrahmansyah (approved):
lgtm with nit comments. thanks, shu heng!

python/observation-publisher/publisher/prediction_log_parser.py
khorshuheng: For users who use PyfuncV3 model, `columns` are expected to be present, and is expected to be in the correct order. Hence, there is no need for the user to explicitly specify the feature orders. Specifying the feature orders is mandatory only for the case of standard model.

So it is possible for both condition to return True, though it shouldn't be in the normal workflow. If it does happen, i believe it's fine for feature_orders to take precedence

PR #564, opened 2024-04-01
feat: handle scenarios where there is no row id but only a single prediction row

khorshuheng:
# Description

In some cases, the user is not able to provide a meaningful row id, because there is only a single prediction row and session id alone will uniquely identify the prediction in this case.  We can handle this case by creating an array of row id with a single empty string element if there is only one prediction but row id is not defined.

# Modifications

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/pkg/inference-logger/logger/mlobs_sink.go
+++ b/api/pkg/inference-logger/logger/mlobs_sink.go
@@ -80,6 +80,11 @@ func (m *MLObsSink) newPredictionLog(rawLogEntry *LogEntry) (*upiv1.PredictionLo
 	if err != nil {
 		return nil, err
 	}
+	// If there is only one prediction, session id alone is enough to for unique prediction id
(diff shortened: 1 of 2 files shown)


leonlnj:
where is rowId derived from? trying to understand why a single prediction row does not have rowId

khorshuheng:
> where is rowId derived from? trying to understand why a single prediction row does not have rowId

session id is unique per Merlin request, whereas row id should be unique per row. Ideally, the user should provide the value for both of them.

However, we have found that in the case of a single prediction row, it is far more convenient for the user to just provide the session_id. So this PR will simplify the user's configuration on standard transformer / custom service that calls Merlin.

leonlnj (approved):

PR #567, opened 2024-04-03
feat: add default sampling rate to mlobs logger

khorshuheng:
# Description

As of now, the mlobs logger logs every single message sent to the logger. As there is no good way at the moment to use different logger configuration for different model, tentatively, we will be adding a default sampling rate that will apply to all models that enable mlobs logger. 

# Modifications

Add a 2% sampling rate to mlobs sink.

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```


api/pkg/inference-logger/logger/mlobs_sink.go
```diff
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 
+	"golang.org/x/exp/rand"
```
ariefrahmansyah, line 9: Why are you using `golang.org/x/exp/rand` instead of `math/rand` package? To my knowledge, the latter does not need to run `seed` first.

khorshuheng: No particular reason actually, i wasn't aware of the math/rand package. I will checkout the math/rand package,

api/pkg/inference-logger/logger/mlobs_sink.go
```diff
@@ -19,6 +20,10 @@ var (
 	ErrMalformedLogEntry = errors.New("malformed log entry")
 )
 
+const (
+	SamplingRate = 0.02
```
tiopramayudi, line 24: The default in pyfunc is 0.01

api/pkg/inference-logger/logger/mlobs_sink.go
```diff
@@ -163,17 +168,19 @@ func (m *MLObsSink) buildNewKafkaMessage(predictionLog *upiv1.PredictionLog) (*k
...
-			m.logger.Errorf("unable to convert log entry: %v", err)
-		}
-		kafkaMessage, err := m.buildNewKafkaMessage(predictionLog)
-		if err != nil {
-			m.logger.Errorf("unable to build kafka message: %v", err)
-		}
-		err = m.producer.Produce(kafkaMessage, m.producer.Events())
-		if err != nil {
-			m.logger.Errorf("unable to produce kafka message: %v", err)
+		if rand.Float64() <= SamplingRate {
```
tiopramayudi, line 171: Nit: can we skip early so the publishing code not part of if branch

tiopramayudi (approved):
Left nit comments the rest is LGTM

PR #570, opened 2024-04-08
bug: skip requirement comparison for pip reqs that is not a package

khorshuheng:
# Description

While the SDK is processing the conda env, and additional merlin dependencies are required, such as merlin-pyfunc-server, the SDK will check the existing conda env to verify if the dependency is already present. If not, the additional requirement will be appended as part of the dependencies.

Unfortunately, this workflow assuems that every line under the `pip` configurations is a python package, which is not necessarily the case. For example, the user might specify trusted host or repository under the pip section.

# Modifications

Skip comparing pip requirements that is not a package while attempting to check if extra merlin dependencies are present in the conda.yaml file.

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/python/sdk/merlin/requirements.py
+++ b/python/sdk/merlin/requirements.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from typing import Any, List, Optional
 
 import yaml
-from packaging.requirements import Requirement
+from packaging.requirements import Requirement, InvalidRequirement
 
 _CONSTRAINTS_FILE_NAME = "constraints.txt"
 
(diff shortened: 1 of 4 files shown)


deadlycoconuts (approved):
Thanks for the quick fix! Just a tiny request, but could you include this new test case in the unit test examples [here](https://github.com/caraml-dev/merlin/tree/main/python/sdk/test/requirements)? You'll also need to add test case in the test params [here](https://github.com/caraml-dev/merlin/blob/main/python/sdk/test/requirements_test.py#L154). Thanks! 🙏

python/sdk/merlin/requirements.py
```diff
@@ -51,11 +51,16 @@ def process_conda_env(
...
-                    break
+                # Skip requirement comparison for pip reqs that is not a package, which can happen if the user specifies
+                # repository or trusted host as part of the pip file
+                try:
+                    pip_req_obj = Requirement(pip_req)
+                    additional_merlin_req_obj = Requirement(additional_merlin_req)
+                    if pip_req_obj.name.lower() == additional_merlin_req_obj.name.lower():
+                        exist = True
+                        break
+                except InvalidRequirement:
```
leonlnj, line 62: instead of doing a try-catch, would it be better to check and skip for entry starting with `--` (remove whitespace leading first)?

this will allow true invalid requirements to slip pass and defeat the purpose of any checking

leonlnj (approved):
thanks!

PR #573, opened 2024-04-18
bugfix: avoid logging non-200 response and don't publish mesasge when there is error

khorshuheng:
# Description

Currently, empty message might be published when there is non-200 response, or when there is error building the Kafka log. The observation publisher will crash as a result because the consumer currently doesn't handle the scenario when the Kafka message value is null.

# Modifications

- Prevent producing Kafka message when the response code is not 200, or when there is error building the Kafka log.
- Skip message with null message for the observation publisher

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/pkg/inference-logger/logger/mlobs_sink.go
+++ b/api/pkg/inference-logger/logger/mlobs_sink.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"net/http"
 
 	"google.golang.org/protobuf/types/known/timestamppb"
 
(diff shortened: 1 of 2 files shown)


api/pkg/inference-logger/logger/mlobs_sink.go
```diff
@@ -168,16 +168,18 @@ func (m *MLObsSink) buildNewKafkaMessage(predictionLog *upiv1.PredictionLog) (*k
 
 func (m *MLObsSink) Sink(rawLogEntries []*LogEntry) error {
 	for _, rawLogEntry := range rawLogEntries {
-		if rand.Float64() >= SamplingRate {
+		if rawLogEntry.ResponsePayload.StatusCode != 200 || rand.Float64() >= SamplingRate {
```
tiopramayudi, line 171: nit: better to use `http.StatusOK`

tiopramayudi (approved):
left one comment the rest is LGTM! Thanks @khorshuheng

PR #574, opened 2024-04-19
Add Image Building API

ariefrahmansyah:
# Description

The image building process of the Merlin models and batch prediction jobs are tightly coupled with the serving flow. If the image-building failed, the serving would fail too.

This PR introduces a new Image Building API to trigger the image building process programmatically and separately from the serving flow.

Note that the Swagger, clients, and Python SDK update will be on separate PRs.

# Modifications

1. Add Image Building API which calls the Image Building Service layer
2. Add Image Building Service layer which calls the pkg/image_builder.go

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
NONE
```


api/service/version_image_service_test.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	tests := []struct {
+		name         string
+		imageBuilder func() *mocks.ImageBuilder
+		args         args
+		want         models.VersionImage
+		wantErr      bool
+	}{
+		{
+			name: "success",
```
deadlycoconuts, line 29: Hmm should we also include some additional simple test cases to ensure that the correct image builder (or no image builder) gets returned when `s.getImageBuilder(model.Type)` is called? This might mean having to set up two different mocked image builders and then checking that `GetVersionImage` gets called on the right mocked image builder 🤔 

Another test case I can think of is when `imageBuilder.GetVersionImage(ctx, model.Project, model, version)` returns a image that is still being built.

api/service/version_image_service_test.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	tests := []struct {
+		name         string
+		imageBuilder func() *mocks.ImageBuilder
+		args         args
+		want         string
+		wantErr      bool
+	}{
+		{
+			name: "success",
```
deadlycoconuts, line 96: One last comment on tests 😅😶 Can we have a tiny test case when `imageBuilder.BuildImage` returns an error? 🙏

deadlycoconuts (approved):
Thanks a lot for this really easy to understand PR @ariefrahmansyah! 🚀 I just left a few comments/suggestions for some cosmetic changes but otherwise everything looks good!

api/service/version_image_service.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	imageRef, err := imageBuilder.BuildImage(ctx, model.Project, model, version, options.ResourceRequest)
+	if err != nil {
+		return "", fmt.Errorf("error building image: %w", err)
+	}
+
+	return imageRef, nil
```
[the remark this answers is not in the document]
ariefrahmansyah: Good question. I don't foresee immediate use of this but I'll keep it because I want versionImageService.BuildImage to return what pkg/imagebuilder.BuildImage return.

PR #575, opened 2024-04-22
Image Building Clients & SDK

ariefrahmansyah:
# Description

Continuation of PR #574:

1. Rework Get Image API to include image building job status and return termination message from job and/or pod.
2. Generate client code & add build_image() function in Merlin SDK.

# Modifications

1. Update swagger.yaml
2. Generate Golang & Python clients
3. Add build_image() function in Merlin Python SDK
4. Modify ImageBuilder interface so that
   a. BuildImage() now accepts `backoffLimit *int32`. During testing, when the image building job got retried, there's a short time when job status is failed hence the GetImageBuildingJobStatus() could return during this retry. Merlin SDK might be terminated here even though the image building job is still retrying. To avoid false negative, I set the backoff limit default value to 0 if image building is triggered from merlin SDK.
   b. Add new GetImageBuildingJobStatus()

# Tests

- [x] Build image via Python SDK

# Checklist
- [x] Added PR label
- [x] Tested locally
- [x] Updated documentation

# Release Notes

```release-note
Add build_image() in SDK
```


swagger.yaml
```diff
@@ -545,6 +547,60 @@ paths:
...
+        in: path
+        required: true
+        schema:
+          type: integer
+      - name: version_id
+        in: path
+        required: true
+        schema:
+          type: integer
+      responses:
```
deadlycoconuts, line 566: Should we also include the a `400` BadRequest (https://github.com/caraml-dev/merlin/blob/b5733e50dec330314f0cf30c02a4fdccf475e903/api/api/response.go#L118) error when the model type is invalid, and a `500` (https://github.com/caraml-dev/merlin/blob/b5733e50dec330314f0cf30c02a4fdccf475e903/api/api/response.go#L127) error when the appropriate image builder cannot be found?

swagger.yaml
```diff
@@ -545,6 +547,60 @@ paths:
...
+        required: true
+        schema:
+          type: integer
+      requestBody:
+        content:
+          "*/*":
+            schema:
+              "$ref": "#/components/schemas/BuildImageOptions"
+        required: false
+      responses:
```
deadlycoconuts, line 597: Similar to the comment above, should we also include the a `400` BadRequest (https://github.com/caraml-dev/merlin/blob/b5733e50dec330314f0cf30c02a4fdccf475e903/api/api/response.go#L118) error when the model type or request body is invalid?

swagger.yaml
```diff
@@ -545,6 +547,60 @@ paths:
...
+          type: integer
+      responses:
+        '200':
+          description: OK
+          content:
+            "*/*":
+              schema:
+                "$ref": "#/components/schemas/VersionImage"
+        '404':
+          description: Version with given `version_id` not found
```
deadlycoconuts, line 574: This error also gets returned when `model_id` cannot be found.

python/sdk/merlin/model.py
```diff
@@ -1115,6 +1105,62 @@ def list_endpoint(self) -> List[VersionEndpoint]:
...
+            body=options,
+        )
+
+        bar = pyprind.ProgBar(
+            100,
+            track_time=True,
+            title=f"Building Docker image for model {self.model.name} version {self.id}",
+        )
+
+        while True:
```
deadlycoconuts, line 1140: Wait, if the image building job fails or gets interrupted somehow, would this loop never end (until 10s * 100 iterations or around 16/17 minutes later)? :O

deadlycoconuts (approved):
Overall everything looks good! Thanks a lot for this PR! 🚀

My main concern's with how users are going to get a more 'up-to-date' signal that an image building job has failed (or has been interrupted) instead of waiting till the polling timeout happens. That said, maybe it's really impossible to do this without changing the API server to manage some state to store the status of all image building jobs, or to check all the running image building-related goroutines 🤔 But oh well if that's the case it's not a big problem either; was just wondering if there's a way around this.

python/sdk/merlin/model.py
ariefrahmansyah: Yes, because the current implementation is not storing any state on the DB, we don't have any reliable source for image existence other than the image repository itself. We depends on the remote repository for the image status.

If we want to have a reliable source about the image building status, I think we should have a new table for this (I don't think we can reuse the deployments table as it has endpoint_status which is specifically used for the version endpoint).

ariefrahmansyah: As discussed offline, I've updated the Get Image API to get image building k8s job as well so that this poll will check the `image_building_job_status` whether to continue polling or break.

api/api/version_image_api_test.go
```diff
@@ -64,6 +65,103 @@ func TestVersionImageController_GetImage(t *testing.T) {
...
+			versionImageService: func() *mocks.VersionImageService {
+				svc := &mocks.VersionImageService{}
+				svc.On("GetImage", mock.Anything, mock.Anything, mock.Anything).
+					Return(models.VersionImage{
+						ProjectID: 1,
+						ModelID:   1,
+						VersionID: 1,
+						ImageRef:  "ghcr.io/caraml-dev/project-model:1",
+						Exists:    true,
+						JobStatus: models.ImageBuildingJobStatusUnknown,
```
leonlnj, line 104: unknown is returned when job is missing?

api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -718,3 +719,27 @@ func (c *imageBuilder) createKanikoJobSpec(
...
 	return job, nil
 }
+
+func (c *imageBuilder) GetImageBuildingJobStatus(ctx context.Context, project mlp.Project, model *models.Model, version *models.Version) (models.ImageBuildingJobStatus, error) {
+	kanikoJobName := c.nameGenerator.generateBuilderJobName(project, model, version)
+
+	jobClient := c.kubeClient.BatchV1().Jobs(c.config.BuildNamespace)
+	job, err := jobClient.Get(ctx, kanikoJobName, metav1.GetOptions{})
+	if err != nil {
+		return models.ImageBuildingJobStatusUnknown, err
```
leonlnj, line 729: what if the job is in pending state? Not started and before active. Or that is covered under active?

leonlnj: just bit odd too that when there is error its unknown status, when there is no error its also unknown status

api/service/version_image_service.go
```diff
@@ -43,6 +44,13 @@ func (s *versionImageService) GetImage(ctx context.Context, model *models.Model,
 	}
 
 	versionImage := imageBuilder.GetVersionImage(ctx, model.Project, model, version)
+
+	jobStatus, err := imageBuilder.GetImageBuildingJobStatus(ctx, model.Project, model, version)
+	if err != nil {
+		log.Errorf("error getting image building job status: %v", err)
```
leonlnj, line 50: in the event of dependencies conflict, this will not be propagated via the sdk right?

api/api/version_image_api_test.go
ariefrahmansyah: Yes, correct.

api/pkg/imagebuilder/imagebuilder.go
ariefrahmansyah: -> what if the job is in pending state? Not started and before active. Or that is covered under active?
It's covered under active. Active is the number of pending and running pods. Ref: https://pkg.go.dev/k8s.io/api@v0.29.2/batch/v1#JobStatus

api/pkg/imagebuilder/imagebuilder_test.go
```diff
@@ -1867,3 +1868,190 @@ func Test_imageBuilder_getHashedModelDependenciesUrl(t *testing.T) {
...
+				return true, &v1.PodList{
+					Items: []v1.Pod{*pod},
+				}, nil
+			},
+			want: models.ImageBuildingJobStatus{
+				State: models.ImageBuildingJobStateFailed,
+				Message: `Error
+
+Job conditions:
+┌───────────────────────────────┬────────┬──────────────────────┬─────────────────────────────────────────────┐
```
leonlnj, line 2000: i dont know if we really want to validate the table 😅  can we just check that `Job has reached the specified backoff limit` and `CondaEnvException: Pip failed` exist for example

api/queue/work/model_service_deployment.go
```diff
@@ -215,7 +215,7 @@ func (depl *ModelServiceDeployment) generateModelOptions(ctx context.Context, mo
 	modelOpt := &models.ModelOption{}
 	switch model.Type {
 	case models.ModelTypePyFunc, models.ModelTypePyFuncV3:
-		imageRef, err := depl.ImageBuilder.BuildImage(ctx, model.Project, model, version, endpoint.ImageBuilderResourceRequest)
+		imageRef, err := depl.ImageBuilder.BuildImage(ctx, model.Project, model, version, endpoint.ImageBuilderResourceRequest, nil)
```
leonlnj, line 218: whats the motivation behind adding backofflimit when all the caller are going to pass a nil?

leonlnj (approved):
some small question but overall LGTM, thanks! feel free to merge when done

api/queue/work/model_service_deployment.go
ariefrahmansyah: Passing nil there for backward compatibility, so if they pass nil, Merlin will use the default value from the configuration (https://github.com/caraml-dev/merlin/blob/main/api/config/config.go#L225).

If SDK specifies a number, it will get propagated to the API and kaniko job spec.

For more background, the original reason we added this retry for the image building job is to retry when the pip installation failed due to an issue that can be recovered (e.g. network error).

api/pkg/imagebuilder/imagebuilder_test.go
ariefrahmansyah: We can, but we will need to modify the assertion function to use some string comparison function. I'll keep it as is for now.

PR #576, opened 2024-04-29
fix(api): Fix autoscaler target string parsing to not drop decimal numbers

deadlycoconuts:
# Description
As of present, users who have set a concurrency target=1 for their deployments and would like them to scale up even earlier (i.e. before the actual concurrency reaches 1) are unable to do so by simply setting the target as any float < 1 (e.g. 0.7), as the Merlin API server rounds it to the nearest integer. This was done so because the autoscaling target annotation in the Knative service manifest `autoscaling.knative.dev/target`, _**seems to require**_ an [INTEGER](https://knative.dev/docs/serving/autoscaling/concurrency/#soft-limit). 

However, when we tried setting the value of the aforementioned annotation with a variety of values (including float values < 1 and 0), we noticed that float values < 1 surprisingly work and correctly modifies the `autoscaler_target_concurrency_per_pod` metric (this values is calculated by the formula `target_concurrency = user_specified_target * target_utilization (default 70)` generated. On the other hand, setting values such as 0 or anything < 0.01 will cause the Knative admission webhook to return an [error](https://github.com/knative/serving/blob/b1c026628485c16c8a9191553b6adde280f2d2b0/pkg/apis/autoscaling/annotation_validation.go#L115) like the one shown below:

```
Warning  InternalError  21s (x16 over 3m5s)  v1beta1Controllers  fails to reconcile predictor: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: target 0 should be at least 0.01: spec.template.metadata.annotations.autoscaling.knative.dev/target
```

As such, this PR makes a tiny change to the concurrency (only) target annotation value, by allowing values up to 2 decimal places (as such values from 0.01 - 0.99 are now allowed) to be set. Values below 0.01 after rounding will cause an error to be returned - this prevents the deployment from being stuck in a pending state because the Knative webhook rejects the annotation's value.

# Modifications
- `api/cluster/resource/utils.go` - Changing of the rounding of the concurrency target annotation value from the closest integer to the closest 2 decimal places

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
Users that have previously set an autoscaling concurrency target of 0.01 - 1 will now experience earlier scaling up of their deployments and can no longer set 0 as an autoscaling concurrency target.
```

### DIFF
--- a/api/cluster/resource/templater_gpu_test.go
+++ b/api/cluster/resource/templater_gpu_test.go
@@ -1211,7 +1211,7 @@ func TestCreateInferenceServiceSpecWithGPU(t *testing.T) {
 						knautoscaling.InitialScaleAnnotationKey:               fmt.Sprint(testPredictorScale),
 						knautoscaling.ClassAnnotationKey:                      knautoscaling.KPA,
 						knautoscaling.MetricAnnotationKey:                     knautoscaling.Concurrency,
-						knautoscaling.TargetAnnotationKey:                     "2",
+						knautoscaling.TargetAnnotationKey:                     "2.00",
 					},
 					Labels: map[string]string{
 						"gojek.com/app":          modelSvc.Metadata.App,
--- a/api/cluster/resource/templater_test.go
+++ b/api/cluster/resource/templater_test.go
@@ -1596,6 +1596,28 @@ func TestCreateInferenceServiceSpec(t *testing.T) {
 			},
 		},
 		{
+			name: "serverless deployment using concurrency autoscaling where target after rounding < 0.01",
+			modelSvc: &models.Service{
+				Name:           modelSvc.Name,
+				ModelName:      modelSvc.ModelName,
+				ModelVersion:   modelSvc.ModelVersion,
+				Namespace:      project.Name,
+				ArtifactURI:    modelSvc.ArtifactURI,
+				Type:           models.ModelTypeTensorflow,
+				Options:        &models.ModelOption{},
+				Metadata:       modelSvc.Metadata,
+				DeploymentMode: deployment.ServerlessDeploymentMode,
+				AutoscalingPolicy: &autoscaling.AutoscalingPolicy{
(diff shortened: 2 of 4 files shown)


api/cluster/resource/templater.go
```diff
@@ -643,6 +643,27 @@ func (t *InferenceServiceTemplater) createAnnotations(modelService *models.Servi
...
+			// in order to allow the deployment to scale up earlier than when the default target utilization (70) and
+			// user-defined target = 1 is set. Note however, that this does not change the actual 'soft limit' on the
+			// number of requests that each replica can serve, which remains as 1.
+			//
+			// For more details, see: https://knative.dev/docs/serving/autoscaling/concurrency/#target-utilization
+			if modelService.AutoscalingPolicy.MetricsType == autoscaling.Concurrency &&
+				modelService.AutoscalingPolicy.TargetValue < 1 {
+				annotations[knautoscaling.TargetUtilizationPercentageKey] = fmt.Sprintf("%.0f",
+					70*modelService.AutoscalingPolicy.TargetValue)
+			}
```
ariefrahmansyah, line 666: Why don't we just use the value given by the user? I still don't understand the reasoning behind multiplying the `modelService.AutoscalingPolicy.TargetValue` with the default value here (70). It might lead to confusion as to why the value propagated to the isvc is different from the one provided by the user.

leonlnj: yea initially we came up with using 1 * the target % to make it below 1. Only after, he found out that float actually works despite the docs 😅

deadlycoconuts: Let me reword the description of this PR to make this clearer.

deadlycoconuts: Done!

leonlnj (approved):
thanks!

ariefrahmansyah (approved):
lgtm. thanks zi yi!

PR #583, opened 2024-05-16
replace tornado with sanic

leonlnj:
Just a POC, only diff is that MLObs publisher init and `SIGTERM` SIGINT` implementation are not implemented


python/pyfunc-server/pyfuncserver/protocol/rest/server_sanic.py
```diff
@@ -0,0 +1,93 @@
...
+
+        self.config = config
+        self.workers = config.workers
+        self.model_manifest = config.model_manifest
+        self.http_port = config.http_port
+        self.metrics_registry = metrics_registry
+        self.registered_models: dict = {}
+        self.register_model(model)
+
+        Sanic.start_method = "fork"
```
leonlnj, line 30: in Tornado, `self._http_server.start(self.workers)` will use `fork` mechanism

image

python/pyfunc-server/pyfuncserver/protocol/rest/server_sanic.py
```diff
@@ -0,0 +1,93 @@
...
+        self.app = Sanic("pyfunc")
+        self.app.update_config({"ACCESS_LOG": False})
```
leonlnj, line 32: turn access log off for better performance - ref https://sanic.dev/en/guide/running/running.html#performance-considerations

python/pyfunc-server/pyfuncserver/protocol/rest/server_sanic.py
```diff
@@ -0,0 +1,85 @@
...
+        self.config = config
+        self.workers = config.workers
+        self.model_manifest = config.model_manifest
+        self.http_port = config.http_port
+        self.metrics_registry = metrics_registry
+        self.registered_models: dict = {}
+        self.register_model(model)
+
+        Sanic.start_method = "fork"
+        self.app = Sanic("pyfunc", dumps=dumps, loads=loads)
```
leonlnj, line 31: this set orjson as default for sanic (essentially doing the same as current imp)

python/pyfunc-server/pyfuncserver/protocol/rest/server_sanic.py
```diff
@@ -0,0 +1,85 @@
...
+        self.app.update_config({"ACCESS_LOG": False})
+        self.setup_routes()
+
+    def setup_routes(self):
+        @self.app.post(r"/v1/models/<model:([a-zA-Z0-9_-]+):predict>")
+        async def predict_handler(request: Request, model):
+            model = self.get_model(model)
+
+            output = model.predict(request.json, headers=request.headers)
```
leonlnj, line 40: the current `validate` function which uses orjson, is called in request.json as well. The `loads` method is set in app above.

image

python/pyfunc-server/pyfuncserver/__main__.py
```diff
@@ -37,16 +35,10 @@
...
 
-logging.getLogger("tornado.access").disabled = True
-
-
 def main():
     config = Config(args.model_dir)
     logging.basicConfig(level=config.log_level)
 
-    # use uvloop as the event loop
-    asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
```
leonlnj, line 48: sanic use exactly this by default as well

```
    if not isinstance(asyncio.get_event_loop_policy(), uvloop.EventLoopPolicy):
        asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
```

PR #584, opened 2024-05-16
Fix PDB and VS not deleted

ariefrahmansyah:
# Description

This PR solves two bugs:

1. Previous PDB not deleted when redeployed
   a. The previous implementation only used Apply to create a new revision's PDB without deleting the previous revision's PDB.
  b. This PR uses Patch instead of Apply so it will only use one PDB to every revision
2. Virtual service not deleted when undeployed
   a. This issue happens due to condition logic that would only delete VS with revision > 1. Removing this condition solve the issue.
   b. This condition was created to maintain backward compatibility to VS without revision, but the implementation is faulty.

# Tests

# Checklist
- [ ] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/cluster/controller.go
+++ b/api/cluster/controller.go
@@ -251,6 +251,7 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 	}
 
 	if c.deploymentConfig.PodDisruptionBudget.Enabled {
+		// Create / update pdb
 		pdbs := createPodDisruptionBudgets(modelService, c.deploymentConfig.PodDisruptionBudget)
 		if err := c.deployPodDisruptionBudgets(ctx, pdbs); err != nil {
 			log.Errorf("unable to create pdb: %v", err)
@@ -327,12 +328,15 @@ func (c *controller) Delete(ctx context.Context, modelService *models.Service) (
 		}
 	}
 
-	if modelService.RevisionID > 1 {
-		vsName := fmt.Sprintf("%s-%s-%s", modelService.ModelName, modelService.ModelVersion, models.VirtualServiceComponentType)
-		if err := c.deleteVirtualService(ctx, vsName, modelService.Namespace); err != nil {
-			log.Errorf("unable to delete virtual service %v", err)
-			return nil, ErrUnableToDeleteVirtualService
-		}
+	vsCfg, err := NewVirtualService(modelService, "")
+	if err != nil {
+		log.Errorf("unable to initialize virtual service builder: %v", err)
+		return nil, errors.Wrapf(err, fmt.Sprintf("%v", ErrUnableToDeleteVirtualService))
+	}
+
+	if err := c.deleteVirtualService(ctx, vsCfg); err != nil {
+		log.Errorf("unable to delete virtual service %v", err)
+		return nil, ErrUnableToDeleteVirtualService
 	}
 
 	return modelService, nil
--- a/api/cluster/controller.go
+++ b/api/cluster/controller.go
@@ -252,7 +252,7 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 
 	if c.deploymentConfig.PodDisruptionBudget.Enabled {
 		// Create / update pdb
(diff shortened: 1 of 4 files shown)


api/cluster/controller.go
```diff
@@ -251,6 +251,7 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 	}
 
 	if c.deploymentConfig.PodDisruptionBudget.Enabled {
+		// Create / update pdb
```
leonlnj, line 254: instead of this maybe change `createPodDisruptionBudgets` to `generatePDBSpecs` ? since nothing is created until deployed

api/cluster/controller.go
```diff
@@ -327,12 +328,15 @@ func (c *controller) Delete(ctx context.Context, modelService *models.Service) (
 		}
 	}
 
-	if modelService.RevisionID > 1 {
-		vsName := fmt.Sprintf("%s-%s-%s", modelService.ModelName, modelService.ModelVersion, models.VirtualServiceComponentType)
-		if err := c.deleteVirtualService(ctx, vsName, modelService.Namespace); err != nil {
-			log.Errorf("unable to delete virtual service %v", err)
-			return nil, ErrUnableToDeleteVirtualService
-		}
+	vsCfg, err := NewVirtualService(modelService, "")
```
leonlnj, line 331: when does modelService.RevisionID <= 0 happen? or was the check useless to begin with?

ariefrahmansyah: When we haven't introduced model version revision features (prior to Merlin 0.34.0).

Yeah, the check was faulty resulting in Merlin not deleting:
1. VS of version endpoints without revision (the ones deployed by Merlin<0.34.0)
2. VS of version endpoints with revision 1 (the first deployment of version endpoint)
And will only delete VS of version endpoints with revision > 1.

api/cluster/pdb.go
```diff
@@ -113,11 +125,15 @@ func (c *controller) deployPodDisruptionBudget(ctx context.Context, pdb *PodDisr
...
+	pdbJSON, err := json.Marshal(pdbSpec)
+	if err != nil {
+		return err
+	}
+
+	forceEnabled := true
 
-	_, err = c.policyClient.PodDisruptionBudgets(pdb.Namespace).Apply(ctx, pdbCfg, metav1.ApplyOptions{FieldManager: "application/apply-patch"})
+	_, err = c.policyClient.PodDisruptionBudgets(pdb.Namespace).
+		Patch(ctx, pdb.Name, types.ApplyPatchType, pdbJSON, metav1.PatchOptions{FieldManager: "application/apply-patch", Force: &forceEnabled})
```
leonlnj, line 136: why do we force patch instead of just apply, i thought apply would do the same? I might be missing something here

api/cluster/pdb.go
```diff
@@ -29,39 +30,50 @@ func NewPodDisruptionBudget(modelService *models.Service, componentType string,
 	labels["serving.kserve.io/inferenceservice"] = modelService.Name
 
 	return &PodDisruptionBudget{
-		Name:                     fmt.Sprintf("%s-%s-%s", modelService.Name, componentType, models.PDBComponentType),
+		Name:                     fmt.Sprintf("%s-%s-%s-%s", modelService.ModelName, modelService.ModelVersion, componentType, models.PDBComponentType),
```
leonlnj, line 33: to confirm my understanding, modelService.Name was `{modelName}-{modelVersion}-{knativeRevision}`, this essentially strip the knative revision so we can apply the patch for a pdb to be reuse?

leonlnj:
was thinking if we can patch isvc instead of create-delete to keep how we manage the resource the same fashion, then i realized the isvc patch does not have `meta.PatchOptions` 😅 

image


image

api/cluster/pdb.go
ariefrahmansyah: it's not knativeRevision but meriln's model version revision.
modelService.Name is {modelName}-{versionID}-{revisionPrefix}{revisionID}

PR #586, opened 2024-05-17
feat(api,ui,sdk): Make CPU limits configurable

deadlycoconuts:
# Description
As of present, users are not able to configure the CPU limits of the pods in which Merlin models and transformers are deployed in - they are instead determined automatically on the platform-level (Merlin API server). Depending on how the API server has been configured, one of the following happens:
- the CPU limit of a model is set as its CPU request value, multiplied by a [scaling factor](https://github.com/caraml-dev/merlin/blob/f1ebe099ea168988b365ee72ce08543b127826e1/api/config/config.go#L364) (e.g. 2 CPU * 1.5) **or,**
    - Note that this is the existing way memory limits are automatically set by the Merlin API server
- the CPU limit is left unset 
    - Note that because KServe does not currently allow CPU limits to be completely unset, the Merlin API server instead sets an [arbitrary value ](https://github.com/caraml-dev/merlin/blob/f1ebe099ea168988b365ee72ce08543b127826e1/api/config/config.go#L363)(ideally one that is very big) as the CPU limit instead

This PR introduces a new workflow which would allow users to instead override the platform-level CPU limits (described in the paragraph above) set on a model. This workflow is available via the UI, SDK and by extension, directly calling the API endpoint of the API server. 

UI:
Screenshot 2024-05-24 at 2 13 46 PM

Screenshot 2024-05-24 at 2 23 42 PM

SDK:
```python
merlin.deploy(
    version_1,
    resource_request=merlin.ResourceRequest(
        min_replica=0,
        max_replica=0,
        cpu_request="0.5",
        cpu_limit="2",
        memory_request="1Gi",
    ),
)
```

In addition, this PR adds a new configuration, `DefaultEnvVarsWithoutCPULimits`, which is a list of env vars that automatically get added to all Merlin models and transformers when CPU limits are not set. This allows the Merlin API server's operators to set env vars platform-wide that can potentially improve these deployments' performance, e.g. env vars involving concurrency. 

# Modifications
- `api/cluster/resource/templater.go` - Refactoring of templater methods to set default env vars when cpu limits are not explicitly set and when the cpu limit scaling factor is set as 0
- `api/config/config.go` - Addition of the new field `DefaultEnvVarsWithoutCPULimits`
- `api/config/config_test.go` - Addition of a new unit test to test the parsing of configs from .yaml files 
- `docs/user/templates/model_deployment/01_deploying_a_model_version.md` - Addition of docs to demonstrate how the platform-level CPU limits can be overriden
- `python/sdk/merlin/resource_request.py` - Addition of a new cpu limit field to the resource request class
- `ui/src/pages/version/components/forms/components/CPULimitsFormGroup.js` - Addition of a new form group to allow cpu limits to be specified on the UI

# Tests
- [x] Deploying existing models (and transformers) with and without CPU limits set

# Checklist
- [x] Added PR label
- [x] Tested locally
- [x] Updated documentation

# Release Notes

```release-note
NONE
```


api/cluster/resource/templater.go
```diff
@@ -910,3 +931,11 @@ func (t *InferenceServiceTemplater) applyDefaults(service *models.Service) {
 		}
 	}
 }
+
+func ParseEnvVars(envVars []corev1.EnvVar) models.EnvVars {
+	var parsedEnvVars models.EnvVars
+	for _, envVar := range envVars {
+		parsedEnvVars = append(parsedEnvVars, models.EnvVar{Name: envVar.Name, Value: envVar.Value})
+	}
+	return parsedEnvVars
```
deadlycoconuts, line 940: This helper function converts the default env vars from the `corev1.EnvVar` type (from the k8s library) to the `models.EnvVar` type (defined and used only in the Merlin API server) so that the existing helper method to merge env vars `models.MergeEnvVars` can be easily reused.

api/cluster/resource/templater_test.go
```diff
@@ -4482,3 +4841,39 @@ func createPyFuncPublisherEnvVars(svc *models.Service, pyfuncPublisher config.Py
 	}
 	return envVars
 }
+
+func sortInferenceServiceSpecEnvVars(isvc kservev1beta1.InferenceServiceSpec) {
```
deadlycoconuts, line 4845: This helper function is only used in this test in order to sort out all the env vars in an inference service spec, so that any method performing a comparison on 2 inference services does not throw errors caused by differences in the order of env vars.

api/config/config_test.go
```diff
@@ -311,3 +316,318 @@ func TestStandardTransformerConfig_ToFeastStorageConfigs(t *testing.T) {
...
+			err = os.Setenv(key, val)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func TestLoad(t *testing.T) {
```
deadlycoconuts, line 335: I added this unit test to test the parsing of configs from .yaml files to the actual Go config struct because I realised I left this out the last time I refactored the Merlin API server to consume configs from .yaml files.

api/models/env_var.go
```diff
@@ -69,19 +69,21 @@ func (evs EnvVars) ToKubernetesEnvVars() []v1.EnvVar {
...
-	}
-	for _, add := range right {
-		if index, exist := envIndexMap[add.Name]; exist {
-			left[index].Value = add.Value
-		} else {
-			left = append(left, add)
+// MergeEnvVars merges multiple sets of environment variables and return the merging result.
+// All the EnvVars passed as arguments will be not mutated.
+// EnvVars to the right have higher precedence.
+func MergeEnvVars(left EnvVars, rightEnvVars ...EnvVars) EnvVars {
```
deadlycoconuts, line 75: I extended this helper method a little so that it now merges multiple lists of env vars instead of just 2 lists (this helps with the unit tests where we're chaining this function multiple times in order to merge multiple lists of env vars). The same rules of priority apply.

ui/src/pages/version/components/forms/components/AutoscalingPolicyFormGroup.js
```diff
@@ -109,27 +109,33 @@ export const AutoscalingPolicyFormGroup = ({
           Autoscaling Policy determines the condition for increasing or
           decreasing number of replicas.
         </Fragment>
-      }>
+      }
+      fullWidth
```
deadlycoconuts, line 113: I've added all this `fullWidth` props everywhere in order to keep the width of all the components in the page consistent when the page is displayed on super large screens.

ui/src/pages/version/components/forms/components/ImageBuilderSection.js
```diff
@@ -1,64 +1,72 @@
...
 
-        <EuiFlexItem>
+      <EuiFlexItem>
         <EuiFormRow
-            label={
+          label={
             <FormLabelWithToolTip
-                label="Image Builder Memory *"
-                content="To set a higher Mem above platform default for Image Building"
+              label="Image Builder Memory"
```
deadlycoconuts, line 51: I removed the asterisk `*` because it doesn't seem to suggest anything (this field isn't compulsory or anything).

ui/src/pages/version/components/forms/components/ImageBuilderSection.js
```diff
@@ -1,64 +1,72 @@
...
+  const {onChange} = useOnChangeHandler(onChangeHandler);
 
-    const { onChange } = useOnChangeHandler(onChangeHandler);
-
-    return (
+  return (
+    <>
+      <EuiTitle size="xs">
+        <h4>{"Image Building Resources"}</h4>
+      </EuiTitle>
```
deadlycoconuts, line 22: I moved the title into this component (the title used to be in the parent component).

.github/workflows/merlin.yml
```diff
@@ -141,6 +141,7 @@ jobs:
           name: sdk-test-${{ matrix.python-version }}
           token: ${{ secrets.CODECOV_TOKEN }}
           working-directory: ./python/sdk
+          codecov_yml_path: ../../.github/workflows/codecov-config/codecov.yml
```
deadlycoconuts, line 144: I've added the codecov threshold setting here @ariefrahmansyah. Thanks for the digging this! 

The configs are inspired from this PR here: https://github.com/kensho-technologies/graphql-compiler/pull/806/files; I removed the other configs because `target`'s default is already `auto` and `base` is supposedly deprecated according to the [docs](https://docs.codecov.com/docs/commit-status#threshold). Note that the paths are slightly different because of the different `working-directory`-s.

api/cluster/resource/templater.go
```diff
@@ -392,28 +401,36 @@ func (t *InferenceServiceTemplater) createTransformerSpec(
...
-			transformer.ResourceRequest.CPURequest, t.deploymentConfig.UserContainerCPULimitRequestFactor,
-		)
-	} else {
-		// TODO: Remove this else-block when KServe finally allows default CPU limits to be removed
-		var err error
-		limits[corev1.ResourceCPU], err = resource.ParseQuantity(t.deploymentConfig.UserContainerCPUDefaultLimit)
-		if err != nil {
-			return nil, err
+	// Set cpu resource limits automatically if they have not been set
+	if transformer.ResourceRequest.CPULimit == nil || transformer.ResourceRequest.CPULimit.IsZero() {
```
leonlnj, line 409: extract this portion out as a func since its the same logic as the one for predictor

ui/src/pages/version/components/forms/components/CPULimitsFormGroup.js
```diff
@@ -0,0 +1,46 @@
...
+  errors = {},
+}) => {
+  const {onChange} = useOnChangeHandler(onChangeHandler);
+
+  return (
+    <EuiDescribedFormGroup
+      title={<p>CPU Limit</p>}
+      description={
+        <Fragment>
+          Use this field to override the platform-level default CPU limit.
```
leonlnj, line 18: this we can omit `Use this field`

ui/src/components/ResourcesConfigTable.js
```diff
@@ -33,6 +34,12 @@ export const ResourcesConfigTable = ({
       title: "CPU Request",
       description: cpu_request,
     },
+    ...(cpu_limit !== undefined && cpu_limit !== "0" && cpu_limit !== "") ? [
```
leonlnj, line 37: if user create with `cpu_limit=0`, the platform set it to `10`, i assume they will see 10 here?

leonlnj (approved):
thanks, LGTM.

api/cluster/resource/templater_test.go
```diff
@@ -722,6 +751,66 @@ func TestCreateInferenceServiceSpec(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "pyfunc spec with cpu limits specified",
```
leonlnj, line 755: not expecting changes here but was thinking how can we make this more concise

1. Create default case, the test input should overwrite/merge into the default. Likewise for the output. 
2. Make assert func part of the test input, check for area of interest only (in this cause the assert func checks for cpu limits)

ui/src/components/ResourcesConfigTable.js
deadlycoconuts: Nope they wouldn't, because when the platform controls the CPU limit, regardless of the value that gets automatically sets, it isn't saved in the DB. Hence when the UI retrieves the config values, the `cpu_limit` value would be empty.

ui/src/pages/version/components/forms/components/CPULimitsFormGroup.js
deadlycoconuts: Done!

api/cluster/resource/templater.go
deadlycoconuts: Okay, I've just extracted it into a new function `getResourceLimitsAndEnvVars`.

api/cluster/resource/templater_test.go
deadlycoconuts: > Make assert func part of the test input, check for area of interest only (in this cause the assert func checks for cpu limits)

Yeah this sounds like a great idea, I was thinking of this too, but I'm not too sure if that'll mean we'll leave the door open to other unexpected bugs, e.g. changing the cpu limit stuff causes a totally unlinked change somewhere else in the min/max replicas. But yeah point 1 is probably something necessary that would cut down most of the lines in this file.

PR #588, opened 2024-05-22
fix: Add resource request's max replica validation

ariefrahmansyah:
# Description

This PR adds validation on resource request max replicas value to avoid the 0 value. The validations were added on API and SDK.

This is to avoid the bug if max replicas = 0 that will result in the no-upper limit of replicas for the inference service which could scaled up indefinitely till the entire cluster exhausted.

# Modifications

1. Add max replicas < 1 validation in API
2. Add max replicas < 1 validation in SDK

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
* merlin-sdk will raise an Exception if ResourceRequest object is initialized with max_repilcas=0.
```

### DIFF
--- a/python/sdk/merlin/resource_request.py
+++ b/python/sdk/merlin/resource_request.py
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from __future__ import annotations
+
 from typing import Optional
 
 import client
+
 
 class ResourceRequest:
     """
@@ -46,7 +48,7 @@ class ResourceRequest:
             cpu_request=response.cpu_request,
             memory_request=response.memory_request,
             gpu_request=response.gpu_request,
-            gpu_name=response.gpu_name
+            gpu_name=response.gpu_name,
         )
 
     @property
@@ -99,7 +101,10 @@ class ResourceRequest:
 
     def validate(self):
         if self._min_replica is None and self._max_replica is None:
-            return 
+            return
 
         if self._min_replica > self._max_replica:
             raise Exception("Min replica must be less or equal to max replica")
+
+        if self._max_replica == 0:
+            raise Exception("Max replica must be greater than 0")
--- a/python/sdk/merlin/resource_request.py
+++ b/python/sdk/merlin/resource_request.py
@@ -106,5 +106,5 @@ class ResourceRequest:
         if self._min_replica > self._max_replica:
             raise Exception("Min replica must be less or equal to max replica")
 
-        if self._max_replica == 0:
+        if self._max_replica < 1:
             raise Exception("Max replica must be greater than 0")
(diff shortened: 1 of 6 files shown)


python/sdk/merlin/resource_request.py
```diff
@@ -99,7 +101,10 @@ def gpu_name(self, gpu_name):
 
     def validate(self):
         if self._min_replica is None and self._max_replica is None:
-            return 
+            return
 
         if self._min_replica > self._max_replica:
```
tiopramayudi, line 106: is it possible for user to not set min replica and set max replica?

tiopramayudi (approved):
add one question, overall is LGTM

python/sdk/merlin/resource_request.py
ariefrahmansyah: Yes, it's possible and it will use the default environment's default resource request value.

tiopramayudi: for that case the below code will be failed right?
```
if self._min_replica > self._max_replica:
```
I assume set the default resource is set in server

ariefrahmansyah: > I assume set the default resource is set in server

Not really. The default resource setting happens in both API and SDK, but the value from SDK takes more precedence.

You can refer here: https://github.com/caraml-dev/merlin/blob/main/api/service/version_endpoint_service.go#L115-L120 and here: https://github.com/caraml-dev/merlin/blob/main/api/service/version_endpoint_service.go#L130-L133

Where for every deployment of model version (the first one or redeployment), the resource request supplied by SDK (or UI) will be used to override the default environment's resource request value.

ariefrahmansyah: I also realized that my first answer to your first question is wrong.

-> is it possible for user to not set min replica and set max replica?
Yes, it's possible. And it might not be using the default environment's default resource request value at all. Because the check in merlin-sdk is only check the whole ResourceRequest object not the max_replicas value: https://github.com/caraml-dev/merlin/blob/main/python/sdk/merlin/model.py#L1265

ariefrahmansyah: I've updated the PR so that in sdk's deploy() function it will fallback to the default environment's resource request min and max replicas if these values are None

tiopramayudi: can you refer me the one in sdk? i saw in [here](https://github.com/caraml-dev/merlin/blob/main/python/sdk/merlin/model.py#L1247) but the validation is conducted on different [object](https://github.com/caraml-dev/merlin/blob/main/python/sdk/merlin/model.py#L1266)

tiopramayudi: nvm it is already addressed

PR #589, opened 2024-05-22
cherry-pick: Add resource request's max replica validation (#588) for hotfix

ariefrahmansyah:
# Description

This PR adds validation on resource request max replicas value to avoid the 0 value. The validations were added on API and SDK.

This is to avoid the bug if max replicas = 0 that will result in the no-upper limit of replicas for the inference service which could scaled up indefinitely till the entire cluster exhausted.

# Modifications

1. Add max replicas < 1 validation in API
2. Add max replicas < 1 validation in SDK

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
* merlin-sdk will raise an Exception if ResourceRequest object is initialized with max_repilcas=0.
```

### DIFF
--- a/api/api/validator.go
+++ b/api/api/validator.go
@@ -44,9 +44,7 @@ var supportedObservabilityModelTypes = []string{
 	models.ModelTypeXgboost,
 }
 
-var (
-	ErrUnsupportedObservabilityModelType = errors.New("observability cannot be enabled not for this model type")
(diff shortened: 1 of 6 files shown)


tiopramayudi (approved):
LGTM

PR #591, opened 2024-06-12
Revamp batch prediction jobs UI

ariefrahmansyah:
# Description

# Modifications

# Tests

# Checklist
- [ ] Added PR label
- [ ] Added unit test, integration, and/or e2e tests
- [ ] Tested locally
- [ ] Updated documentation
- [ ] Update Swagger spec if the PR introduce API changes
- [ ] Regenerated Golang and Python client if the PR introduces API changes

# Release Notes

```release-note

```

PR #594, opened 2024-07-23
fix(ui): Fix yup validation bugs

deadlycoconuts:
# Description
Similar to https://github.com/caraml-dev/turing/pull/387, this PR to fixes broken validation schemas caused by the upgrade of the `yup` package in PR #591 - these changes can largely be grouped into the following:

- The `Schema.when` function has been [updated](https://github.com/jquense/yup?tab=readme-ov-file#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema), such that when the builder object (one with the fields `is`, `then`, and sometimes `otherwise` defined) is passed as an argument, the `then` field accepts ONLY functions of the following signature `(schema: Schema) => Schema` instead of `Schema` previously. 

  Existing schemas that do not follow this convention have been updated and turned into arrow functions. 
  
  Example:
  ```javascript
  // before
  some_field: yup.string().when("some_other_field", {
    is: "nop",
    then: yup.string().required("this is needed")
  }),

  // after
  some_field: yup.string().when("some_other_field", {
    is: "nop",
    then: (_) => yup.string().required("this is needed")
  }),
  ```
- When using the same `Schema.when` function but passing a function directly as an argument instead of the builder object, the signature of the expected function has also been updated from `(value, schema)=> Schema): Schema` to `(values: any[], schema) => Schema): Schema`.

  Existing schemas that do not follow this convention now have been updated to reflect the new expected array:

  Example:
  ```javascript
  // before
  some_field: yup.string().when("some_other_field", (some_other_field, schema) =>
    some_other_field ? yup.string().required("this is needed") : schema
  ),

  // after
  some_field: yup.string().when("some_other_field", ([some_other_field], schema) =>
    some_other_field ? yup.string().required("this is needed") : schema
  ),
  ```
## Unrelated Update
This PR also bumps up the version of the `caraml-dev/ui-lib` library to the latest version.

# Modifications
- UI bug fixes
- Update to `caraml-dev/ui-lib` library

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
NONE
```

### DIFF
--- a/ui/src/pages/version/components/forms/validation/schema.js
+++ b/ui/src/pages/version/components/forms/validation/schema.js
@@ -2,6 +2,7 @@ import * as yup from "yup";
 
 const cpuRequestRegex = /^(\d{1,3}(\.\d{1,3})?)$|^(\d{2,5}m)$/,
   memRequestRegex = /^\d+(Ei?|Pi?|Ti?|Gi?|Mi?|Ki?)?$/,
+  gpuRequestRegex= /^\d+$/,
   envVariableNameRegex = /^[a-z0-9_]*$/i,
   dockerImageRegex =
     /^([a-z0-9]+(?:[._-][a-z0-9]+)*(?::\d{2,5})?\/)?([a-z0-9]+(?:[._-][a-z0-9]+)*\/)*([a-z0-9]+(?:[._-][a-z0-9]+)*)(?::[a-z0-9]+(?:[._-][a-z0-9]+)*)?$/i;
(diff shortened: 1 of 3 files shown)


ui/src/pages/version/components/forms/validation/schema.js
```diff
@@ -19,9 +20,9 @@ const resourceRequestSchema = (maxAllowedReplica) => yup.object().shape({
   gpu_request: yup
     .string()
     .typeError("GPU value is required")
-    .when("gpu_resource_type", {
-      is: (v) => v !== undefined && v !== "None",
-      then: yup.string().required("GPU value is required"),
+    .when("gpu_name", {
+      is: (v) => v !== undefined,
+      then: (_) => yup.string().matches(gpuRequestRegex, { message:'Valid GPU value is required'}).required("GPU value is required"),
```
deadlycoconuts, line 25: This validation was never working in the first place since `None` is selected by default, and the way this was implemented allows the `None` string to go through as a valid value (when in reality it shouldn't). I've implemented a quick fix here to allow only integers but a better implementation would be to use a validation context containing the list of allowable GPU values (this comes from the environments context and can be filtered out after an environment is selected), and then check if the given value exists within that list. This however needs more time (it's not rocket science though) and is left out of this PR.

leonlnj (approved):

deadlycoconuts:
Thanks for the quick approval! Merging this now! 🚀

PR #595, opened 2024-07-23
fix merged env vars, new vars should overwrite the old one

bthari:
# Description

When user want to deploy and add the environment variables of Merlin model's endpoint, instead of overwriting the old environment variables with the new variables, both of the variables will be [appended instead](https://github.com/caraml-dev/merlin/blob/5b37909b8ee3972469f2fd2a4677596af78c7f99/api/service/version_endpoint_service.go#L235). E.g.

- Current endpoint's vars: `A = false`
- New endpoint's vars: `B = true`
- Final deployed endpoint's vars: `A = false, B = true`

These changes are intended to fix it, so when user updates the endpoint environment variables, it will overwrite the old environment variables.

# Modifications

Modify the `override` function, which is use to override left endpoint with right endpoint, so it will replace the environment variables of the left (old) version endpoint with right (new) version endpoint's environment variables.

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
deploying or redeploying model's endpoint with environment variables will now replace all old environment variables with new ones
```

### DIFF
--- a/api/service/version_endpoint_service.go
+++ b/api/service/version_endpoint_service.go
@@ -231,8 +231,8 @@ func (k *endpointService) override(left *models.VersionEndpoint, right *models.V
 
 	// override env vars
 	// Configure environment variables for Pyfunc model
-	if len(right.EnvVars) > 0 {
-		left.EnvVars = models.MergeEnvVars(left.EnvVars, right.EnvVars)
+	if right.EnvVars != nil {
+		left.EnvVars = right.EnvVars
 	}
 
 	// override protocol
--- a/api/service/version_endpoint_service_test.go
+++ b/api/service/version_endpoint_service_test.go
@@ -669,6 +669,12 @@ func TestDeployEndpoint(t *testing.T) {
 								CPURequest:    resource.MustParse("1"),
 								MemoryRequest: resource.MustParse("1Gi"),
 							},
+							EnvVars: models.EnvVars{
+								{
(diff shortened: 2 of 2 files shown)


deadlycoconuts (approved):
Thanks for this PR! 🥳 The changes LGTM! 🚀 I've only briefly looked at the the UI and SDK's code to see if they also send an empty list as the env vars variable (when no env vars are specified) instead of a `nil` value to the API server, and it seems like they do (this removes all existing env vars on the model). If you haven't already tried updating a model with 0 env vars specified via the UI and SDK, could you do so? 🙏🏼 

We'd also need to document your change as a user-facing change, since env vars are no longer being merged and are instead overwriting the old ones. Could you document this in your PR description under the 'Release Notes' section too? Thanks! 

On a completely separate and unrelated note, our SDK's `deploy` method has a lot of redundant functions (like [this](https://github.com/caraml-dev/merlin/blob/a92c533930d69169d7184ad1f462eb01db1e8976/python/sdk/merlin/model.py#L1797)) that can be removed or simplified (I probably did a bad job cleaning them up the last time I touched this 🤦🏻‍♂️

deadlycoconuts:
Awesome! Thanks for checking it again and for sharing the screenshot! The release notes sound good too! Feel free to merge anytime you're ready!

PR #598, opened 2024-07-26
feat: Add more metrics and error log for observability publisher

tiopramayudi:
# Description

Add kafka lag metrics and error log for easier troubleshoot when issue arise regarding observability publisher
# Modifications

* Add scheduled daemon to calculate lag for kafka consumer 
* Add more error log during failure
* Add container port for observability publisher deployment so it can be scraped by the prometheus agent
# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/pkg/observability/deployment/deployment.go
+++ b/api/pkg/observability/deployment/deployment.go
@@ -316,7 +316,7 @@ func (c *deployer) applyDeployment(ctx context.Context, data *models.WorkerData,
 	deploymentV1 := appV1.Deployments(c.targetNamespace())
 
 	applyDeploymentFunc := func(data *models.WorkerData, secretName string, isExistingDeployment bool) (*appsv1.Deployment, error) {
-		deployment, err := c.createDeploymentSpec(ctx, data, secretName)
+		deployment, err := c.createDeploymentSpec(data, secretName)
 		if err != nil {
 			return nil, err
 		}
@@ -342,7 +342,7 @@ func (c *deployer) getLabels(data *models.WorkerData) map[string]string {
 	return labels
 }
 
-func (c *deployer) createDeploymentSpec(ctx context.Context, data *models.WorkerData, secretName string) (*appsv1.Deployment, error) {
+func (c *deployer) createDeploymentSpec(data *models.WorkerData, secretName string) (*appsv1.Deployment, error) {
 	labels := c.getLabels(data)
 
 	cfgVolName := "config-volume"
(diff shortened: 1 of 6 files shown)


python/observation-publisher/publisher/prediction_log_consumer.py
```diff
@@ -141,9 +142,54 @@ def __init__(
...
         self._batch_size = config.batch_size
+        self._topic = config.topic
         self._consumer.subscribe([config.topic])
         self._poll_timeout = config.poll_timeout_seconds
 
+        background_job_thread = Thread(target=self._emit_metrics)
+        background_job_thread.setDaemon(True)
+        background_job_thread.start()
+
+    def _emit_metrics(self):
```
bthari, line 153: I want to confirm something, is the objective of this function is to calculate the kafka lag every 60 seconds?

tiopramayudi: yes

deadlycoconuts: > please reconfirm what is the acceptable interval

Are we going to remove this comment in https://github.com/caraml-dev/merlin/pull/598/files#diff-95854776df8c83183573e19f3bf6e2bfe61f73ccc4077d92ee9b64385c13d403R159 then? 😅

api/pkg/observability/deployment/deployment.go
```diff
@@ -394,14 +394,21 @@ func (c *deployer) createDeploymentSpec(ctx context.Context, data *models.Worker
...
+							},
 						},
 					},
 					Volumes: []corev1.Volume{
 						{
 							Name: cfgVolName,
 							VolumeSource: corev1.VolumeSource{
 								Secret: &corev1.SecretVolumeSource{
-									SecretName: c.getSecretName(data),
+									SecretName: secretName,
```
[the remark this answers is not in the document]
tiopramayudi: no, it just to use everything that is passed to the function, the caller use `getSecretName` as `secretName` parameter in this function so the value would be the same.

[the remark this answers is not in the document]
deadlycoconuts: Ah got it, just noticed it was previously unused. Thanks!

python/observation-publisher/publisher/prediction_log_consumer.py
bthari: when will this function be called repeatedly to log the lag, afaik the `start()` on `Thread` only call the target function once?

tiopramayudi: nice catch! it should be looping forever in target function

deadlycoconuts (approved):
Thanks a lot for the changes @tiopramayudi and @bthari for reviewing this! Don't think I have any other comments on the PR 🚀

bthari:
Thank you for the changes! LGTM 🚀

PR #599, opened 2024-07-29
fix: pdb validity check & deletion

bthari:
# Description
## Background
When creating Pod Disruption Budget, Kubernetes gives two ways for user to specify the allowed disruption amount, use `minAvailable` or `maxUnavailable` . The example of behavior is as below: 
```
current replica = 2

Use MinAvailable:
min available = 80%
allowed disruption = current replica - min available replica = 2 - ceil(80% * 2) = 2 - 2 = 0 

Use MaxUnavailable:
max unavailable = 20%
allowed disruption = ceil(20% * 2) = ceil(0.4) = 1
```

### 1. Miscalculation on PDB validity check

Merlin do a validity check on the PDB configuration, if the configuration doesn’t allow for any disruption, then Merlin will not create the PDB, this to avoid Kubernetes can’t remove any replica because all replica must be up. When the configuration in Merlin use the `maxUnavailability`, the calculation will be ([ref1](https://github.com/caraml-dev/merlin/blob/8c4930dc2c3f3edf3f1a862a99dd3cb5730c50cb/api/cluster/pdb.go#L88) & [ref2](https://github.com/caraml-dev/merlin/blob/8c4930dc2c3f3edf3f1a862a99dd3cb5730c50cb/api/cluster/pdb.go#L94-L96)):
```
minPercentage = (100 - maxUnavailability)%

if not (minPercentage * minReplica < minReplica)
	don't create PDB 
```
Which doesn’t reflect the real calculation on Kubernetes. Ex: 
```
replica: 5
maxUnavailability: 10%

Kubernetes calculation:
allowed disruption = ceil(10% * 5) = ceil(0.5) = 1

Merlin calculation:
minPercentage = (100 - 10)% = 90%
minAvailableReplica = ceil(90% * 5) = ceil(4.5) = 5 
allowed disruption = 5 - 5 = 0 
```
And therefore in the example above, Merlin will not create any PDB configuration because it thinks that if the PDB is created, it might not allowed any pod to be disrupted. Merlin avoid this because in the event of node scale down, the pods can’t be removed and causing deadlock. But actually, the Kubernetes calculation does allow for 1 pod to be disrupted. 

### 2. Unused PDB not deleted
When a model is redeployed, it will create a new revision ID, and there's a chance that unused PDB for the previous  revision is not deleted. Scenario: 
1. Current state of model: has predictor & transformer PDB 
2. User redeploy a new model version and this version doesn't have PDB
3. Newer model version is deployed, old unused PDB is not deleted (because Merlin doesn't check or delete if previously PDB exist) 

**Notes:** the old PDB will not affect the new model deployment, because in the `selector.matchLabels` there's a label of `inferenceservice:{modelName}-{versionID}-{revisionID}`, and the revisionID will be incremented on every new deployment. So it is safe to not delete PDB, but it will going to confuse user. 

# Modifications
Changes: 

- Fix the validity checking of the PDB. The `minAvailable` will use the current behavior, but if Merlin config uses `maxUnavailable` it will create PDB as long as the number is bigger than 0. 
- Merlin will create PDB with a fix name of `{name}-{versionID}-{componentType}-pdb`, where `componentType` is either `predictor` or `transformer`. The changes in the code will compare those two fix names with the new PDBs name, then for name that doesn't exist in new PDBs name, delete it. 

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
fix wrong validation check of PDB and delete unused PDB from previous version model
```

### DIFF
--- a/api/cluster/controller.go
+++ b/api/cluster/controller.go
@@ -260,7 +260,7 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 
 		unusedPdbs := getUnusedPodDisruptionBudgets(modelService, pdbs)
 		if err := c.deletePodDisruptionBudgets(ctx, unusedPdbs); err != nil {
-			log.Warnf("unable to delete unused pdb: %v", err)
+			log.Warnf("unable to delete model name %s, version %s unused pdb: %v", modelService.ModelName, modelService.ModelVersion, err)
 		}
 	}
 
--- a/api/cluster/controller.go
+++ b/api/cluster/controller.go
@@ -257,6 +257,11 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 			log.Errorf("unable to create pdb: %v", err)
 			return nil, errors.Wrapf(err, fmt.Sprintf("%v", ErrUnableToCreatePDB))
 		}
+
+		stalePdbs := getUnusedPodDisruptionBudgets(modelService, pdbs)
+		if err := c.deletePodDisruptionBudgets(ctx, stalePdbs); err != nil {
+			log.Warnf("unable to delete stale pdb: %v", err)
+		}
 	}
 
 	s, err = c.waitInferenceServiceReady(s)
--- a/api/cluster/controller.go
+++ b/api/cluster/controller.go
@@ -257,11 +257,6 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 			log.Errorf("unable to create pdb: %v", err)
 			return nil, errors.Wrapf(err, fmt.Sprintf("%v", ErrUnableToCreatePDB))
 		}
-
-		unusedPdbs := getUnusedPodDisruptionBudgets(modelService, pdbs)
-		if err := c.deletePodDisruptionBudgets(ctx, unusedPdbs); err != nil {
-			log.Warnf("unable to delete model name %s, version %s unused pdb: %v", modelService.ModelName, modelService.ModelVersion, err)
-		}
 	}
 
 	s, err = c.waitInferenceServiceReady(s)
@@ -293,11 +288,20 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 		inferenceURL = vsCfg.getInferenceURL(vs)
 	}
 
-	// Delete previous inference service
+	// Delete previous inference service and pdb
 	if modelService.CurrentIsvcName != "" {
 		if err := c.deleteInferenceService(ctx, modelService.CurrentIsvcName, modelService.Namespace); err != nil {
 			log.Errorf("unable to delete prevision revision %s with error %v", modelService.CurrentIsvcName, err)
 			return nil, errors.Wrapf(err, fmt.Sprintf("%v (%s)", ErrUnableToDeletePreviousInferenceService, modelService.CurrentIsvcName))
+		}
+
+		unusedPdbs, err := c.getUnusedPodDisruptionBudgets(ctx, modelService)
+		if err != nil {
+			log.Warnf("unable to get model name %s, version %s unused pdb: %v", modelService.ModelName, modelService.ModelVersion, err)
+		} else {
+			if err := c.deletePodDisruptionBudgets(ctx, unusedPdbs); err != nil {
+				log.Warnf("unable to delete model name %s, version %s unused pdb: %v", modelService.ModelName, modelService.ModelVersion, err)
+			}
 		}
 	}
 
--- a/api/cluster/controller.go
+++ b/api/cluster/controller.go
@@ -258,9 +258,9 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 			return nil, errors.Wrapf(err, fmt.Sprintf("%v", ErrUnableToCreatePDB))
 		}
 
-		stalePdbs := getUnusedPodDisruptionBudgets(modelService, pdbs)
-		if err := c.deletePodDisruptionBudgets(ctx, stalePdbs); err != nil {
-			log.Warnf("unable to delete stale pdb: %v", err)
+		unusedPdbs := getUnusedPodDisruptionBudgets(modelService, pdbs)
+		if err := c.deletePodDisruptionBudgets(ctx, unusedPdbs); err != nil {
+			log.Warnf("unable to delete unused pdb: %v", err)
 		}
 	}
 
--- a/api/cluster/pdb.go
+++ b/api/cluster/pdb.go
@@ -37,7 +37,7 @@ func NewPodDisruptionBudget(modelService *models.Service, componentType string, 
 	labels["serving.kserve.io/inferenceservice"] = modelService.Name
 
 	return &PodDisruptionBudget{
-		Name:                     fmt.Sprintf("%s-%s-%s-%s", modelService.ModelName, modelService.ModelVersion, componentType, models.PDBComponentType),
+		Name:                     buildPDBName(modelService, componentType),
 		Namespace:                modelService.Namespace,
 		Labels:                   labels,
 		MaxUnavailablePercentage: pdbConfig.MaxUnavailablePercentage,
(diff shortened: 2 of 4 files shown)


api/cluster/controller.go
```diff
@@ -257,6 +257,11 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 			log.Errorf("unable to create pdb: %v", err)
 			return nil, errors.Wrapf(err, fmt.Sprintf("%v", ErrUnableToCreatePDB))
 		}
+
+		unusedPdbs := getUnusedPodDisruptionBudgets(modelService, pdbs)
+		if err := c.deletePodDisruptionBudgets(ctx, unusedPdbs); err != nil {
+			log.Warnf("unable to delete unused pdb: %v", err)
```
bthari, line 263: If the name of the PDB that want to be deleted is not exist, this will not return an error. But if for any other reason this call throw an error, I only give a log warning because the previous version PDB will not affect the newly deployed model version anyway (see **notes** in PR description for detail)

tiopramayudi: can you add pdb name in the warning log?

api/cluster/pdb.go
```diff
@@ -76,40 +83,70 @@ func (cfg PodDisruptionBudget) BuildPDBSpec() (*policyv1.PodDisruptionBudget, er
...
 	} else {
 		// PDB config is not set properly, we can't apply it.
 		return pdbs
 	}
 
-	if modelService.ResourceRequest != nil &&
-		math.Ceil(float64(modelService.ResourceRequest.MinReplica)*
-			minAvailablePercent) < float64(modelService.ResourceRequest.MinReplica) {
+	if (modelService.ResourceRequest != nil) &&
+		(maxUnavailablePercent > 0 || (pdbConfig.MinAvailablePercentage != nil &&
```
tiopramayudi, line 103: can you extract the logic to determine the eligibility of pdb creation in a separate function?

api/cluster/pdb.go
```diff
@@ -76,40 +83,70 @@ func (cfg PodDisruptionBudget) BuildPDBSpec() (*policyv1.PodDisruptionBudget, er
...
+func getUnusedPodDisruptionBudgets(modelService *models.Service, pdbs []*PodDisruptionBudget) []*PodDisruptionBudget {
+	unusedPdbs := []*PodDisruptionBudget{}
+
+	if len(pdbs) == len(pdbComponents) {
+		return unusedPdbs
+	}
+
+	defaultPDBSNames := []string{}
+	for _, componentType := range pdbComponents {
+		defaultPDBSNames = append(defaultPDBSNames, fmt.Sprintf("%s-%s-%s-%s", modelService.ModelName, modelService.ModelVersion, componentType, models.PDBComponentType))
```
tiopramayudi, line 130: pdb name logic also declared in the `NewPodDisruptionBudget`, can we use the same function to generate pdb name to reduce mistake when updating the name convention?

api/cluster/pdb.go
```diff
@@ -76,40 +83,70 @@ func (cfg PodDisruptionBudget) BuildPDBSpec() (*policyv1.PodDisruptionBudget, er
...
+	}
+
+	newPDBSNames := make(map[string]bool)
+	for _, pdb := range pdbs {
+		newPDBSNames[pdb.Name] = true
+	}
+
+	for _, name := range defaultPDBSNames {
+		if !newPDBSNames[name] {
```
tiopramayudi, line 139: nit: can we `continue` if the pdf names is exist instead

deadlycoconuts:
> And therefore in the example above, Merlin will not create any PDB configuration because it thinks that no disruption is allowed.

Just a tiny comment on this phrasing (nothing wrong with your interpretation/implementation), it's not because no disruption is allowed that's why PDBs are not created by Merlin. In fact, not creating PDBs simply allow all sorts of disruptions to happen 😅 

The reason why Merlin currently does not create PDBs when minAvailableReplica >= minReplica, is because that would **_normally_** imply that the deployment **cannot** afford any pod being disrupted. At this point, this aligns with our expectation of how increasing the `maxUnavailability` value should work - the smaller the max number of pods that are allowed to be unavailable, the 'safer' the deployment should be because we want more pods to be protected.

However, when minAvailableReplica >= minReplica, this causes a separate problem when a node with those pods needs to scale down, because now those pods cannot be removed, causing a deadlock. As such, Merlin's 'fix' is to not create PDBs at all for such cases. And yes, this means the pods of this deployment becomes entirely unprotected - in other words, the most unsafe possible.

Hence the current implementation ironically makes it such that the more a user lowers the `maxUnavailability` value, the safer it becomes, until a certain point where it just falls off a cliff and becomes completely unsafe (when minAvailableReplica >= minReplica and no PDBs are created).

api/cluster/controller.go
bthari: I’m a bit unsure what’s the best way to do this. The `deletePodDisruptionBudgets` is used to delete multiple PDBs, will it be odd to return only one name `(name string, err error)` for logging purpose, or is that fine? 

Should I iterate and delete the pdb like this? But I think it will be redundant with `deletePodDisruptionBudgets` function
```go
unusedPdbs := getUnusedPodDisruptionBudgets(modelService, pdbs)
for _, pdb := range unusedPdbs {
 if err := c.deletePodDisruptionBudget(ctx, pdb); err != nil {
  log.Warnf("unable to delete unused pdb %s: %v", pdb.Name, err)
 }
}
```
 Do you have any advice?

tiopramayudi: I think merlin model and version should sufficient

deadlycoconuts: Actually I've got a completely different take on this - before calling `deployPodDisruptionBudgets`, can we simply delete all existing PDBs using `deletePodDisruptionBudgets`? That way we don't need to deal with additional logic in `getUnusedPodDisruptionBudgets`? I'm referring to something like this:
```go
		existingPDBs := getExistingPodDisruptionBudgets(modelService)
		if err := c.deletePodDisruptionBudgets(ctx, existingPDBs); err != nil {
			// do something
		}
		// Create new pdbs
		pdbs := generatePDBSpecs(modelService, c.deploymentConfig.PodDisruptionBudget)
		if err := c.deployPodDisruptionBudgets(ctx, pdbs); err != nil {
			log.Errorf("unable to create pdb: %v", err)
			return nil, errors.Wrapf(err, fmt.Sprintf("%v", ErrUnableToCreatePDB))
		}
```

I see two ways to get the existing PDBs, one is to use the Kube API to retrieve a list of actual PDBs for that deployment (potentially using a label selector like this: https://github.com/caraml-dev/merlin/blob/24d79eb81a3a9e373f3c12bcf134a982231c03c0/api/cluster/pdb.go#L30), OR to simply generate a list of potential PDB names (only 2 combinations really, model only or model+transformer) since we already have the naming convention for PDBs, and then attempt to delete all of them. In the second case, we can choose to not return any error (and just log them if necessary).

deadlycoconuts: Oh and I don't want to ruin the party, but we might have to consider porting these changes to Turing too 👀

bthari: Okay, got it!

bthari:
> Just a tiny comment on this phrasing (nothing wrong with your interpretation/implementation), it's not because no disruption is allowed that's why PDBs are not created by Merlin. In fact, not creating PDBs simply allow all sorts of disruptions to happen 😅

Thank you so much @deadlycoconuts for explaining in detail on the current behavior of Merlin and why Merlin doesn’t create a PDB if it may cause a deadlock! I’ve rephrased the description as to not create confusion, hopefully it’s a clearer one. 

> (maybe we can always specify that 1 pod at minimum, is always available for disruption whenever this edge case happens, so that such deployments get the protection from PDBs without blocking node scaling down events)

Yes, I think we can have a discussion about this case of PDB fallback configuration, rather than having no PDB at all 😁

api/cluster/controller.go
bthari: Oh, I didn’t see your @deadlycoconuts  comment when sending my latest comment lol.

When we run the delete command first, it could yield potential extra operations. E.g. previously there’s both the predictor & transformer PDB, we will then delete that two PDBs, and recreate two PDBs. While if we create it first, we will only need to patch the PDBs (because it has the same name), and delete nothing. There’s also a lookup operation if we want to use the label selector. My question is, considering there’s a lot of things happening when deploying a model already, is the potential operations is fine? If yes, I think we can move with your suggestion bcs the code looks more intuitive. 
> In either case, we would need need to update the implementation of this method here (called when undeploying Merlin models) too, so that we aren't deleting only PDBs that would've been created by generatePDBSpecs:

Does this mean if there’s a PDB not created by `generatePDBSpecs`, we will delete it to? 

And, what’s going on with Turing? 😅

deadlycoconuts: Haha no worries lol, I'm not constantly following the comments in this PR either 😂

Yeah what you said it's true, we'll definitely be making more calls to the Kube API server and not take advantage of the 'patch' operation in k8s by first deleting the existing PDBs before recreating them (if any). To be honest I think any additional operations should be fine, if you were to perform a list operation with the label selector to get the existing PDBs to delete them, that'll only be 3 more API calls at most, which is fine (adding/removing PDBs is also an almost instantaneous operation unlike adding/removing pods). 

> Does this mean if there’s a PDB not created by generatePDBSpecs, we will delete it to?

Hmm I thought over about this again, okay it's fine to keep it as it is; with your fix, all PDBs should be correct (there shouldn't be undeleted PDBs) with respect to the current deployment's resource requests, so using `generatePDBSpecs` would generate the correct set of PDBs to delete. (I was thinking of the possibility of just deleting all PDBs found for that particular model version but I guess that's unnecessary once we fix the redeployment workflow).

Oh and regarding the Turing comment, my bad, I checked how PDBs were created/deleted again and Turing actually doesn't have this bug because of the way how Turing tracks every single modification to a router as a version change (so its PDBs are always removed and recreated with a different name), unlike in Merlin where you can redeploy the same model version with a different set of deployment configs.

bthari: After a bit of discussion with @tiopramayudi, I think it will be safer to delete the PDB once the [new inference service is ready](https://github.com/caraml-dev/merlin/blob/24d79eb81a3a9e373f3c12bcf134a982231c03c0/api/cluster/controller.go#L262). 

Suppose that for any reason the new inference service is not ready (maybe bcs of timeout), Merlin will not going to serve the new model revision, it will go back to serve the current model revision instead. If we delete the PDB first, then the PDB that is used for the current revision is now already gone.

But, there is another issue with this case. If the new PDB is deployed but `waitInferenceServiceReady` got a timeout, the PDB now will not match the current pod's label anyway (because PDB has the same name, and the label selector has been replaced to the fail inference service's revisionID ([ref1](https://github.com/caraml-dev/merlin/blob/24d79eb81a3a9e373f3c12bcf134a982231c03c0/api/cluster/pdb.go#L30) & [ref2](https://github.com/caraml-dev/merlin/blob/24d79eb81a3a9e373f3c12bcf134a982231c03c0/api/cluster/pdb.go#L58))). 

Here’s what I’m thinking of. First, the name of the PDB will include the revision number (again):
```
{modelName}-{modelVersion}-{revisionID}-{componentType}-pdb
```

So, this new PDB will not patch the config of the current PDB. Then, once the inference service is up and everything is okay, the unused PDB will be deleted, can put it at the very end of the function, maybe right before [this](https://github.com/caraml-dev/merlin/blob/24d79eb81a3a9e373f3c12bcf134a982231c03c0/api/cluster/controller.go#L299). 

```go
unusedPDBs := getUnusedPodDisruptionBudgets(modelService)
if err := c.deletePodDisruptionBudgets(ctx, existingPDBs); err != nil {
	// log if error
}

```
I’m still a bit unsure on how to get the unused PDBs efficiently though because we need to filter it out based on model name & version, but the PDB label doesn’t have a version field. Maybe can use label selector where `app = {model name} AND inferenceservice != {inference service name}`, then filter out the name which doesn’t match the version number with regex.

deadlycoconuts: Got it I agree that it's safer to delete the PDB only after the new inference is ready, and that we can delete the PDBs at the very end of the function. What about doing it [here](https://github.com/caraml-dev/merlin/blob/24d79eb81a3a9e373f3c12bcf134a982231c03c0/api/cluster/controller.go#L293), just after deleting the old inference service? 

I also like the idea of reintroducing the revision number back into the PDB names again so that makes it much easier for us to retrieve and delete the correct PDBs. 

> I’m still a bit unsure on how to get the unused PDBs efficiently though because we need to filter it out based on model name & version, but the PDB label doesn’t have a version field. Maybe can use label selector where app = {model name} AND inferenceservice != {inference service name}, then filter out the name which doesn’t match the version number with regex.

Hmm can we just attempt to perform 2 deletion operations? We can construct the names of the (existing) PDBs entirely using whatever's contained within the `modelService` object - it contains `CurrentIsvcName` (which is the existing isvc's name), which is equivalent to `{modelName}-{modelVersion}-{revisionID}` already. All we need to do is append `-{componentType}-pdb` and then attempt to delete both of them.

bthari: Hi I’ve updated the code for this case 😁

I said in the comment that the PDB’s `label` and `matchLabels.Selector` are the same, but as I’ve mentioned in our meeting, I think it will be simpler to modify that. So, Merlin will add `model-version-id` labels to PDB, but the `selector` will follow current behavior. With this approach, the `getUnusedPodDisruptionBudgets` will get all unused PDB based on model name and version, instead of just the previous latest version PDB (in case of deletion failure, it can be picked up another time user redeploy the endpoint).

The name of the PDB will still use a revisionID so `{modelName}-{modelVersion}-{revisionID}-{componentType}-pdb` and I put the deletion code after the line you mentioned.

And I’ll get back again after finding out how to remove all the other PDB from previous/current Merlin’s version that is still running in our cluster

api/cluster/pdb_test.go
```diff
@@ -17,13 +17,18 @@ func TestPodDisruptionBudget_BuildPDBSpec(t *testing.T) {
...
+	}
+	defaultSelectors := map[string]string{
 		"gojek.com/app": "sklearn-sample-s",
 	}
 
 	type fields struct {
 		Name                     string
 		Namespace                string
 		Labels                   map[string]string
+		Selectors                map[string]string
```
tiopramayudi, line 31: can you add new test for function `NewPodDisruptionBudget`

bthari: Sure, I've added the unit test for this

tiopramayudi (approved):
LGTM ! Thanks @bthari for discovering the edge case and fix it

PR #601, opened 2024-08-12
feat: add webhook to model version deployment & undeployment

bthari:
# Description
When deploying or undeploying model endpoint, other entity might want to get a trigger to automate their process or for Merlin other process. This PR add webhooks call (based on the [webhook from MLP](https://github.com/caraml-dev/mlp/blob/main/api/pkg/webhooks/README.md)), so on model version deployment/undeployment it will call the configured webhooks. 

# Modifications
**Main Changes:** 

Added webhooks call on
- Model version pre-deployment: will ignore the success response (for this version) and will stop/fail the deployment version if any `async` webhook fail
- Model version post-deployment: will ignore error, only log the error if any occur during the call
- Model version post-undeployment: will ignore error, only log the error if any occur during the call

Request payload to webhook: 
- Request:
  - `event_type`: name of event which triggers the webhook
  - `versionEndpoint`: object version endpoint

**Side effect changes**:
- With MLP update in go.mod, the assert function is also updated. Previously, the `assert.InEpsilon` can pass when item in actual slice is in expected slice, even though the expected slice might have more items; now the `testify/assert` will check the two slices length first ([ref](https://github.com/stretchr/testify/pull/1483/files)) -> Added some changes to fix the unit test in `TestToFloat64List`

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
add configurable webhook call in endpoint deployment and undeployment
```

### DIFF
--- a/api/queue/work/model_service_deployment_test.go
+++ b/api/queue/work/model_service_deployment_test.go
@@ -191,7 +191,7 @@ func TestExecuteDeployment(t *testing.T) {
 			}(),
 		},
 		{
-			name:    "Success: with calling webhook",
+			name:    "Success: with calling webhooks",
 			model:   model,
 			version: version,
 			endpoint: &models.VersionEndpoint{
--- a/api/queue/work/model_service_deployment_test.go
+++ b/api/queue/work/model_service_deployment_test.go
@@ -14,7 +14,8 @@ import (
 	eventMock "github.com/caraml-dev/merlin/pkg/observability/event/mocks"
 	"github.com/caraml-dev/merlin/queue"
 	"github.com/caraml-dev/merlin/storage/mocks"
-	"github.com/caraml-dev/mlp/api/pkg/webhooks"
+	webhook "github.com/caraml-dev/merlin/webhooks"
+	webhookManager "github.com/caraml-dev/mlp/api/pkg/webhooks"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"gorm.io/gorm"
@@ -73,7 +74,7 @@ func TestExecuteDeployment(t *testing.T) {
 		storage           func() *mocks.VersionEndpointStorage
 		controller        func() *clusterMock.Controller
 		imageBuilder      func() *imageBuilderMock.ImageBuilder
-		webhookManager    func() webhooks.WebhookManager
+		webhookManager    func() webhookManager.WebhookManager
 		eventProducer     *eventMock.EventProducer
 	}{
 		{
@@ -119,8 +120,8 @@ func TestExecuteDeployment(t *testing.T) {
 				mockImgBuilder := &imageBuilderMock.ImageBuilder{}
 				return mockImgBuilder
 			},
-			webhookManager: func() webhooks.WebhookManager {
-				webhookManager := webhooks.NewMockWebhookManager(t)
+			webhookManager: func() webhookManager.WebhookManager {
+				webhookManager := webhookManager.NewMockWebhookManager(t)
 				webhookManager.On("IsEventConfigured", mock.Anything).Return(false)
 				return webhookManager
 			},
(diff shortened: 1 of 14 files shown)


api/queue/work/model_service_deployment_test.go
```diff
@@ -178,6 +190,56 @@ func TestExecuteDeployment(t *testing.T) {
...
+				return ctrl
+			},
+			imageBuilder: func() *imageBuilderMock.ImageBuilder {
+				mockImgBuilder := &imageBuilderMock.ImageBuilder{}
+				return mockImgBuilder
+			},
+			webhookManager: func() webhooks.WebhookManager {
+				webhookManager := webhooks.NewMockWebhookManager(t)
+				webhookManager.On("IsEventConfigured", mock.Anything).Return(true)
+				webhookManager.On("InvokeWebhooks", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil)
```
tiopramayudi, line 239: nit: instead of `mock.Anything` can you use exact parameter that will be passed?

api/service/version_endpoint_service.go
```diff
@@ -281,6 +301,19 @@ func (k *endpointService) UndeployEndpoint(ctx context.Context, environment *mod
...
+	// calling webhooks if there's any webhooks configured
+	if k.webhookManager != nil && k.webhookManager.IsEventConfigured(webhooks.OnModelVersionUndeployed) {
+		body := &webhooks.VersionEndpointRequest{
+			EventType:       webhooks.OnModelVersionUndeployed,
+			VersionEndpoint: endpoint,
+		}
+
+		err = k.webhookManager.InvokeWebhooks(ctx, webhooks.OnModelVersionUndeployed, body, webhookManager.NoOpCallback, webhookManager.NoOpErrorHandler)
+		if err != nil {
+			log.Warnf("unable to invoke webhooks for event type: %s, model: %s, version: %s, error: %v", webhooks.OnModelVersionDeployed, model.Name, version.ID, err)
```
tiopramayudi, line 313: should it be `webhooks.OnModelVersionUndeployed`

api/service/version_endpoint_service.go
```diff
@@ -132,6 +138,20 @@ func (k *endpointService) DeployEndpoint(ctx context.Context, environment *model
...
+	// calling webhooks if there's any webhooks configured
+	if k.webhookManager != nil && k.webhookManager.IsEventConfigured(webhooks.OnModelVersionPredeployment) {
+		body := &webhooks.VersionEndpointRequest{
+			EventType:       webhooks.OnModelVersionPredeployment,
+			VersionEndpoint: endpoint,
+		}
+
+		err := k.webhookManager.InvokeWebhooks(ctx, webhooks.OnModelVersionPredeployment, body, webhookManager.NoOpCallback, webhookManager.NoOpErrorHandler)
+		if err != nil {
+			log.Errorf("unable to invoke webhooks for event type: %s, model: %s, version: %s, error: %v", webhooks.OnModelVersionDeployed, model.Name, version.ID, err)
```
tiopramayudi, line 150: should it be `webhooks.OnModelVersionUndeployed`?

api/queue/work/model_service_deployment_test.go
bthari: Can do for the eventType (and have changed it). But for the third argument/payload (`*versionEndpointRequest`), I don’t think can do it, because the `*jobArgs` [is getting marshaled and unmarshaled](https://github.com/caraml-dev/merlin/blob/75729fd51b63689178882a81ff1a940b01131bab/api/queue/work/model_service_deployment.go#L57-L62) so I can't really compare the pointer address as it will differs from `tt.endpoint`

api/service/version_endpoint_service.go
bthari: Thank you far catching these two!

tiopramayudi (approved):
LGTM ! Thanks @bthari

PR #603, opened 2024-08-29
fix: string validation in variable input schema

bthari:
# Background
When trying to input a `jsonPath` value type in standard transformer page (for variable input & feast table input), the value that is passed to the yup validation is type of object (e.g. `{ "jsonPath": "$.id" }`), instead of just the string value of jsonPath (`$.id`) inputted in the form. This prevent user to submit the form when their actual input is valid.  

Screenshot 2024-08-29 at 13 18 40
Screenshot 2024-08-29 at 13 40 48

# Modifications
Use a custom yup test function to validate `value`; if the value passed is type of object & has the field of `jsonPath`, it will compare the `jsonPath` value to the original schema; else can just compare the original value to the yup schema.

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
fix variable input validation in transformer config ui page
```

### DIFF
--- a/ui/src/pages/version/components/forms/validation/schema.js
+++ b/ui/src/pages/version/components/forms/validation/schema.js
@@ -67,13 +67,13 @@ const environmentVariableSchema = yup.object().shape({
   value: yup.string(),
 });
 
-const stringValueSchema = (name) => yup.mixed().test(
+const stringValueSchema = (name, key) => yup.mixed().test(
   `value input schema`,
   `${name} must be a \`string\` but the final value was $\{value}`,
   (value, { createError }) => {
     try {
-      if (typeof value == "object" && "jsonPath" in value) {
-        return yup.string().required(`${name} is required`).validateSync(value["jsonPath"])
+      if (typeof value == "object" && key in value) {
+        return yup.string().required(`${name} is required`).validateSync(value[key])
       }
       return yup.string().required(`${name} is required`).validateSync(value)
     } catch (e) {
@@ -107,7 +107,7 @@ const feastEntitiesSchema = yup.object().shape({
   name: yup.string().required("Entity Name is required"),
   valueType: yup.string().required("Entity Value Type is required"),
   fieldType: yup.string().required("Input Type is required"),
-  field: stringValueSchema("Input Value"),
+  field: stringValueSchema("Input Value", "jsonPath"),
 });
 
 const feastFeaturesSchema = yup.object().shape({
@@ -128,7 +128,7 @@ export const feastInputSchema = yup.object().shape({
 const variableInputSchema = yup.object().shape({
   name: yup.string().required("Name is required"),
   type: yup.string().required("Type is required"),
-  value: stringValueSchema("Value"),
+  value: stringValueSchema("Value", "jsonPath"),
 });
 
 const tablesInputSchema = yup.object().shape({
--- a/ui/src/pages/version/components/forms/validation/schema.js
+++ b/ui/src/pages/version/components/forms/validation/schema.js
@@ -70,11 +70,15 @@ const environmentVariableSchema = yup.object().shape({
 const stringValueSchema = (name) => yup.mixed().test(
   `value input schema`,
   `${name} must be a \`string\` but the final value was $\{value}`,
-  (value) => {
-    if (typeof value == "object" && "jsonPath" in value) {
-      return yup.string().required("Value is required").isValid(value["jsonPath"])
+  (value, { createError }) => {
+    try {
+      if (typeof value == "object" && "jsonPath" in value) {
+        return yup.string().required(`${name} is required`).validateSync(value["jsonPath"])
+      }
+      return yup.string().required(`${name} is required`).validateSync(value)
+    } catch (e) {
+      return createError({ message: e.message })
     }
-    return yup.string().required("Value is required").isValid(value)
   }
 )
 
@@ -103,7 +107,7 @@ const feastEntitiesSchema = yup.object().shape({
   name: yup.string().required("Entity Name is required"),
   valueType: yup.string().required("Entity Value Type is required"),
   fieldType: yup.string().required("Input Type is required"),
-  field: stringValueSchema("input field"),
+  field: stringValueSchema("Input Value"),
 });
 
 const feastFeaturesSchema = yup.object().shape({
--- a/ui/src/pages/version/components/forms/validation/schema.js
+++ b/ui/src/pages/version/components/forms/validation/schema.js
@@ -67,6 +67,17 @@ const environmentVariableSchema = yup.object().shape({
   value: yup.string(),
 });
 
+const variableValueInputSchema = (name) => yup.mixed().test(
+  `value input schema`,
+  `${name} must be a \`string\` but the final value was $\{value}`,
+  (value) => {
+    if (typeof value == "object" && "jsonPath" in value) {
+      return yup.string().required("Value is required").isValid(value["jsonPath"])
+    }
+    return yup.string().required("Value is required").isValid(value)
+  }
+)
+
 export const versionEndpointSchema = (maxAllowedReplica) => yup.object().shape({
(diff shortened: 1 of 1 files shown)


ui/src/pages/version/components/forms/validation/schema.js
```diff
@@ -67,6 +67,17 @@ const environmentVariableSchema = yup.object().shape({
   value: yup.string(),
 });
 
+const stringValueSchema = (name) => yup.mixed().test(
```
bthari, line 70: `yup.mixed()` will take in any type of variable, therefore both the `object` and `string` will pass this validation check. the further validation will be put in the `.test` part, where it will compare the original schema to the input or the value of `jsonPath` (depending on input type)

ui/src/pages/version/components/forms/validation/schema.js
```diff
@@ -67,6 +67,21 @@ const environmentVariableSchema = yup.object().shape({
...
+  `value input schema`,
+  `${name} must be a \`string\` but the final value was $\{value}`,
+  (value, { createError }) => {
+    try {
+      if (typeof value == "object" && "jsonPath" in value) {
+        return yup.string().required(`${name} is required`).validateSync(value["jsonPath"])
+      }
+      return yup.string().required(`${name} is required`).validateSync(value)
+    } catch (e) {
+      return createError({ message: e.message })
```
deadlycoconuts, line 80: Where is this `createError` function defined? :O

ui/src/pages/version/components/forms/validation/schema.js
deadlycoconuts: Since `jsonPath` is hardcoded in this function, could we either:
1) Rename the name of this function to reflect that it only checks for the value of the string that corresponds to the `jsonPath` key, e.g. maybe something like `jsonPathValueSchema`, or 
2) Make this function take in the actual key that it needs to access the string value i.e. make it the key value an argument in this function alongside `name`

deadlycoconuts:
You might've already brought this up previously (I have goldfish memory sorry) but I was wondering why the javascript struct that yup validates is slightly different from what appears in the yaml config because they look like this:

Yup schema:
```javascript
const variableInputSchema = yup.object().shape({
  name: yup.string().required("Name is required"),
  type: yup.string().required("Type is required"),  // expected to be `jsonPath`
  value: yup.string().required("Value is required"),  // was originally expected to be a string, but now it's getting `{"jsonPath": "$.id"}` instead, which isn't recognised as a string
});

```

Yaml:
```yaml
name: id
  fromJson:
    jsonPath: $.id
```

Notice how the `type` is specified as a separate field in Javascript, and `value` contains an object string. Apparently this appears to be intentional (see [this](https://github.com/search?q=repo%3Acaraml-dev%2Fmerlin%20newItem%5B%22value%22%5D&type=code) for more details):
```javascript
if (newItem.value && newItem.value.jsonPath === undefined) {
  newItem["value"] = { jsonPath: newItem.value };
}
```

So it seems like `value` showing up as the object `{"jsonPath": "$.id"}` might be normal after all. I just have no idea why this wasn't a problem previously. Maybe the previous version of yup accepted that as a string or something :/

deadlycoconuts (approved):

ui/src/pages/version/components/forms/validation/schema.js
bthari: Let me change it to point 2 and test for a bit

ui/src/pages/version/components/forms/validation/schema.js
bthari: It's from [yup.test() second argument ](https://www.npmjs.com/package/yup#schematestname-string-message-string--function--any-test-function-schema), and found out how to use this from [here](https://github.com/jquense/yup/issues/222#issuecomment-557457026) (but not using it with `this.` because I'm using arrow function instead of regular function) 

Backstory: docs said that the return value should be true or false or `ValidationError`. I kinda expect when I'm throwing a `ValidationError`, the error shown up in the UI would be this custom `ValidationError` I defined, but it shows nothing somehow 😅 turns out I need to use this `createError` for it to work

bthari:
@deadlycoconuts thank you for always takes time in breaking down things!

> So it seems like value showing up as the object {"jsonPath": "$.id"} might be normal after all. I just have no idea why this wasn't a problem previously. Maybe the previous version of yup accepted that as a string or something :/ In any case, I don't think it's going to be easy to dig the original behaviour unless you were to downgrade yup and try performing the same checks again, so I think we can just stick to the fix you've implemented and keep in mind that there may be something deeper going on 😅 

Yes, I also think the casting from `$.id` (string) to `{"jsonPath": "$.id"}` (object) is intentional from Merlin side looking at the final yaml configuration and the proto config on the transformer backend side in [here](https://github.com/caraml-dev/merlin/blob/da2803ab384a83d352663100f299f2c0699836dd/protos/merlin/transformer/spec/variable.proto#L15) and [here](https://github.com/caraml-dev/merlin/blob/da2803ab384a83d352663100f299f2c0699836dd/protos/merlin/transformer/spec/common.proto#L16-L24), so I only fix the validation part.  I don't understand how previously the yup validation pass tho (but we upgrade it from `0.29` to `1.4`, there must be so much changes but I can't find the exact yup changes which change this behaviour).


I was also trying to change [this lines](https://github.com/caraml-dev/merlin/blob/da2803ab384a83d352663100f299f2c0699836dd/ui/src/pages/version/components/forms/components/transformer/components/table_inputs/VariablesInput.js#L53-L58) 
```js
      case "jsonpath":
        if (newItem.value && newItem.value.jsonPath === undefined) {
          newItem["value"] = { jsonPath: newItem.value };
        }
        newItem["jsonPathConfig"] = newItem.value;
```
into 
```js
      case "jsonpath":
        if (newItem.value && newItem.value.jsonPath === undefined) {
          newItem["value"] = newItem.value ;
        }
        newItem["jsonPathConfig"] = { jsonPath: newItem.value };
```
so the `newItem["value"]` (that will be validated by yup) doesn't change into an object, and only `newItem["jsonPathConfig"]` that will have those `jsonPath` object, but the yaml config got broken 🥲 so I just stick into the this validation fix

ui/src/pages/version/components/forms/validation/schema.js
deadlycoconuts: Ahh got it, TIL hahaha 😅 Thanks for sharing that. Just one more short question though, how is the error message that shows up different if you were to `return false` vs `return createError({ message: e.message })`? 🤔

bthari: So, when this func return `false` (tho none here will return false) it's going to show the default `"${name} must be a string but the final value was $\{value}"` error I put as the message argument. If I throw e.g. `createError({ message: "new error"})`, it will show `"new error"` as the message instead

Now, I've mentioned that the function will actually never return a `false`, it's because `.validateSync()` will either return the value or `ValidationError`. When `.validateSync()` throw an error, I'll simply throw the error again (correctly using the `createError`), so the value input check will has the same error message for every field in [here](https://github.com/caraml-dev/merlin/blob/da2803ab384a83d352663100f299f2c0699836dd/ui/src/pages/version/components/forms/validation/schema.js#L113-L117)

example of before using `createError` (I return false if there's an error) vs using `createError` 

Screenshot 2024-08-29 at 14 42 29

Screenshot 2024-09-02 at 14 23 33

deadlycoconuts: Ahh got it, thanks for the really concise and clear explanation! I was about to ask if using a simpler error message would've made it simpler 😅 like use `String input value is required` instead of `${name} must be a string but the final value was $\{value}` since it seems like the `value` variable can't be parsed correctly (weird behaviour I don't know why this happens with `yup`) when a `ValidationError` is thrown. But your implementation works too, thanks a lot for investigating this!

PR #604, opened 2024-09-06
chore: add managed by label to namespace

bthari:
# Description

When creating a namespace object in Kubernetes, add label “{prefix}/managed: true”. This will help when we want to modify resources created by CaraML services in a shared cluster, e.g. configuring log for the managed namespace only

By default the namespace prefix would `caraml.dev/`, and it can be configured in the config yaml

# Modifications
- As the creation of label for Kubernetes’ object in Merlin previously is bind to metadata struct, while namespace doesn’t have metada, in this changes the functionality of labelling (labeller) is moved to it’s own package
- Add one more prefix in `InitKubernetesLabeller` initiation
- Add the `{prefix}/managed: true` when creating namespace

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/cluster/labeller/labeller.go
+++ b/api/cluster/labeller/labeller.go
@@ -1,0 +1,84 @@
+package labeller
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const (
+	LabelAppName          = "app"
+	LabelComponent        = "component"
+	LabelEnvironment      = "environment"
+	LabelOrchestratorName = "orchestrator"
+	LabelStreamName       = "stream"
+	LabelTeamName         = "team"
+	LabelManaged          = "managed"
+)
+
+var reservedKeys = map[string]bool{
+	LabelAppName:          true,
+	LabelComponent:        true,
+	LabelEnvironment:      true,
+	LabelOrchestratorName: true,
+	LabelStreamName:       true,
+	LabelTeamName:         true,
+}
+
+var (
+	prefix           string
+	nsPrefix         string
+	environment      string
+	validPrefixRegex = regexp.MustCompile("^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?(/)$")
+)
+
+// InitKubernetesLabeller builds a new KubernetesLabeller Singleton
+func InitKubernetesLabeller(p, ns, e string) error {
+	if err := isValidPrefix(p, "prefix"); err != nil {
+		return err
+	}
+	if err := isValidPrefix(ns, "namespace prefix"); err != nil {
+		return err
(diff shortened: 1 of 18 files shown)


tiopramayudi (approved):
LGTM! Thanks @bthari

api/models/metadata.go
```diff
@@ -95,17 +57,17 @@ func (metadata *Metadata) Scan(value interface{}) error {
...
+		labeller.GetLabelName(labeller.LabelEnvironment):      labeller.GetEnvironment(),
+		labeller.GetLabelName(labeller.LabelOrchestratorName): orchestratorValue,
+		labeller.GetLabelName(labeller.LabelStreamName):       metadata.Stream,
+		labeller.GetLabelName(labeller.LabelTeamName):         metadata.Team,
 	}
 
 	for _, label := range metadata.Labels {
 		// skip label that is trying to override reserved key
-		if _, usingReservedKeys := reservedKeys[prefix+label.Key]; usingReservedKeys {
+		if labeller.IsReservedKey(labeller.GetPrefix() + label.Key) {
```
deadlycoconuts, line 70: Hang on if we add the prefix to the label key, e.g. `gojek.com/` + `team`, and look up its result in the `reservedKeys` map, we won't be able to find `gojek.com/team` in the map right? Since the map contains the label key values only i.e. `team`. Was this even working before? 😦

deadlycoconuts (approved):
LGTM! 🚀 Just left a couple of comments but everything looks good! Thanks for the refactoring!

api/models/metadata.go
bthari: You're right, this will only work if the prefix is `""` (but it can't be `""`). Was this working before? I honestly don't think so 😅 but Turing also work the same way, so I just keep the functionality as is

api/cluster/labeller/labeller.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+var reservedKeys = map[string]bool{
+	LabelAppName:          true,
+	LabelComponent:        true,
+	LabelEnvironment:      true,
+	LabelOrchestratorName: true,
+	LabelStreamName:       true,
+	LabelTeamName:         true,
+}
```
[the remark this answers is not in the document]
bthari: oh right, let me bump it in another PR

api/models/metadata.go
deadlycoconuts: Hmm alright I'll create a card for this in the backlog so that we can fix this in the future. Thanks for clarifying!

PR #606, opened 2024-09-12
chore: add new label to reserved key list

bthari:
# Description
Previously I added a new kubernetes label in #604, this should've been added to reserved key list

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/cluster/labeller/labeller.go
+++ b/api/cluster/labeller/labeller.go
@@ -22,6 +22,7 @@ var reservedKeys = map[string]bool{
 	LabelOrchestratorName: true,
 	LabelStreamName:       true,
 	LabelTeamName:         true,
+	LabelManaged:          true,
 }
 
 var (


deadlycoconuts (approved):

deadlycoconuts:
Actually my bad, I suddenly thought about this again, and was thinking that the reserved key list might not actually be used to filter out the `managed` label at all because that label is only applied on namespaces right? 

I don't think we allow users to make changes on the namespace labels so we might not actually need to add this to the reserved list after all. We might need to double check that the reserved list is only used to verify labels that users can change (and if that's the case we don't need to add this `managed` label anymore since users can't change namespace labels).

bthari:
Now that you mentioned it, @deadlycoconuts, what I need to be clarified is what's the meaning of "label that user can change"? Is it the one with the prefix, e.g. `prefix/app`?

My confusion comes from the inconsistency [here](https://github.com/caraml-dev/merlin/blob/f9140f5e5a00f82b236a9734f8f7119781e319fc/api/models/metadata.go#L70) (which as you've mentioned, will not work anyway 😅), where we validate is it reserved based on `prefix + key`:
```go
if labeller.IsReservedKey(labeller.GetPrefix() + label.Key) {
	continue
}
```
but on [here](https://github.com/caraml-dev/merlin/blob/f9140f5e5a00f82b236a9734f8f7119781e319fc/api/models/metadata.go#L84), when we actually put the name into the label, it wouldn't use the prefix
```go
labels[label.Key] = label.Value
```

So, I'm thinking that maybe instead of just the `app`, `stream`, etc label on the `reservedKeys`, it should've contained the complete name of our label (e.g. `prefix/steam`) so then we can also add `prefix/managed` to the list? 

But all in all, I think it's okay to close this merge and not add the `managed` label to the list, then we can revisit this `reservedKeys` ordeal later 😂

PR #607, opened 2024-09-19
feat: Using secret to attach identity to pod for observability publisher

tiopramayudi:
# Description
Since we are trying to move away from cloud-specific feature, hence as first step in observability publisher is to use secret for propagate identity instead of using existing workload identity way

# Modifications
Add `GOOGLE_APPLICATION_CREDENTIALS` for propagate google identity to pod
# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/config/observability.go
+++ b/api/config/observability.go
@@ -4,16 +4,16 @@ import "time"
 
 // ObservabilityPublisher
 type ObservabilityPublisher struct {
-	ArizeSink          ArizeSink
-	BigQuerySink       BigQuerySink
-	KafkaConsumer      KafkaConsumer
-	ImageName          string
-	DefaultResources   ResourceRequestsLimits
-	EnvironmentName    string
-	Replicas           int32
-	TargetNamespace    string
-	ServiceAccountName string
-	DeploymentTimeout  time.Duration `default:"30m"`
+	ArizeSink                ArizeSink
+	BigQuerySink             BigQuerySink
+	KafkaConsumer            KafkaConsumer
+	ImageName                string
+	DefaultResources         ResourceRequestsLimits
+	EnvironmentName          string
+	Replicas                 int32
+	TargetNamespace          string
+	DeploymentTimeout        time.Duration `default:"30m"`
+	ServiceAccountSecretName string
 }
 
 // KafkaConsumer
--- a/api/pkg/observability/deployment/deployment_test.go
+++ b/api/pkg/observability/deployment/deployment_test.go
@@ -49,10 +49,10 @@ const (
 	timeoutError deploymentStatus = "timeout_error"
 
 	namespace          = "caraml-observability"
-	serviceAccountName = "caraml-observability-sa"
+	serviceAccountName = "caraml-observability-sa-secret"
 )
 
-func createDeploymentSpec(data *models.WorkerData, resourceRequest corev1.ResourceList, resourceLimit corev1.ResourceList, imageName string) *appsv1.Deployment {
+func createDeploymentSpec(data *models.WorkerData, resourceRequest corev1.ResourceList, resourceLimit corev1.ResourceList, imageName string, serviceAccountSecretName string) *appsv1.Deployment {
 	labels := data.Metadata.ToLabel()
 	labels[appLabelKey] = data.Metadata.App
 	numReplicas := int32(2)
@@ -80,7 +80,6 @@ func createDeploymentSpec(data *models.WorkerData, resourceRequest corev1.Resour
 					},
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName: serviceAccountName,
 					Containers: []corev1.Container{
 						{
 							Name:  "worker",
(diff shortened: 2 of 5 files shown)


api/pkg/observability/deployment/identity.go
```diff
@@ -0,0 +1,52 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+func enrichIdentityToPod(podSpec corev1.PodSpec, secretName string, containerNames []string) corev1.PodSpec {
+	secretVolume := createVolumeFromSecret(secretName)
+	updatedPodSpec := podSpec.DeepCopy()
+
+	containerExist := false
+	for idx, containerSpec := range updatedPodSpec.Containers {
+		updatedContainerSpec := containerSpec
```
bthari, line 15: just curious, can't we use `containerSpec` directly? why it should be assigned to `updatedContainerSpec`?

api/pkg/observability/deployment/deployment_test.go
```diff
@@ -49,10 +49,10 @@ const (
 	timeoutError deploymentStatus = "timeout_error"
 
 	namespace          = "caraml-observability"
-	serviceAccountName = "caraml-observability-sa"
+	serviceAccountName = "caraml-observability-sa-secret"
```
deadlycoconuts, line 52: nit: Can we also rename the variable name to `serviceAccountSecretName`?

api/config/observability.go
```diff
@@ -4,16 +4,16 @@ import "time"
...
+	ArizeSink                ArizeSink
+	BigQuerySink             BigQuerySink
+	KafkaConsumer            KafkaConsumer
+	ImageName                string
+	DefaultResources         ResourceRequestsLimits
+	EnvironmentName          string
+	Replicas                 int32
+	TargetNamespace          string
+	DeploymentTimeout        time.Duration `default:"30m"`
+	ServiceAccountSecretName string
```
deadlycoconuts, line 16: I guess we're completely removing the ability for the platform maintainers of the Merlin API server to have the observability worker containers use workload identity? Just wanted to confirm if we've decided to completely replace credential access via workload identity with mounted GCP secrets (as opposed to giving the platform maintainers an option between the two).

api/pkg/observability/deployment/identity.go
```diff
@@ -0,0 +1,52 @@
+package deployment
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func enrichIdentityToPod(podSpec corev1.PodSpec, secretName string, containerNames []string) corev1.PodSpec {
```
deadlycoconuts, line 9: nit: Since this helper function only works by enriching pods with mounted Google application credentials stored in k8s secrets, can we mention this in the name of the function in some way? 😅 e.g. `enrichPodWithGoogleApplicationCredentials`, `mountGoogleApplicationCredentialsToPod` ...

api/pkg/observability/deployment/identity.go
```diff
@@ -0,0 +1,52 @@
...
+	secretVolume := createVolumeFromSecret(secretName)
+	updatedPodSpec := podSpec.DeepCopy()
+
+	containerExist := false
+	for idx, containerSpec := range updatedPodSpec.Containers {
+		updatedContainerSpec := containerSpec
+		for _, targetName := range containerNames {
+			if targetName != updatedContainerSpec.Name {
+				continue
+			}
```
deadlycoconuts, line 19: nit: We probably aren't expecting the number of elements in `containerNames` to be super long, but can we make `containerNames` into something like a set (e.g. `map[string]struct{}`, or the set type from the [golang-collections](https://pkg.go.dev/github.com/golang-collections/collections/set) package, etc.) to check for the presence of containers in it instead of looping through a list? It feels sad to be looping through nested loops 😥

deadlycoconuts (approved):
Thanks a lot for the changes @tiopramayudi ! I just left a couple of tiny comments but the main one's about removing the support for workload identity completely 🤔 (If it's not the scope of this MR or if it's something that isn't planned it's okay too 😅)

api/pkg/observability/deployment/identity.go
tiopramayudi: ah previously I planned to use an interface that has multiple implementation for the different cloud providers, but to make it simpler I only created utility method with general name. I can change the name for now

api/pkg/observability/deployment/identity.go
tiopramayudi: this is initially to copy the value of the containerSpec but this is necessary due to deepCopy of root object in the above, I'll update the code

api/config/observability.go
tiopramayudi: yes for now, we can add this again into shared library that can enrichPodIdentity based on some config e.g cloud provider, workload identity flag

PR #608, opened 2024-09-20
fix(batch-predictor, pyfunc-server): Pin setuptools max version

deadlycoconuts:
# Description
It seems like from the 20th of September 2024, CICD tasks that require building a python environment started [failing](https://github.com/caraml-dev/merlin/actions/runs/10938668959/job/30411359538#step:7:2657) with the following error when the ancient `mlflow` package tries to log artifacts:
```
ModuleNotFoundError: No module named 'distutils._modified'
```

This seems to be an error [reported](https://github.com/pyinstaller/pyinstaller/issues/8793) recently by other people in the `pyinstaller` repo, with some users mentioning that `distutils` is used by `setuptools`, which just had a new version released recently. Since it doesn't seem like we're importing `distutils` directly, I've decided to simply pin down the max version of `setuptools` to `<75`. Note that the version of `setuputils` somehow needs to be specified in the variable passed to this argument `install_requires` here 

https://github.com/caraml-dev/merlin/blob/a9184ede2126d71c84f6458bcc44515dd78b0d03/python/pyfunc-server/setup.py#L49

As such, I've added `setuptools<75` in both the batch predictor and pyfunc server's requirements.txt file.

Note that this PR technically does not fix the root cause of the problem; it's only a band-aid solution for the moment until we migrate to newer versions of Mlflow **_very shortly_**. When that happens I'll revert the changes in this PR.

# Modifications
- `.github/workflows/release.yml` - Updated the versions of `setuptools` in the build jobs for the python-related components
- `python/batch-predictor/requirements.txt` - Added `setuptools<75` to the requirements of the batch predictor
- `python/pyfunc-server/requirements.txt` - Added `setuptools<75` to the requirements of the pyfunc server

# Tests

# Checklist
- [x] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
           cache: pip
       - name: Install dependencies
         working-directory: ./python/sdk
-        run: pip install setuptools>=64 setuptools_scm>=8 twine wheel
+        run: pip install setuptools>=64,<75 setuptools_scm>=8 twine wheel
       - name: Build and publish
         env:
           TWINE_USERNAME: ${{ secrets.pypi_username }}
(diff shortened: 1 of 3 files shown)


tiopramayudi (approved):
LGTM! Thanks @deadlycoconuts

deadlycoconuts:
Thanks for the quick review @tiopramayudi ! I'll merge this now!

PR #611, opened 2024-09-25
docs: add documentation for webhook

bthari:
# Description
Adding documentation about the webhook created in this PR #601; how to add the config, what event is available, and the information about the payload. 

# Modifications

# Tests

# Checklist
- [x] Added PR label
- [ ] Tested locally
- [x] Updated documentation

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/user/templates/webhook/01_webhook_config.md
+++ b/docs/user/templates/webhook/01_webhook_config.md
@@ -5,14 +5,19 @@
 When an event happens (e.g. creating new model's version, deploy endpoint), we might want to trigger action in other services, Merlin provides a way this by triggering a webhook call. 
 The webhook that will be called is configured per event type, and it could contain multiple endpoints which can be configured as synchronous or asynchronous call. 
 
-For the detail of available configuration please refer to [MLP library docs](https://github.com/caraml-dev/mlp/blob/main/api/pkg/webhooks/README.md), this documentation will only cover the basic config and the paylod of webhook which we has implemented.
+For the detail of available configuration please refer to [MLP library docs](https://github.com/caraml-dev/mlp/blob/main/api/pkg/webhooks/README.md), this documentation will only cover the basic config and the payload of webhook which we have implemented.
 
-## 1. Webhook Configuration
+## 1. Webhook Configuration 
 In Merlin's yaml for config you can add new field for the webhook's configuration which will looks like this:
 ```yaml
 WebhooksConfig:
   Enabled: true
   Config:
+    on-model-version-predeployment:
+      - URL: http://127.0.0.1:8000/async-webhook
+        Method: POST
+        Async: true
+        Name: async
     on-model-version-deployed:
       - URL: http://127.0.0.1:8000/async-webhook
         Method: POST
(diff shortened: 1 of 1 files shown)


docs/user/templates/webhook/01_webhook_config.md
```diff
@@ -0,0 +1,75 @@
...
+The webhook that will be called is configured per event type, and it could contain multiple endpoints which can be configured as synchronous or asynchronous call. 
+
+For the detail of available configuration please refer to [MLP library docs](https://github.com/caraml-dev/mlp/blob/main/api/pkg/webhooks/README.md), this documentation will only cover the basic config and the paylod of webhook which we has implemented.
+
+## 1. Webhook Configuration
+In Merlin's yaml for config you can add new field for the webhook's configuration which will looks like this:
+```yaml
+WebhooksConfig:
+  Enabled: true
+  Config:
```
deadlycoconuts, line 15: Can we give an example of `on-model-version-predeployment` here too? Just so it's clearer that this webhook exists.

docs/user/templates/webhook/01_webhook_config.md
```diff
@@ -0,0 +1,75 @@
...
+        Method: POST
+        FinalResponse: true
+        Name: sync
+```
+## 2. Webhook Events and Payload
+### 2.1. Model Version's Endpoint
+**Events:**
+- `on-model-version-predeployment`: will be triggered before the version endpoint is deployed, if the sync webhook return an error (non `200` response), the deployment will be **aborted**. The successful response of this webhook will not affect the deployment process, as we do not process the response
+- `on-model-version-deployed`: will be triggered after the version endpoint is successfully deployed (either new or updated)
+- `on-model-version-undeployed`: will be triggered after the version endpoint is successfully undeployed
```
deadlycoconuts, line 32: Tiny nit on this - could we first specify what is expected to happen normally (i.e. 200 response) before explaining what would happen in a scenario where a webhook returns an error? 😅 For example,

- `on-model-version-predeployment` - will be triggered before the version endpoint is deployed, and requires a 200 response from the webhook for the deployment to proceed. If any of the sync webhooks return an error (non `200` response), the deployment will be aborted. In both cases the responses of the webhooks will not be processed.

We can do the same for the other 2 descriptions for `on-model-version-deployed` and `on-model-version-undeployed` too (they don't have information on what the outcome would be if the webhooks return a success or error).

docs/user/templates/webhook/01_webhook_config.md
```diff
@@ -0,0 +1,75 @@
...
+  "version_endpoint": {
+    "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+    "version_id": 0,
+    "status": "pending",
+    ...
+  }
+}
+```
+
+For the the complete field `version_endpoint` please refer to our [swagger docs](https://github.com/caraml-dev/merlin/blob/main/swagger.yaml).
```
deadlycoconuts, line 55: Nit: We can actually use a relative link instead of an absolute link for this: https://github.blog/news-insights/product-news/relative-links-in-markup-files/.

docs/user/templates/webhook/01_webhook_config.md
```diff
@@ -0,0 +1,75 @@
...
+
+**Response**
+
+As of writing this document (25 September 2024) and mentioned previously, the successful response will not affect the process of our deployment. But for future usage, we're expecting the response of the final synchronous webhook that will be called to be in the following format:
```
deadlycoconuts, line 59: Thanks for adding the date 📅 😄 Makes it look neat!

deadlycoconuts (approved):
Thanks a lot for this PR @bthari ! The docs really make the webhooks part of Melrin much clearer so that's a lot for that! 🚀🙏🏼 I just added some minor comments here and there but feel free to merge this PR once you think you're done addressing the changes!

bthari:
Thank you @deadlycoconuts for the review and giving suggestions to make the explanation clearer!

PR #612, opened 2024-10-07
feat: add webhook call to other events

bthari:
# Description
Previously on #601 I added webhook for model version endpoint related event, and in here the event will be expanded into a model, model endpoint, model version related event, as we also want to have an action (from other service) to be triggered if these events happen. 

# Modifications

- created another package for webhook interface 
- add event for:
  - model created
  - model endpoint created/updated/deleted
  - model version created/updated/deleted
- change previous event name of `on-model-version-*` to `on-version-endpoint-*` 

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Notes for Reviewer
The version of MLP used here has a validation bug (which is updated on [MLP#117](https://github.com/caraml-dev/mlp/pull/117)). The code could still work with workaround (e.g. set `FinalResponse: true` in one async webhook if user use _all_ async webhook, but it will be confusing for user since async webhook response is expected to not be used anywhere), ~~so preferably to merge this PR after updating the MLP version as dependencies.~~ (MLP version will be updated with the s3 PR) 

# Release Notes

```release-note
add webhook event call if there's changes on model, model endpoint, model version
```


api/api/model_endpoints_api.go
```diff
@@ -124,6 +125,9 @@ func (c *ModelEndpointsController) CreateModelEndpoint(r *http.Request, vars map
 		return InternalServerError(fmt.Sprintf("Error creating model endpoint: %v", err))
 	}
 
+	// trigger webhook call
+	_ = c.Webhook.TriggerModelEndpointEvent(ctx, webhook.OnModelEndpointCreated, endpoint)
```
bayu-aditya, line 129: seems for all webhook trigger error is ignored. Is it expected? 
How about if we add log warning while the webhook error?

api/webhook/webhook.go
```diff
@@ -0,0 +1,141 @@
...
+	OnModelVersionCreated,
+	OnModelVersionUpdated,
+	OnModelVersionDeleted,
+	OnVersionEndpointPredeployment,
+	OnVersionEndpointDeployed,
+	OnVersionEndpointUndeployed,
+}
+
+type Webhook struct {
+	WebhookManager webhookManager.WebhookManager
```
bayu-aditya, line 42: seems we can make `WebhookManager` become private, wdyt?

api/api/model_endpoints_api.go
bthari: Yes it's expected because as of now, the usage of the webhook doesn't modify the resource in Merlin, meaning if there's error happening the webhook call, we do not want to rollback/cancel the changes of the resource. 

And the log warning is inside the `TriggerModel`, I think it suffice?

api/queue/work/model_service_deployment_test.go
```diff
@@ -712,8 +714,10 @@ func TestExecuteDeployment(t *testing.T) {
 				mockImgBuilder.On("BuildImage", context.Background(), mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return("", errors.New("Failed to build image"))
 				return mockImgBuilder
 			},
-			webhookManager: func() webhookManager.WebhookManager {
-				return webhookManager.NewMockWebhookManager(t)
+			webhook: func() *webhookMock.Client {
+				//w := webhookMock.NewClient(t)
```
tiopramayudi, line 718: if not required can u remove this?

api/queue/work/model_service_deployment_test.go
```diff
@@ -672,8 +672,10 @@ func TestExecuteDeployment(t *testing.T) {
 				mockImgBuilder := &imageBuilderMock.ImageBuilder{}
 				return mockImgBuilder
 			},
-			webhookManager: func() webhookManager.WebhookManager {
-				return webhookManager.NewMockWebhookManager(t)
+			webhook: func() *webhookMock.Client {
+				//w := webhookMock.NewClient(t)
+				//w.On("TriggerVersionEndpointEvent", mock.Anything, webhook.OnVersionEndpointDeployed, mock.Anything).Return(nil)
```
tiopramayudi, line 677: if not required can u remove this?

api/webhook/webhook.go
```diff
@@ -0,0 +1,141 @@
...
+	"context"
+
+	"github.com/caraml-dev/merlin/log"
+	"github.com/caraml-dev/merlin/models"
+	"github.com/caraml-dev/mlp/api/pkg/webhooks"
+	webhookManager "github.com/caraml-dev/mlp/api/pkg/webhooks"
+)
+
+var (
+	OnModelCreated = webhooks.EventType("on-model-created")
```
tiopramayudi, line 13: if we have a dedicated function for each event do we need to expose this?

bthari: made changes for this! there will only be one function in interface which will be used by all event. please recheck it, thank you

api/api/model_endpoints_api.go
bthari: I've changed the way this interface work now. Please check it again :D There will only be a general function that can be called by all event, with option to setup the body, success callback, and error callback (by default the error will be default callback from the library). 

The error is now returned and processed, most of it will just log it with a `warning` level thought. If we do want to fail it or return the error to user, I believe we need a mechanism to rollback the creation/update of resource (because it will be weird if we simply return an error but actually the creation is success), perhaps later if we have a clear usecase of how we're going to fail/rollback this, we can update it again

tiopramayudi (approved):
LGTM, thanks @bthari !

PR #615, opened 2024-11-05
feat(batch-predictor): remove CPU limits from SparkApplication resource

mbruner:
# Description

Similar to kserve models, this PR  removes CPU limits for batch prediction jobs. If limits are set, executors will experience significant throttling because batch jobs tend to consume all available CPU by its nature.

# Modifications
* SparkApplication resource no longer has `coreLimit` set in driver and executor specs. This field is optional according to [the specs](https://github.com/kubeflow/spark-operator/blob/master/docs/api-docs.md#sparkoperator.k8s.io/v1beta2.SparkPodSpec) and it should be safe to simply remove it.

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
Remove CPU Limits for batch prediction jobs
```

### DIFF
--- a/api/batch/resource.go
+++ b/api/batch/resource.go
@@ -143,7 +143,7 @@ func (t *BatchJobTemplater) createDriverSpec(job *models.PredictionJob) (v1beta2
 	}
 
 	core := getCoreRequest(userCPURequest)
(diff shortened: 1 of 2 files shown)


deadlycoconuts (approved):

PR #616, opened 2024-11-14
chore(sdk): Update mlflow version in sdk

deadlycoconuts:
# Description
This PR is the **first part** (1 out of 3) of a series of PRs to update the version of Mlflow used by Merlin (to `1.26.1`*). These changes are:

2. Update 

I also made some minor refactoring to the requirements files so that they are stored in a separate file and the reading of these files is consistent with what is done in the batch predictor and pyfunc server `setup.py` files.

*I've selected `1.26.1` because it's the latest version of Mlflow available that is before Mlflow v2, that still allows `pandas==2.2.0` to be installed (this is a hard requirement needed by the `arize` package).

# Modifications
- Update Mlflow version to `1.26.1` (and other dependencies) in the SDK as well as observation publisher's requirements file since the unit tests there import the SDK's requirements file's dependencies
- ~~Update unit tests to ensure that the updated `PythonModelContext` from Mlflow is initialised with the required arguments~~ (will only be needed for newer versions of Mlflow)

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
NONE
```


python/sdk/setup.py
```diff
@@ -22,49 +22,11 @@
...
-    "python_dateutil>=2.5.3",
-    "PyYAML>=5.4",
-    "six>=1.10",
-    "urllib3>=1.26",
-    "numpy<=1.23.5",  # Temporary pin numpy due to https://numpy.org/doc/stable/release/1.20.0-notes.html#numpy-1-20-0-release-notes
-    "caraml-auth-google==0.0.0.post7",
-    "pydantic==2.5.3"
-]
+with open("requirements.txt") as f:
+    REQUIRE = f.read().splitlines()
```
deadlycoconuts, line 26: Refactored the listing of requirements away into a separate file, just like how it's done in the batch predictor and pyfunc server packages (see https://github.com/caraml-dev/merlin/blob/main/python/batch-predictor/setup.py#L24 for an example).

mbruner (approved):
lgtm, thanks!

deadlycoconuts:
Thanks for the quick review. I've lowered the upgrade back to 1.26.1 because it's latest version (below version 2) that does not have any restrictions preventing `pandas==2.2.0` from being installed (which is a requirement of the `arize` package we're using). 

I'll let the e2e tests pass and merge this after that.

PR #617, opened 2024-11-14
feat(mlflow): Update Mlflow, API server, pyfunc server and batch predictor

deadlycoconuts:
# Description
This PR is the **second part** (2 out of 3) of a series of PRs to update the version of Mlflow used by Merlin (to `1.26.1`*). These changes are:

2. Update 

# Main Modifications
- `.github/workflows/external.yml` - Updated the CICD job to release a new version of the Mlflow image
- `api/mlflow/mlflow.go` - Added a header in all requests sent to the Mlflow server from the Merlin API server because it's now required
- `api/mlflow/response.go` - Changed the data type of certain fields in Mlflow's responses sent to Merlin
- `mlflow/Dockerfile` - Updated Dockerfile of the Mlflow server with newer dependencies
- `python/batch-predictor/requirements.txt` - Updated the requirements file with a newer version of the SDK which requires version `1.26.1` of Mlflow
- `python/pyfunc-server/requirements.txt` - Updated the requirements file with a newer version of the SDK which requires version `1.26.1` of Mlflow

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
NONE
```


.github/workflows/external.yml
```diff
@@ -22,9 +22,10 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push MLflow Docker image
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v6
         with:
           context: mlflow
+          push: true
```
deadlycoconuts, line 28: This is now needed for newer versions of the `docker/build-push-action` (version 5 also requires this but since we haven't been pushing new mlflow images for years no one noticed this) since the default value is `false` which causes the image to be built but not pushed. See https://github.com/docker/build-push-action?tab=readme-ov-file#customizing for more details.

.github/workflows/merlin.yml
```diff
@@ -400,8 +400,9 @@ jobs:
       K3D_CLUSTER: merlin-cluster
       LOCAL_REGISTRY_PORT: 12345
       LOCAL_REGISTRY: "dev.localhost"
+      DOCKER_REGISTRY: "dev.localhost:12345"
       INGRESS_HOST: "127.0.0.1.nip.io"
-      MERLIN_CHART_VERSION: 0.13.4
+      MERLIN_CHART_VERSION: 0.13.18
```
deadlycoconuts, line 405: The e2e tests were using a relatively older version of the Merlin helm chart and I've updated it to the latest version.

api/mlflow/mlflow.go
```diff
@@ -61,6 +61,7 @@ func (mlflow *client) doCall(req *request, resp interface{}) error {
 		return err
 	}
 
+	httpReq.Header.Set("Content-Type", "application/json")
```
deadlycoconuts, line 64: This is now required to be explicitly specified when sending requests to the Mlflow server https://mlflow.org/docs/latest/deployment/deploy-model-locally.html#accepted-input-formats.

scripts/e2e/deploy-merlin.sh
```diff
@@ -28,7 +28,7 @@ install_merlin() {
     --set deployment.image.registry=${DOCKER_REGISTRY} \
     --set deployment.image.repository=merlin \
     --set deployment.image.tag=${VERSION} \
-    --set transformer.image=${DOCKER_REGISTRY}/merlin-transformer:${VERSION} \
+    --set rendered.overrides.StandardTransformerConfig.ImageName=${DOCKER_REGISTRY}/merlin-transformer:${VERSION} \
```
deadlycoconuts, line 31: Previously the e2e tests weren't even correctly using the temporarily built transformer image that's pushed to the local image registry just before the tests were run. They were instead using an already released version of the image that's much older (0.35.0) https://github.com/caraml-dev/merlin/actions/runs/11853812453/job/33034971181#step:12:2147. This fix ensures that only the locally built and pushed transformer image is used https://github.com/caraml-dev/merlin/actions/runs/11866485521/job/33073372549#step:12:2111

scripts/e2e/deploy-merlin.sh
```diff
@@ -48,6 +48,8 @@ install_merlin() {
     --set imageBuilder.predictionJobBaseImages."3\.10\.*".mainAppPath=/home/spark/merlin-spark-app/main.py \
     --set ingress.host=merlin.caraml.${INGRESS_HOST} \
     --set mlflow.ingress.host=merlin-mlflow.caraml.${INGRESS_HOST} \
+    --set mlflow.image.repository=caraml-dev/mlflow \
+    --set mlflow.image.tag=1.26.1 \
```
deadlycoconuts, line 52: These tags need to be manually set for now since the Merlin helm chart hasn't been updated to use newer versions of Mlflow.

scripts/e2e/setup-cluster.sh
```diff
@@ -103,6 +103,12 @@ install_knative() {
     kubectl rollout status deployment/domainmapping-webhook -n knative-serving -w --timeout=${TIMEOUT}
     kubectl rollout status deployment/webhook -n knative-serving -w --timeout=${TIMEOUT}
 
+    # Update knative config-deployment config map to allow resolving of the local e2e image repository
+    kubectl get configmap -n knative-serving config-deployment -o yaml > temp.yaml
+    yq -i '.data.registries-skipping-tag-resolving |= ("\(.),${DOCKER_REGISTRY}" | envsubst)' temp.yaml
+    kubectl apply -f temp.yaml
+    kubectl rollout restart deployment -n knative-serving controller
```
deadlycoconuts, line 110: These additional steps are needed to ensure Knative does not perform additional steps to resolve the local e2e image registry (the image already exists but Knative attempts to connect to the registry as an extra 
step before it pulls any image from it). This error shows up when that happens https://github.com/caraml-dev/merlin/actions/runs/11861830172/job/33060193167#step:17:7853. See https://github.com/knative/serving/issues/14117 and https://stackoverflow.com/questions/73619481/unable-to-pull-from-knative-but-able-from-a-normal-deploy for more details.

mlflow/Dockerfile
```diff
@@ -9,11 +9,11 @@ RUN apt-get update && \
...
 RUN pip install google-cloud-storage
-RUN pip install psycopg2-binary==2.8.6
+RUN pip install psycopg2-binary==2.9.10
 
-ARG BOTO3_VERSION=1.7.12
-ARG MLFLOW_VERSION=1.3.0
-ARG SQLALCHEMY_VERSION=1.4.46
+ARG BOTO3_VERSION=1.35.39
+ARG MLFLOW_VERSION=1.26.1
+ARG SQLALCHEMY_VERSION=1.4.54
```
shydefoo, line 16: Are these version bumps all related to new version of mlflow? Or we're just bumping them to use a newer version?

shydefoo (approved):
LGTM, thanks for the detailed notes in the PR explaining each step.

mlflow/Dockerfile
deadlycoconuts: I'm just bumping them to use a later version for now (the versions originally specified will block later versions of mlflow past version 1.26.1 to be installed). SQL alchemy's intentionally kept below version 2 because version 1.26.1 of Mlflow is only compatible with version 1 of it.

deadlycoconuts:
Thanks a lot for the quick review! :D I'll merge this soon!

PR #618, opened 2024-11-18
chore(sdk): Update pyfunc server and batch predictor default version

deadlycoconuts:
# Description
This PR is the **third part** (3 out of 3) of a series of PRs to update the version of Mlflow used by Merlin (to `1.26.1`*). These changes are:

2. Update 

Since the pyfunc server and batch predictor versions have already been updated in the PR #617, we need to ensure that the SDK is able to use these newly released versions by default.

# Modifications
- `python/sdk/merlin/model.py` - Update the version bounds of the default pyfunc server and batch predictor version in the SDK

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
NONE
```

### DIFF
--- a/python/sdk/merlin/model.py
+++ b/python/sdk/merlin/model.py
@@ -1005,9 +1005,9 @@ class ModelVersion:
                 "log_pyfunc_model is only for PyFunc, PyFuncV2 and PyFuncV3 model"
             )
 
-        merlin_requirements = ["merlin-pyfunc-server<0.42.0"]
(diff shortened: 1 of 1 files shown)


tiopramayudi (approved):
LGTM! Thanks @deadlycoconuts

deadlycoconuts:
Thanks for the quick review! Yeah in the future maybe we need to relook at the bumping of versions manually. I honestly can't think of any reason why we shouldn't bump this automatically the more I think about it 🤔

PR #605, opened 2024-09-12
feat(api): Refactor API server to introduce support for Docker image registries and s3 bucket support

deadlycoconuts:
## Note 🚨
~~This PR should not be merged without the changes in https://github.com/caraml-dev/mlp/pull/116 being merged, published and imported as dependencies first. This PR is currently only using a branch of a fork (the source of the quoted PR) of the MLP repository.~~ The dependent PR has been merged.

# Description
In order to provide support for using image registries that use [Docker registry credentials](https://docs.docker.com/reference/cli/docker/login/#configure-the-credential-store) as well as [AWS S3-based blob storage services](https://docs.aws.amazon.com/code-library/latest/ug/go_2_s3_code_examples.html), this PR refactors the API server to support these additional image registry and bob storage options. More concretely, these are the following changes made:

- Set up the workflow needed to allow platform maintainers to configure the image registry that the API server will push images to (Docker or Google Cloud/Artifact Registry) as well as the blob storage service that it should read model artifacts from/write files to (S3-based store/Google Cloud Storage)
- Allow the API server to access a configured Docker registry to check if an image is available
- Allow the API server to check and hash model dependencies in a configured S3-based store
- Allow Kaniko jobs spun up by the API server to use load model artifacts from a configured S3-based store when building model images
- Allow Kaniko jobs spun up by the API server to push images to the configured Docker registry

# Modifications
- `api/cmd/api/setup.go` - Make the initialisation of the image builder set up the artifact service type and docker registry correctly depending on the one set up
- `api/config/config.go` - Introduce new configs for platform maintainers to specify the `KanikoPushRegistryType` and the `KanikoDockerCredentialSecretName`
- `api/pkg/imagebuilder/imagebuilder.go` - Make changes to the image builder to configure the Kaniko job spec correctly depending on the selected registry type and blob storage type
- `python/batch-predictor/docker/app.Dockerfile` - Add steps to the batch predictor docker image to authenticate and pull model artifacts correctly depending on the configured blob storage type 
- `python/batch-predictor/docker/base.Dockerfile` - Add steps to the base batch predictor image to install the AWS CLI
- `python/pyfunc-server/docker/Dockerfile` - Add steps to the pyfunc server docker image to authenticate and pull model artifacts correctly depending on the configured blob storage type
- `python/pyfunc-server/docker/base.Dockerfile` - Add steps to the base pyfunc server image to install the AWS CLI

# Tests

# Checklist
- [x] Added PR label
[...]

PR #621, opened 2024-12-20
feat(image-builder): Add field to allow api server env vars to be propagated to kaniko jobs

deadlycoconuts:
# Description
This PR introduces a new mechanism for environment variables from the Merlin API server to be propagated to **_the build environment of the Kaniko build jobs_** that it spins up, reducing the need for redundant repetition of configuration, especially if these environment variables are common to both the Merlin API server and the build environment of the Kaniko build jobs.

Since these variables are passed to the Kaniko build jobs as build arguments (as opposed to environment variables of the container where the image gets built; see https://github.com/GoogleContainerTools/kaniko/issues/2824 for more details), these variables get passed as additional arguments in the Kaniko build job.

This PR additionally introduces a cluster connectivity check (done by a simple list pods operation) for the image builder to ensure that the cluster controller has been successfully initialised at start up of the Merlin API server, instead of allowing errors to only be thrown when an image building job has been triggered.

# Modifications
- `api/cmd/api/setup.go` - Addition of a connectivity check during initialisation of the image builder
- `api/pkg/imagebuilder/imagebuilder.go` - Addition of a step to add Merlin API server environment variables as Kaniko build args

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
NONE
```

### DIFF
--- a/api/config/config.go
+++ b/api/config/config.go
@@ -459,8 +459,11 @@ type MlflowConfig struct {
 	// Note that the Kaniko image builder needs to be configured correctly to have the necessary credentials to download
 	// the artifacts from the blob storage tool depending on the artifact service type selected (gcs/s3). For gcs, the
 	// credentials can be provided via a k8s service account or a secret but for s3, the credentials can be provided via
-	// additional arguments in the config KanikoAdditionalArgs e.g.
+	// 1) additional arguments in the config KanikoAdditionalArgs e.g.
 	// --build-arg=[AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY/AWS_DEFAULT_REGION/AWS_ENDPOINT_URL]=xxx
+	// OR
+	// 2) additional arguments in the config KanikoAPIServerEnvVars, which will pass the specified environment variables
+	// PRESENT within the Merlin API server's container to the image builder as build arguments
 	ArtifactServiceType string `validate:"required,oneof=nop gcs s3"`
 }
 


api/pkg/imagebuilder/imagebuilder.go
```diff
@@ -638,12 +639,17 @@ func (c *imageBuilder) createKanikoJobSpec(
...
 	var volumes []v1.Volume
 	var volumeMounts []v1.VolumeMount
-	var envVar []v1.EnvVar
+	var envVars []v1.EnvVar
 
 	// Configure additional credentials for specific image registries and artifact services
 	kanikoArgs = c.configureKanikoArgsToAddCredentials(kanikoArgs)
 	volumes, volumeMounts = c.configureVolumesAndVolumeMountsToAddCredentials(volumes, volumeMounts)
-	envVar = c.configureEnvVarsToAddCredentials(envVar)
+	envVars = c.configureEnvVarsToAddCredentials(envVars)
```
bthari, line 647: nit: I assume this is to get the S3 credential, but cmiiw, based on the previous #605 MR, the args is expected to be passed in the KanikoAdditionalArgs, so perhaps can modify [this chunk of note](https://github.com/caraml-dev/merlin/blob/main/api/config/config.go#L461-L462), if we're going to change it this way?

bthari (approved):
Just one tiny comment, but the rest LGTM, so feel free to merge it once it's addressed. Thank you for the changes and improvement! 🚀

api/pkg/imagebuilder/imagebuilder.go
deadlycoconuts: Ahh, yes the motivation was to allow the image building job to get access to the S3 credentials. Technically it's still possible to pass the credentials via `KanikoAdditionalArgs` but the only downside is that the secret might be exposed if `KanikoAdditionalArgs` is specified explicitly within a certain repo (like what we do). I'll reword the comments you mentioned a little to suggest multiple ways to pass the S3 credentials (including the one introduced in this PR). Thanks for suggesting this!

deadlycoconuts:
Thanks for the quick review and for spotting the issue with the comments! Merging this now! 🚀

PR #625, opened 2025-01-28
feat(api,sdk): Add user-configurable secrets for deployments

deadlycoconuts:
# Description
This PR introduces a new feature that allows users to mount secrets stored in MLP into their Merlin batch prediction workloads and Merlin model deployments. This is similar to the existing feature whereby users are able to mount their Google Cloud Platform's service account key stored in MLP as a secret within the batch workloads. 

## Batch jobs
The user-configured MLP secrets are basically stored within the [existing secret that gets created](https://github.com/caraml-dev/merlin/blob/967788eb98a0120dd20bec692ab04bcea8f08fc9/api/batch/controller.go#L166) for each batch job run, before the Spark application gets created. The Spark driver and executor access the user-configured MLP secrets as env vars that have references to this k8s secret. This secret similarly gets deleted after completion of the job.

## Model deployments
No existing process exists to create a k8s secret prior to the deployment of a KServe inference service so this PR introduces some changes to the deployment workflow to create this secret (not unlike what happens for batch jobs). This secret gets deleted when the model is undeployed, when there's a new successful deployment or when the deployment fails. Similar to what happens for batch jobs, the KService predictor (Merlin model) and transformer then access the user-configured MLP secrets as env vars that have references to this k8s secret.

## DB migration
A small DB migration is required to create new columns in the `version_endpoints` and `transformers` table to store the configured MLP secrets. The batch jobs don't require this update as those secrets are stored within an existing column that's already of the `jsonb` type.

This PR only introduces the necessary API server and SDK changes to make this happen. The UI changes will be introduced in a separate PR to keep this PR short. 

# Modifications
- `api/batch/controller.go` - [Batch jobs] Minor refactoring to retrieve GCP service account secret and user-configured secrets from MLP for 
- `api/batch/resource.go` - [Batch jobs] Addition of env vars with references to a k8s secret (containing all the user-configured secrets) to the spark driver and executors
- `api/cluster/controller.go` - [Model deployments] Minor refactoring of the existing cluster controller so that it now contains an MLP API client object to retrieve MLP secrets with; introduction of additional helper methods to create/delete secrets for predictors and transformers
- `api/cluster/resource/templater.go` - [Model deployments] Refactoring of the KServe inference service manifest templater to add user-configured secrets as env vars with references to a k8s secret containing all the user-configured secrets)
- `api/cluster/secret.go` - [Model deployments] Addition of new k8s secret creation/deletion methods using the cluster controller
- `api/cmd/api/main.go` - Minor refactoring to parse the MLP API client to cluster controllers initialised
- `api/models/prediction_job.go` - [Batch jobs] Addition of the secrets field to batch jobs 
- `api/models/secret.go` - Addition of the new struct `Secret` that contains information about the name of the secret as stored in MLP as well as the name of the environment variable the secret will be available as
- `api/models/service.go` - [Model deployments] Addition of the secrets field to Merlin model services (this is just an intermediate struct used by Merlin in the API server)
- `api/models/transformer.go` - [Model deployments] Addition of the secrets field to Merlin transformers (this struct gets stored in the DB)
- `api/models/version_endpoint.go` - [Model deployments] Addition of the secrets field to Merlin version endpoints (this struct gets stored in the DB)
- `db-migrations/41_version_endpoints_add_secrets.**.sql` - Addition of DB migration scripts to add a new column for secrets to be stored
- `python/sdk/merlin/endpoint.py` - Addition of steps to allow secrets to be initialised correctly in a `VersionEndpoint` object
- `python/sdk/merlin/model.py` - Addition of steps to allow secrets to be correctly parsed when the `deploy` function is called
- `python/sdk/merlin/mounted_mlp_secret.py` - Addition of a new class to represent each user-configurable secret
- `swagger.yaml` - Update to swagger API specs

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
NONE; DATABASE MIGRATION REQUIRED
```


api/cluster/resource/templater.go
```diff
@@ -918,13 +921,38 @@ func (t *InferenceServiceTemplater) getResourceLimitsAndEnvVars(
...
+			if index, exist := envIndexMap[add.Name]; exist {
+				left[index].Value = add.Value
+				left[index].ValueFrom = add.ValueFrom
+			} else {
+				left = append(left, add)
+			}
+		}
+	}
+	return left
 }
```
deadlycoconuts, line 950: This is basically just a copy of whatever already has been implemented [here](https://github.com/caraml-dev/merlin/blob/main/api/models/env_var.go#L75). Basically I've refactored the code a little to perform merging of lists of env vars with the final `corev1.EnvVar` datatype (used by k8s, kserve, etc.) instead of the intermediate `model.EnvVar` datatype (only used by Merlin) in order to simplify the merging process. Some env vars do not come from the `model.EnvVar` datatype (like env vars with references to secrets) and it's weird to perform merging only with the `model.EnvVar` datatype. 

This simplification also removes the need for the `ParseEnvVars` helper function below.

api/cmd/api/setup.go
```diff
@@ -46,7 +46,7 @@ type deps struct {
 	imageBuilderJanitor     *imagebuilder.Janitor
 }
 
-func initMLPAPIClient(ctx context.Context, cfg config.MlpAPIConfig) mlp.APIClient {
+func initMLPAPIClient(cfg config.MlpAPIConfig) mlp.APIClient {
```
deadlycoconuts, line 49: Removed the `ctx` argument because it is unused.

api/cluster/controller.go
```diff
@@ -210,6 +215,11 @@ func (c *controller) Deploy(ctx context.Context, modelService *models.Service) (
 		return nil, errors.Wrapf(err, fmt.Sprintf("%v (%s)", ErrUnableToCreateNamespace, modelService.Namespace))
 	}
 
+	err = c.createSecrets(ctx, modelService, projectID)
+	if err != nil {
+		return nil, fmt.Errorf("failed creating secret for deployment %s: %w", modelService.Name, err)
+	}
```
deadlycoconuts, line 221: During deploy time, we create a secret containing all the MLP secrets beforehand. This secret gets deleted when the model gets undeployed, updated, or when the deployment fails.

api/batch/controller.go
```diff
@@ -320,6 +321,27 @@ func (c *controller) onUpdate(old, new interface{}) {
 	}
 }
 
+func (c *controller) getMLPSecrets(ctx context.Context, predictionJob *models.PredictionJob, namespace string) (map[string]string, error) {
+	secretMap := make(map[string]string)
+	// Retrieve Google Service Account secret from MLP
+	googleServiceAccountSecret, err := c.mlpAPIClient.GetSecretByName(ctx, predictionJob.Config.ServiceAccountName, int32(predictionJob.ProjectID))
```
shydefoo, line 327: Q: Is there always a serviceaccount attached to the predictionJob config by default?

api/batch/manifest.go
```diff
@@ -219,16 +219,14 @@ func (m *manifestManager) DeleteDriverAuthorization(ctx context.Context, namespa
 	return nil
 }
 
-func (m *manifestManager) CreateSecret(ctx context.Context, predictionJobName string, namespace string, data string) (string, error) {
+func (m *manifestManager) CreateSecret(ctx context.Context, predictionJobName string, namespace string, secretMap map[string]string) (string, error) {
```
shydefoo, line 222: nit: same as before to prevent confusion between method names xD

api/batch/controller.go
```diff
@@ -158,12 +158,13 @@ func (c *controller) Submit(ctx context.Context, predictionJob *models.Predictio
...
-	secret, err := c.mlpAPIClient.GetSecretByName(ctx, predictionJob.Config.ServiceAccountName, int32(predictionJob.ProjectID))
+	// Get MLP secrets for Google service account and user-specified MLP secrets
+	secretMap, err := c.getMLPSecrets(ctx, predictionJob, namespace)
 	if err != nil {
-		return fmt.Errorf("service account %s is not found within %s project: %w", predictionJob.Config.ServiceAccountName, namespace, err)
+		return fmt.Errorf("error retrieving secrets: %w", err)
 	}
 
-	_, err = c.manifestManager.CreateSecret(ctx, predictionJob.Name, namespace, secret.Data)
+	_, err = c.manifestManager.CreateSecret(ctx, predictionJob.Name, namespace, secretMap)
```
shydefoo, line 167: Q: do different batch prediction jobs have unique names? In case a new batch prediction job is submitted before the previous one completes, the secret names should be unique right?

api/service/version_endpoint_service.go
```diff
@@ -246,6 +246,12 @@ func (k *endpointService) override(left *models.VersionEndpoint, right *models.V
 		left.EnvVars = right.EnvVars
 	}
 
+	// override secrets
+	// Configure environment variables for Pyfunc model
+	if right.Secrets != nil {
```
shydefoo, line 251: Q: So if there are no secrets the new deployment will reuse the old secrets?

shydefoo (changes requested):

shydefoo (commented):
Thanks for the quick and elegant solution. Just some small comments and questions!

shydefoo (commented):
Thanks for the very quick and elegant solutiont to this! Left some small comments and questions.

api/batch/controller.go
deadlycoconuts: Yep. Hmm, I wouldn't say by default, but rather, requests to the API server to trigger prediction jobs are expected to have a service account name specified. The API server doesn't have any validation checks for that field to be present (unfortunately) but the UI and SDK do. That field always gets accessed here https://github.com/caraml-dev/merlin/blob/967788eb98a0120dd20bec692ab04bcea8f08fc9/api/batch/controller.go#L161 for all requests to trigger prediction jobs.

api/batch/controller.go
deadlycoconuts: Yep, they do, the `predictionJob.Name` comes from `jobName` here https://github.com/caraml-dev/merlin/blob/967788eb98a0120dd20bec692ab04bcea8f08fc9/api/service/prediction_job_service.go#L145 which itself is generated using a (supposedly) unique id as a suffix for each job run `strconv.FormatInt(p.clock.Now().UnixNano(), 10)[:13]`.

api/service/version_endpoint_service.go
deadlycoconuts: It's more like if `secrets==nil` then it'll reuse old secrets. If `secrets==[]` then no secrets will be configured.

api/cluster/secret.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/caraml-dev/merlin/log"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (c *controller) createSecret(ctx context.Context, secretName string, namespace string, secretMap map[string]string) (string, error) {
```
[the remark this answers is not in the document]
deadlycoconuts: Ahaha good point, yep I'll change that :D Thanks!

api/batch/manifest.go
deadlycoconuts: Thanks for suggesting this change! 😅 I've also renamed `CreateSecret -> CreateK8sSecret` and `DeleteSecret -> DeleteK8sSecret`.

api/batch/controller.go
shydefoo: ah ok cool thanks!

shydefoo (approved):
LGTM! Thanks for clarifying the questions too

api/batch/controller.go
shydefoo: got it!

deadlycoconuts:
Thanks for the quick review and for the suggestions! Gonna merge this now! 🚀

PR #631, opened 2025-02-18
feat(api): Add maxcompute sink to log prediction logs for Model Observability

vinoth-gojek:
# Description
The changes aims to enable MaxCompute as a sink for Model Observability prediction logs

# Modifications
1. Adds MaxCompute config as sink_types when publisher-observer for a model is deployed
2. Adds changes to create, alter table and write records for observability prediction logs

# Note to Reviewer
1. Work around - At the moment MaxCompute doesn't support adding partitions on Timestamp columns, so we create partition on day with trunc_time(timestamp, 'day')
2. PYODPS at the moment don't support the below functionality in the SDK, so we use the execute_sql
   a. Creating tables with auto partition on date
   b. Alter table to add new columns
3. There is one more blocking error from MaxCompute - df.persist doesn't insert rows with overwrite=False throwing an error "Partition doesn't exist"

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation


api/pkg/observability/deployment/deployment.go
```diff
@@ -282,6 +282,18 @@ func (c *deployer) createSecretSpec(data *models.WorkerData) (*corev1.Secret, er
...
+		Type: MaxCompute,
+		Config: MaxComputeSink{
+			Project:         c.consumerConfig.MaxComputeSink.Project,
+			Dataset:         c.consumerConfig.MaxComputeSink.Dataset,
+			TTLDays:         c.consumerConfig.MaxComputeSink.TTLDays,
+			AccessKeyID:     c.consumerConfig.MaxComputeSink.AccessKeyID,
+			AccessKeySecret: c.consumerConfig.MaxComputeSink.AccessKeySecret,
+			AccessUrl:       c.consumerConfig.MaxComputeSink.AccessUrl,
+		},
+	})
```
naufalandika, line 295: what about adding the `MaxComputeSink` in the `observationSinks` initiation
```
observationSinks := []ObservationSink{
	{
		Type: BQ,
		Config: BigQuerySink{
			Project: c.consumerConfig.BigQuerySink.Project,
			Dataset: c.consumerConfig.BigQuerySink.Dataset,
			TTLDays: c.consumerConfig.BigQuerySink.TTLDays,
		},
	},{
		Type: MaxCompute,
		Config: MaxComputeSink{
			Project:         c.consumerConfig.MaxComputeSink.Project,
			Dataset:         c.consumerConfig.MaxComputeSink.Dataset,
			TTLDays:         c.consumerConfig.MaxComputeSink.TTLDays,
			AccessKeyID:     c.consumerConfig.MaxComputeSink.AccessKeyID,
			AccessKeySecret: c.consumerConfig.MaxComputeSink.AccessKeySecret,
			AccessUrl:       c.consumerConfig.MaxComputeSink.AccessUrl,
		},
	}
}
```

vinoth-gojek: ohh good point, I am updating this, just that I might have to fix the test again

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -314,6 +320,236 @@ def write(self, dataframe: pd.DataFrame):
...
+        df = ODPSDataFrame(dataframe)
+        for i in range(0, self.retry.retry_attempts + 1):
+            try:
+                temp_table_id = f"{self.write_location}_{random.randint(10000, 99999)}"
+                df.persist(temp_table_id, create_table=True, lifecycle=1)
+                schema_fields = self.get_schema_fields(df.schema)
+                column_names_original_table = ','.join(schema_fields)
+                cast_datetime_field_to_timestamp = ["cast(" + PREDICTION_LOG_TIMESTAMP_COLUMN + " as timestamp)" if column_name == PREDICTION_LOG_TIMESTAMP_COLUMN else column_name for column_name in schema_fields]
+                column_values_from_temp_table = ','.join(cast_datetime_field_to_timestamp)
+                insert_into_query = "insert into {table_name} ( {cols} ) select {values} from {temp_table}".format(
```
vinoth-gojek, line 531: There is no sdk support at the moment to insert data into an auto partitioned table without overwrite. The workaround suggested is a. insert the data into temp table, b. insert the temp table into the desired table using insert into

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -314,6 +320,236 @@ def write(self, dataframe: pd.DataFrame):
...
+            if not schema_diff:
+                return original_table
+            alter_table_query = "alter table {table_name} add columns ( {cols} )".format(
+                table_name=self.table_name_with_dataset,
+                cols=self.get_column_values_for_query(schema_fields=schema_diff),
+            )
+            return self._client.execute_sql(alter_table_query)
+            
+        except NoSuchObject:
+            create_table_query = "create table {table_name} ( {cols} )  auto partitioned by (trunc_time({timestamp_column}, 'day') as request_timestamp_pt)".format(
```
vinoth-gojek, line 437: MaxCompute at the moment doesn't support creation of partition columns on timestamp. The auto partition is a workaround suggested by the MC team

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -314,6 +320,236 @@ def write(self, dataframe: pd.DataFrame):
...
+        Returns the MaxCompute table location with the dataset prefix
+        for instance, the returned table name will be {dataset}.{table_name}
+        :return:
+        """
+        table_name = f"prediction_log_{self._project}_{self._model_id}".replace("-", "_").replace(
+            ".", "_"
+        )
+        return f"{self.dataset}.{table_name}"
+    
+    def get_schema_fields(self, schema) -> List[str]:
```
deadlycoconuts, line 513: We can add an underscore to the name of this method too, since I don't think users are expected to call this directly.

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -314,6 +320,236 @@ def write(self, dataframe: pd.DataFrame):
...
+        for field in schema:
+            fields.append(field.name)
+
+        return fields
+
+
+    def write(self, dataframe: pd.DataFrame):
+        df = ODPSDataFrame(dataframe)
+        for i in range(0, self.retry.retry_attempts + 1):
+            try:
```
deadlycoconuts, line 524: Since this `try` block is pretty huge and doing multiple transactions with the MaxCompute, i.e. 1) create temp table 2) copy from temp table into original table 3) delete temp table, should we break this up into simpler blocks so it's easier to not only debug (with appropriate error messages returned  for each step) and also cleanup if necessary (e.g. should we clean up the temp table if the copying step failed, or if there are retries remaining, should we reuse the same temp table again?).

deadlycoconuts (changes requested):
Thanks a lot for this PR! Most of the comments are pretty minor refactoring changes. The most important one would probably be the one on the creation/copying from temp tables. Let me know what you think about that one :)

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -314,6 +320,236 @@ def write(self, dataframe: pd.DataFrame):
...
+    @property
+    def table_name_with_dataset(self) -> str:
+        """
+        Returns the MaxCompute table location with the dataset prefix
+        for instance, the returned table name will be {dataset}.{table_name}
+        :return:
+        """
+        table_name = f"prediction_log_{self._project}_{self._model_id}".replace("-", "_").replace(
+            ".", "_"
+        )
```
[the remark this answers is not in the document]
vinoth-gojek: There is a small difference, the table name generated is dataset_table-name, rather than project_dataset_table-name

python/observation-publisher/publisher/observation_sink.py
vinoth-gojek: I feel the function still small and in generally doing one action as whole, inserting records to the maxcompute table. Since the temp table is created with lifetime of 1 (minimum unit available) day, it will be cleared after a day if in case of any error happening.

python/observation-publisher/publisher/observation_sink.py
```diff
@@ -314,6 +320,236 @@ def write(self, dataframe: pd.DataFrame):
...
+    taking time to update the table schema / create new table.
+    Attributes:
+        enabled: Whether to retry failed write attempts
+        retry_attempts: Number of retry attempts
+        retry_interval_seconds: Interval between retry attempts
+    """
+
+    enabled: bool = False
+    retry_attempts: int = 4
+    retry_interval_seconds: int = 30
```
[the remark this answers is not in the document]
vinoth-gojek: Yeah, this is a valid point, can I track it as a separate task (probably post cutover), as the MLObs support for maxcompute is already heavily delayed :(

python/observation-publisher/publisher/observation_sink.py
deadlycoconuts: Okay sure thing it's not that much of a problem them if temp tables are cleared within a day.

python/observation-publisher/publisher/observation_sink.py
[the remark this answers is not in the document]
deadlycoconuts: Ah I see, oh well okay then let's keep both methods :)

python/observation-publisher/publisher/observation_sink.py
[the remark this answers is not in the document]
deadlycoconuts: Ah I see, sounds good, there's no rush to do this. Thanks!

deadlycoconuts (approved):

PR #633, opened 2025-03-05
feat(api): allow enable mlobs from API directly

naufalandika:
# Description
- deprecate `Model.ObservabilitySupported`
- use `VersionEndpoint.ModelObservability.Enabled` as SSOT for MLObs configuration

# Modifications
- remove the use of `Model.ObservabilitySupported`
- use `VersionEndpoint.EnableModelObservability` value as `VersionEndpoint.ModelObservability.Enabled` if `VersionEndpoint.ModelObservability` is not set
- remove unit test with `Model.ObservabilitySupported` logic
- adjust unit test mock call

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/api/api/validator.go
+++ b/api/api/validator.go
@@ -179,13 +179,11 @@ func modelObservabilityValidation(endpoint *models.VersionEndpoint, model *model
 		if !endpoint.IsModelMonitoringEnabled() {
 			return nil
 		}
+
 		if !slices.Contains(supportedObservabilityModelTypes, model.Type) {
 			return fmt.Errorf("%s: %w", model.Type, ErrUnsupportedObservabilityModelType)
 		}
 
-		if !model.ObservabilitySupported {
-			return fmt.Errorf("model observability is not supported for this model")
-		}
 		return nil
 	})
 }
--- a/api/api/version_endpoints_api.go
+++ b/api/api/version_endpoints_api.go
@@ -154,6 +154,7 @@ func (c *EndpointsController) CreateEndpoint(r *http.Request, vars map[string]st
 		}
 
 		newEndpoint.EnvironmentName = env.Name
+		newEndpoint.SetModelObservabilityEnabledWithEnableModelObservabilityIfNil()
 	}
 
 	validationRules := []requestValidator{
@@ -209,6 +210,8 @@ func (c *EndpointsController) UpdateEndpoint(r *http.Request, vars map[string]st
 	if !ok {
 		return BadRequest("Unable to parse body as version endpoint resource")
 	}
+
+	newEndpoint.SetModelObservabilityEnabledWithEnableModelObservabilityIfNil()
 
 	env, err := c.AppContext.EnvironmentService.GetEnvironment(newEndpoint.EnvironmentName)
 	if err != nil {
--- a/api/api/version_endpoints_api.go
+++ b/api/api/version_endpoints_api.go
@@ -154,7 +154,7 @@ func (c *EndpointsController) CreateEndpoint(r *http.Request, vars map[string]st
 		}
 
 		newEndpoint.EnvironmentName = env.Name
-		newEndpoint.SetModelObservabilityEnabledWithEnableModelObservabilityIfNil()
+		newEndpoint.SetModeObservabilityIfNil()
 	}
 
 	validationRules := []requestValidator{
@@ -211,7 +211,7 @@ func (c *EndpointsController) UpdateEndpoint(r *http.Request, vars map[string]st
 		return BadRequest("Unable to parse body as version endpoint resource")
 	}
 
-	newEndpoint.SetModelObservabilityEnabledWithEnableModelObservabilityIfNil()
+	newEndpoint.SetModeObservabilityIfNil()
 
 	env, err := c.AppContext.EnvironmentService.GetEnvironment(newEndpoint.EnvironmentName)
 	if err != nil {
--- a/api/models/model_endpoint.go
+++ b/api/models/model_endpoint.go
@@ -38,6 +38,10 @@ type ModelEndpoint struct {
 }
 
 func (me *ModelEndpoint) GetVersionEndpoint() *VersionEndpoint {
+	if me == nil {
+		return nil
+	}
+
 	if me.Rule == nil || len(me.Rule.Destination) == 0 {
 		return nil
 	}
--- a/api/models/model_observability.go
+++ b/api/models/model_observability.go
@@ -16,6 +16,14 @@ type ModelObservability struct {
 	PredictionLogIngestionResourceRequest *WorkerResourceRequest `json:"prediction_log_ingestion_resource_request"`
 }
 
+func (mo *ModelObservability) IsEnabled() bool {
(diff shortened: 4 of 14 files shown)


api/api/version_endpoints_api.go
```diff
@@ -210,6 +211,8 @@ func (c *EndpointsController) UpdateEndpoint(r *http.Request, vars map[string]st
 		return BadRequest("Unable to parse body as version endpoint resource")
 	}
 
+	newEndpoint.SetModelObservabilityEnabledWithEnableModelObservabilityIfNil()
```
vinoth-gojek, line 214: Suggest to rename this to "SetModeObservabilityIfNil" and add comments with historical context of why we do this inside the "SetModeObservabilityIfNil" function

api/models/model_endpoint.go
```diff
@@ -38,6 +38,10 @@ type ModelEndpoint struct {
 }
 
 func (me *ModelEndpoint) GetVersionEndpoint() *VersionEndpoint {
+	if me == nil {
```
vinoth-gojek, line 41: curious, why is this being added now? wondering how the logic works without this check prior!

api/models/version_endpoint.go
```diff
@@ -134,10 +134,11 @@ func (ve *VersionEndpoint) IsServing() bool {
 }
 
 func (ve *VersionEndpoint) IsModelMonitoringEnabled() bool {
-	if ve.ModelObservability == nil {
-		return ve.EnableModelObservability
+	if ve == nil {
```
vinoth-gojek, line 137: Similar question again, is this check added as an impact created by the MR changes?! This case of ve == nil is unexpected right?

api/pkg/observability/event/event.go
```diff
@@ -34,10 +34,6 @@ func NewEventProducer(jobProducer queue.Producer, observabilityPublisherStorage
 }
 
 func (e *eventProducer) ModelEndpointChangeEvent(modelEndpoint *models.ModelEndpoint, model *models.Model) error {
-	if !model.ObservabilitySupported {
-		return nil
```
vinoth-gojek, line 38: I feel below check of !ve.IsModeMonitoringEnabled() would need to be in place of !model.ObservabilitySupported

If not, wouldn't the undeploy action get triggered?! I am not sure if the unDeployAction checks if ve.IsModeMonitoringEnabled() is true

vinoth-gojek (commented):
Thanks @naufalandika for the changes to clean up ModelObservability flags, merging 3 different flags to one, thus enabling users to onboard themselves to MLObs without any manual db changes

I have some questions and suggested added, overall it looks good to me!

api/pkg/observability/event/event.go
vinoth-gojek: ohh I might be wrong here, moving the ve.IsModeMonitoringEnabled() would actually stop the undeployment from happening.

I notice a different problematic scenario here, will this then trigger undeploy action for all model versions endpoints irrespective of whether it has (observability-publisher) consumer running or not?

naufalandika: > ohh I might be wrong here, moving the ve.IsModeMonitoringEnabled() would actually stop the undeployment from happening.

correct



> I notice a different problematic scenario here, will this then trigger undeploy action for all model versions endpoints irrespective of whether it has (observability-publisher) consumer running or not?

it should only undeploy if the publisher is not nil, we check it [here](https://github.com/caraml-dev/merlin/pull/633/files/f4d592d4f6fc5e98a5e1a4eb958bc0e70dc762f9#diff-b6404490aff014183caa61423d8e8cd30a2ca764067652f7a64be5620b8678e5L51-L53)

vinoth-gojek: ohh thats awesome!

api/api/version_endpoints_api.go
naufalandika: Updated!

api/models/version_endpoint.go
naufalandika: the reason is related to the review above. and because the `modelEndpoint` [here](https://github.com/caraml-dev/merlin/pull/633/files#diff-b6404490aff014183caa61423d8e8cd30a2ca764067652f7a64be5620b8678e5R67) might be nil, then the `versionEndpoint` might also be nil

api/api/validator.go
```diff
@@ -179,13 +179,11 @@ func modelObservabilityValidation(endpoint *models.VersionEndpoint, model *model
 		if !endpoint.IsModelMonitoringEnabled() {
 			return nil
 		}
+
```
vinoth-gojek, line 182: nit: remove the new line

api/models/model_endpoint.go
```diff
@@ -38,6 +38,10 @@ type ModelEndpoint struct {
...
+		return nil
+	}
+
```
vinoth-gojek, line 44: nit: remove the new line

api/models/model_observability.go
```diff
@@ -16,6 +16,14 @@ type ModelObservability struct {
 	PredictionLogIngestionResourceRequest *WorkerResourceRequest `json:"prediction_log_ingestion_resource_request"`
 }
 
+func (mo *ModelObservability) IsEnabled() bool {
+	if mo == nil {
+		return false
+	}
+
+	return mo.Enabled
```
vinoth-gojek, line 24: nit: remove the new line above

api/models/version_endpoint.go
```diff
@@ -134,10 +134,11 @@ func (ve *VersionEndpoint) IsServing() bool {
...
+		return false
 	}
-	return ve.ModelObservability.Enabled
+
```
vinoth-gojek, line 140: nit: remove the new line

api/models/version_endpoint.go
```diff
@@ -184,6 +185,25 @@ func (ve *VersionEndpoint) ParsedURL() (*url.URL, error) {
...
+// [TODO]: deprecate this after deprecating VersionEndpoint.EnableModelObservability
+// perviously we only have VersionEndpoint.EnableModelObservability and now we want to deprecate it
+// and only read/write to VersionEndpoint.ModelObservability instead. to allow backward compatibility if the user
+// only set VersionEndpoint.EnableModelObservability but not VersionEndpoint.ModelObservability we will use the
+// VersionEndpoint.EnableModelObservability value as VersionEndpoint.ModelObservability.EnableModelObservability
+func (ve *VersionEndpoint) SetModeObservabilityIfNil() {
+	if ve == nil {
+		return
+	}
+
```
vinoth-gojek, line 197: I feel removing the new lines in this function would be better

vinoth-gojek (approved):
hi @naufalandika overall the changes lgtm!

api/models/version_endpoint.go
naufalandika: updated!

api/models/version_endpoint.go
naufalandika: updated!

PR #634, opened 2025-03-13
draft: Cicd/sdk versioning improvement

andreas-aji:
# Description
- Background: we want to use latest merlin-sdk on batch-predictor and pyfunc-server. Previously we cannot do that because these packages relies on merlin-sdk versions that have been uploaded to PyPI. The idea is making the development and pipeline smoother by enabling these packages to install merlin-sdk from local source if the version is not yet available on pypi.

# Modifications
- Added `make update-sdk-version VERSION=1.2.3` at `python/` directory to update all merlin-sdk version to all dependent packages.
- Added conditional handling on `make setup` command in `batch-predictor` and `pyfunc-server` package. When installing `merlin-sdk`, if not found on pypi then build merlin-sdk locally and then install.

# Tests
- Test github actions pipelines if it's working correctly.

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -5,14 +5,10 @@ observation-publisher:
 	@echo "Building image for observation publisher..."
 	@docker build -t ${OBSERVATION_PUBLISHER_IMAGE_TAG} -f observation-publisher/Dockerfile --progress plain .
 
-SDK_VER = $(subst v,,$(SDK_VERSION))
+VER = $(subst v,,$(VERSION))
 
-.PHONY: upgrade-sdk-version
-upgrade-sdk-version:
-	python -m pip install --upgrade pip
-	pip install setuptools setuptools_scm twine wheel
-	sed -i -e "s|merlin-sdk.*|merlin-sdk==$(SDK_VER)|g" ./batch-predictor/requirements.txt
-	sed -i -e "s|merlin-sdk.*|merlin-sdk==$(SDK_VER)|g" ./pyfunc-server/requirements.txt
-	cd sdk && sed -i -e "s|VERSION = \".*\"|VERSION = \"$(SDK_VER)\"|g" ./merlin/version.py
-	cd sdk && python setup.py sdist bdist_wheel
-	pip install sdk/dist/merlin_sdk-$(SDK_VER)-py3-none-any.whl
+# to update merlin-sdk version in packages that depend on it
+.PHONY: update-sdk-version
+update-sdk-version:
+	sed -i -e "s|merlin-sdk.*|merlin-sdk==$(VER)|g" ./batch-predictor/requirements.txt
+	sed -i -e "s|merlin-sdk.*|merlin-sdk==$(VER)|g" ./pyfunc-server/requirements.txt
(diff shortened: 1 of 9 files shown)


python/batch-predictor/requirements.txt
```diff
@@ -1,6 +1,6 @@
 cloudpickle==2.0.0
 findspark==2.0.1
-merlin-sdk==0.46.0
+merlin-sdk==0.0.0
```
shydefoo, line 3: noob q: is there any special meaning by setting this to 0.0.0? Or this is just a placeholder?

python/Makefile
```diff
@@ -4,3 +4,11 @@ OBSERVATION_PUBLISHER_IMAGE_TAG ?= observation-publisher:dev
 observation-publisher:
 	@echo "Building image for observation publisher..."
 	@docker build -t ${OBSERVATION_PUBLISHER_IMAGE_TAG} -f observation-publisher/Dockerfile --progress plain .
+
+VER = $(subst v,,$(VERSION))
+
+# to update merlin-sdk version in packages that depend on it
+.PHONY: update-sdk-version
+update-sdk-version:
```
shydefoo, line 12: Might be useful to add this in the readme in case we forget to set this when developing

shydefoo (commented):
Thanks for the changes! Will leave it to ziyi/bthari to approve in case there are some stuff i may have missed out.

python/Makefile
andreas-aji: With the latest adjustment, this function is not necessary anymore. For pipeline it will be automatically updated with tag as the version. And for local dev, `make setup` will build and install the latest merlin-sdk code but with version `0.0.0` (the version will just be a place holder).

I will delete this function.

python/batch-predictor/requirements.txt
andreas-aji: This is just a place holder now. Running `make setup` will always build and install merlin-sdk so it will always be the latest version.

But on pipeline it will be updated to the tag. Do you think we need to add some comment here stating that this is a placeholder?

python/Makefile
andreas-aji: Removed this function.

python/batch-predictor/requirements.txt
andreas-aji: Turns out inline comments are not supported by pipenv, will not add comment for now.

python/batch-predictor/Makefile
```diff
@@ -34,3 +36,14 @@ proto:
...
 	@$(PROTOC_PATH) -I=. --python_out=merlinpyspark --mypy_out=merlinpyspark --go_out=pkg/ --go-json_out=pkg/ --go_opt=paths=source_relative spec/*.proto
+
+
+.PHONY: build_install_sdk
+
+SDK_VER := $(shell grep '^merlin-sdk' requirements.txt | awk -F '[=<>]+' '{print $$2}')
+
+build_install_sdk:
+	cd ../sdk && pip install setuptools setuptools_scm twine wheel
+	cd ../sdk && sed -i -e "s|VERSION = \".*\"|VERSION = \"${SDK_VER}\"|g" ./merlin/version.py
```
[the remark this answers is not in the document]
andreas-aji: Oh right, we don't need this anymore since for local dev we always use version 0.0.0. I will simply this.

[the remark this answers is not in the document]
andreas-aji: Updated!

PR #635, opened 2025-03-14
disallow undeploy and redeploy of version endpoint in pending deployment 

vinoth-gojek:
# Description
Right now, the model version endpoints in pending state is not allowed to be redeployed/undeployed from UI. But the same is possible via sdk
The change disallow the undeploy and redeploy of version endpoint in pending deployment state

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

Tested by using an endpoint in pending and the undeploy and redeploy were blocked as expected

➜  ✗ curl localhost:8080/v1/models/411/versions/2/endpoint/4308d895-541d-40e7-b0f8-4e0d7c3ac669 -X PUT   
{"error":"Version Endpoint 4308d895-541d-40e7-b0f8-4e0d7c3ac669 is still pending and cannot be redeployed"}
➜  ✗ curl localhost:8080/v1/models/411/versions/2/endpoint/4308d895-541d-40e7-b0f8-4e0d7c3ac669 -X DELETE
{"error":"Version Endpoint 4308d895-541d-40e7-b0f8-4e0d7c3ac669 is still pending and cannot be undeployed"}
➜

### DIFF
--- a/api/api/version_endpoints_api.go
+++ b/api/api/version_endpoints_api.go
@@ -205,10 +205,6 @@ func (c *EndpointsController) UpdateEndpoint(r *http.Request, vars map[string]st
 		return InternalServerError(fmt.Sprintf("Error getting endpoint: %v", err))
 	}
 
-	if endpoint.IsPending() {
-		return BadRequest(fmt.Sprintf("Version Endpoint %s is still pending and cannot be redeployed", endpointID))
-	}
-
 	newEndpoint, ok := body.(*models.VersionEndpoint)
 	if !ok {
 		return BadRequest("Unable to parse body as version endpoint resource")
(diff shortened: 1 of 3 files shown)


deadlycoconuts (approved):
LGTM! But can we just add a unit test for this if it's not too complicated? 😅

vinoth-gojek:
hi @deadlycoconuts yes, I have added tests.
Also realised the UpdateEndpoint is already handling the scenario of "disallowing pending deployments from redeployments". So modified to add the check only for delete endpoint

deadlycoconuts:
Got it, thanks for adding the unit tests! Feel free to merge this anytime!

PR #636, opened 2025-03-20
feat: add extra config option for kafka

opan:
# Description
We have a usecase where we need to connect to Kafka V3 and we need to configure the SSL properties. Adding an option so that user can provide an extra option during kafka producer initialization as a kafka `.properties` files. This option is expected to be a path to the config file. So we have to mount it before hand in the deployment. Either as a secret or configmap

# Modifications
Add options to CLI inference-logging API

# Tests

# Checklist
- [ ] Added PR label
- [ ] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/api/cmd/inference-logger/main.go
+++ b/api/cmd/inference-logger/main.go
@@ -33,14 +33,14 @@ import (
 )
 
 var (
-	logUrl                = flag.String("log-url", "localhost:8002", "The URL to send request/response logs to")
-	additionalKafkaConfig = flag.String("kafka-config-extra", "", "Additional Kafka configuration. This should be path to a file containing the configuration")
-	port                  = flag.String("port", "9081", "Logger port")
-	componentPort         = flag.String("component-port", "8080", "Component port")
-	workers               = flag.Int("workers", 5, "Number of workers")
-	logMode               = flag.String("log-mode", string(merlinlogger.LogModeAll), "Whether to log 'request', 'response' or 'all'")
-	inferenceService      = flag.String("inference-service", "my-model-1", "The InferenceService name to add as header to log events")
-	namespace             = flag.String("namespace", "my-project", "The namespace to add as header to log events")
+	logUrl           = flag.String("log-url", "localhost:8002", "The URL to send request/response logs to")
+	kafkaConfigPath  = flag.String("kafka-config-path", "", "Additional Kafka configuration. This should be path to a file containing the configuration")
+	port             = flag.String("port", "9081", "Logger port")
+	componentPort    = flag.String("component-port", "8080", "Component port")
+	workers          = flag.Int("workers", 5, "Number of workers")
+	logMode          = flag.String("log-mode", string(merlinlogger.LogModeAll), "Whether to log 'request', 'response' or 'all'")
+	inferenceService = flag.String("inference-service", "my-model-1", "The InferenceService name to add as header to log events")
+	namespace        = flag.String("namespace", "my-project", "The namespace to add as header to log events")
 
 	// These flags are not needed by our logger but provided by Kserve, hence we need to parse it to avoid error.
 	sourceUri     = flag.String("source-uri", "", "The source URI to use when publishing cloudevents")
@@ -346,8 +346,8 @@ func getLogSink(
 			"compression.type":  merlinlogger.CompressionType,
 		}
 
-		if additionalKafkaConfig != nil && *additionalKafkaConfig != "" {
-			err := addKafkaConfig(kafkaCfg, *additionalKafkaConfig)
+		if kafkaConfigPath != nil && *kafkaConfigPath != "" {
+			err := addKafkaConfig(kafkaCfg, *kafkaConfigPath)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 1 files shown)


api/cmd/inference-logger/main.go
```diff
@@ -338,14 +340,18 @@ func getLogSink(
...
 		// Initialize the producer
 		var kafkaProducer merlinlogger.KafkaProducer
+		kafkaCfg := &kafka.ConfigMap{
+			"bootstrap.servers": url,
+			"message.max.bytes": merlinlogger.MaxMessageBytes,
+			"compression.type":  merlinlogger.CompressionType,
+		}
+
+		if additionalKafkaConfig != nil && *additionalKafkaConfig != "" {
+			addKafkaConfig(kafkaCfg, *additionalKafkaConfig)
```
bthari, line 350: 😮 is it expected that we don't catch the error returned by this function?

api/cmd/inference-logger/main.go
```diff
@@ -389,3 +399,32 @@ func getLogSink(
...
+	file, err := os.Open(kafkaConfig)
+	if err != nil {
+		return &kafka.ConfigMap{}, fmt.Errorf("failed to open kafka config file: %w", err)
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.TrimSpace(line) == "" || strings.HasPrefix(line, "#") {
```
bthari, line 413: what is this validation trying to handle? why it's expecting (and only handle) the special character #? 🧐

api/cmd/inference-logger/main.go
```diff
@@ -389,3 +399,32 @@ func getLogSink(
 		return merlinlogger.NewConsoleSink(log), nil
 	}
 }
+
+func addKafkaConfig(cfg *kafka.ConfigMap, kafkaConfig string) (*kafka.ConfigMap, error) {
```
bthari, line 403: nit: I don't think we need to return the `*kafka.ConfigMap`, as it's not used by the caller anyway

api/cmd/inference-logger/main.go
```diff
@@ -32,13 +33,14 @@ import (
...
 var (
-	logUrl           = flag.String("log-url", "localhost:8002", "The URL to send request/response logs to")
-	port             = flag.String("port", "9081", "Logger port")
-	componentPort    = flag.String("component-port", "8080", "Component port")
-	workers          = flag.Int("workers", 5, "Number of workers")
-	logMode          = flag.String("log-mode", string(merlinlogger.LogModeAll), "Whether to log 'request', 'response' or 'all'")
-	inferenceService = flag.String("inference-service", "my-model-1", "The InferenceService name to add as header to log events")
-	namespace        = flag.String("namespace", "my-project", "The namespace to add as header to log events")
+	logUrl                = flag.String("log-url", "localhost:8002", "The URL to send request/response logs to")
+	additionalKafkaConfig = flag.String("kafka-config-extra", "", "Additional Kafka configuration. This should be path to a file containing the configuration")
```
bthari, line 37: super nit: can we change the name to indicate it's a file path, `kafka-config-path/file`?

bthari:
Thank you for enabling this capability! I left a few minor comments, the rest should be good :) Just have another small question, how do you intend to mount the `.properties` file into the deployment?

opan:
@bthari Currenly I'm thinking to just attach the file as configmap or secret into merlin deployment, something like this:
```
      ...
      
        - mountPath: /etc/kafka
          name: kafka-config
          ...
      volumes:
      - name: kafka-config
        secret:
          defaultMode: 420
          secretName: kafka-config
```

But I'm open to any suggestion.

api/cmd/inference-logger/main.go
opan: oh yes, we can add error checking there.

api/cmd/inference-logger/main.go
opan: sure, we can remove it.

api/cmd/inference-logger/main.go
opan: `#` is a commented line inside the properties files. Something like [this](https://www.ibm.com/docs/en/oala/1.3.7?topic=collection-apache-kafka-broker-configuration-example)

and first condition is the empty line between the config, something people wants to make separation between config group.

api/cmd/inference-logger/main.go
opan: make sense, how about `kafkaConfigFilePath`?

bthari:
@opan when you say, `merlin deployment` is it this Merlin deployment or the model deployment? Because, correct me if I’m wrong, this logger will run as a second container in the model (kserve inferenceservice) pod, so the config needs to be mounted there, not in the Merlin deployment itself

@deadlycoconuts may have other informations on this

deadlycoconuts:
@opan yeah I think @bthari is right here - the Merlin logger (where you added the additional config to) is deployed in each Merlin model/transformer pod by KServe and that logger container is not the same as the Merlin API server deployment (or the Merlin deployment in short). We'll need to make KServe mount the config file onto each Merlin model/transformer pod for that to work. Let me dig around to provide you with some links to get this to work.

opan:
@bthari I see, so i missunderstood the setup. Can we try to mount the config file via some hook or some tool like kyverno?

opan:
@bthari I think we can close this PR. we found a way to connect to kafka cluster without configuring the SSL. Thank you for the helps!

PR #638, opened 2025-04-10
Modify maxcompute data type for float

vinoth-gojek:
# Description
The MR modifies datatype of prediction log table from float to Double and enable print logs by adding env variable

# Modifications
1. modifies datatype of prediction log table from float to Double

The incoming prediction logs are stored in a temp table and the temp table is inserted to the prediction log table
The temp table is created using df.persist() which only supports double and bigint (and not float and int)
So it is necessary for the prediction log table to be using double type to maintain the compatibility and avoid casting

2. Adds Environment variable to print logs

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

### DIFF
--- a/python/observation-publisher/publisher/observation_sink.py
+++ b/python/observation-publisher/publisher/observation_sink.py
@@ -542,6 +542,7 @@ class MaxComputeSink(ObservationSink):
                         f"Errors when inserting rows to MaxCompute, retrying attempt {i}/{self.retry.retry_attempts}"
                     )
                     time.sleep(self.retry.retry_interval_seconds)
+        print(f"Failed to write to MaxCompute after {self.retry.retry_attempts} attempts")
 
 
 def new_observation_sink(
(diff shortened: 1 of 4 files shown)


python/observation-publisher/publisher/observation_sink.py
```diff
@@ -542,8 +542,6 @@ def write(self, dataframe: pd.DataFrame):
                         f"Errors when inserting rows to MaxCompute, retrying attempt {i}/{self.retry.retry_attempts}"
                     )
                     time.sleep(self.retry.retry_interval_seconds)
-            
-        print(f"Failed to write to MaxCompute after {self.retry.retry_attempts} attempts")
```
vinoth-gojek, line 546: Removing an incorrect print statement

deadlycoconuts (approved):
Thanks for the changes and for investigating this bug!

PR #640, opened 2025-04-22
fix(logger): Improve error handling to return errors when creating Kafka topics

deadlycoconuts:
# Description
It appears that calling `CreateTopics` [here](https://github.com/caraml-dev/merlin/blob/eb52b7f240e3cc516026816e15b3b248a1aa3850/api/pkg/inference-logger/logger/kafka_sink.go#L59) in `NewKafkaSink` sometimes doesn't return errors within the `err` return value, but instead in each of the individual `topicResults` (this is a slice) elements returned. 

This happens in certain scenarios, e.g. when there are authorisation problems creating new topics, which causes errors to be returned within each element of the `topicResults` slice instead (leaving `err` as `nil` instead):

Screenshot 2025-04-22 at 15 19 52

This is problematic because it means that even though there were problems creating the topic, a `nil` error gets returned, with a `nil` sink value.

Tracing this back to where `NewKafkaSink` gets called in `getLogSink`, https://github.com/caraml-dev/merlin/blob/main/api/cmd/inference-logger/main.go#L362, and to where `getLogSink` gets called, https://github.com/caraml-dev/merlin/blob/main/api/cmd/inference-logger/main.go#L130, it's easy to see that the error isn't properly caught (since the error returned is `nil`) at any part of the call stack. 

To make things worse, because `logSink` is `nil`, it continues getting passed around the entire logger code as if it were non-nil, creating huge problems downstream, like null pointer exceptions (and crashes) because `logSink` is not at all expected to be `nil`.

This PR simply refactors the error handling of the Kafka sink initialisation step, so that any errors are properly returned and subsequently caught by the upstream callers.

# Modifications
- `api/pkg/inference-logger/logger/kafka_sink.go` - Improve error handling

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
None
```

### DIFF
--- a/api/pkg/inference-logger/logger/kafka_sink.go
+++ b/api/pkg/inference-logger/logger/kafka_sink.go
@@ -63,9 +63,12 @@ func NewKafkaSink(
 			ReplicationFactor: ReplicationFactor,
 		},
 	})
+	if err != nil {
+		return nil, err
+	}
 	for _, result := range topicResults {
 		if result.Error.Code() != kafka.ErrNoError && result.Error.Code() != kafka.ErrTopicAlreadyExists {
-			return nil, err
+			return nil, fmt.Errorf("failed to create topic %s: %w", result.Topic, result.Error)
 		}
 	}
 


vinoth-gojek (approved):

andreas-aji (approved):

deadlycoconuts:
Thanks for the quick reviews everyone! 🚀 Merging this now!

PR #642, opened 2025-04-24
feat(ui): update templating of monitoring url

anantadwi13:
# Description
In this PR, we are moving the templating of monitoring url from classic string concat to [mustache template](https://github.com/janl/mustache.js). This change allows user to configure the monitoring url dynamically without being constrained by our pre-defined url format

# Modifications
- add new mustache dependency
- update `createMonitoringUrl` logic to use mustache template
- remove duplicated code of `createMonitoringUrl`

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
There will be a breaking change for the next released version. Users need to update REACT_APP_MONITORING_DASHBOARD_JOB_BASE_URL env var to use the latest format.
```

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -15,6 +15,7 @@
     "dagre-d3-react": "^0.2.4",
     "js-yaml": "^4.1.0",
     "moment": "^2.30.1",
+    "mustache": "^4.2.0",
     "object-assign-deep": "^0.4.0",
     "proper-url-join": "^2.1.1",
     "react": "^18.3.1",
--- a/ui/src/pages/job/JobPage.js
+++ b/ui/src/pages/job/JobPage.js
@@ -100,6 +100,7 @@ const JobPage = () => {
             <EuiFlexItem>
               <JobDetailTabNavigation
                 project={project}
+                model={model}
                 job={job}
                 selectedTab={section}
               />
--- a/ui/src/pages/job/components/JobActions.js
+++ b/ui/src/pages/job/components/JobActions.js
@@ -16,7 +16,7 @@ function isActiveJob(status) {
   return ["pending", "running"].includes(status);
 }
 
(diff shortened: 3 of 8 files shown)


deadlycoconuts (approved):
Thanks for this change! I didn't know the batch prediction job monitoring urls were generated at the UI level (the online model monitoring urls are templated within the API server/backend).

I have just one tiny additional request: it seems like this `createMonitoringUrl` function defined in another file here https://github.com/caraml-dev/merlin/blob/be0bdf9bc6274661b58a8ebc88dcaf215371e9e9/ui/src/pages/job/components/JobsTable.js#L86
which is terrible since both functions are exactly the same. Could you help do some tiny cleanup of the places where that duplicate is being used, and instead make those places use this utils method instead? 🙏🏼 Thanks!

anantadwi13:
> Thanks for this change! I didn't know the batch prediction job monitoring urls were generated at the UI level (the online model monitoring urls are templated within the API server/backend).
> 
> I have just one tiny additional request: it seems like this `createMonitoringUrl` function defined in another file here
> 
> https://github.com/caraml-dev/merlin/blob/be0bdf9bc6274661b58a8ebc88dcaf215371e9e9/ui/src/pages/job/components/JobsTable.js#L86
> 
> 
> which is terrible since both functions are exactly the same. Could you help do some tiny cleanup of the places where that duplicate is being used, and instead make those places use this utils method instead? 🙏🏼 Thanks!

Hi @deadlycoconuts, I have just made some patches. Could you please review it? Thanks 🙏

anantadwi13:
closed this pr due to a problem with github workflow when using forked-repository.  github will not allow forked-repository workflows to access main repository secrets

PR #648, opened 2025-05-05
feat(api): Add alert path prefix as configuration field

deadlycoconuts:
# Description
This PR simply introduces a new configuration field for the Merlin API server to make the alert path prefix (currently hardcoded as `alerts/merlin`) into a configurable field. This allows for more customisable paths to be used when setting up alerts using the current GitOps approach.

# Modifications
- `api/config/config.go` - Added a new configuration field `AlertPathPrefix`
- `api/service/model_endpoint_alert_service.go` - Made changes to the current way an alert file path is generated

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
NONE
```

### DIFF
--- a/api/config/testdata/base-configs-1.yaml
+++ b/api/config/testdata/base-configs-1.yaml
@@ -115,8 +115,9 @@ FeatureToggleConfig:
       BaseURL: https://test.io/
       DashboardRepository: dashboards/merlin
       DashboardBranch: master
-      AlertRepository: alerts/merlin
+      AlertRepository: alerts_repository/merlin
       AlertBranch: master
+      AlertPathPrefix: alerts/merlin
     WardenConfig:
       APIHost: https://test.io/
   ModelDeletionConfig:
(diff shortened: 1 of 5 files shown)


api/config/testdata/base-configs-1.yaml
```diff
@@ -115,8 +115,9 @@ FeatureToggleConfig:
       BaseURL: https://test.io/
       DashboardRepository: dashboards/merlin
       DashboardBranch: master
-      AlertRepository: alerts/merlin
+      AlertRepository: alerts_repository/merlin
```
deadlycoconuts, line 118: This change isn't technically needed but I did this to make the field (and its purpose) more easily distinguishable from the new `AlertPathPrefix` field.

bthari (approved):
Thank you for adding the config! LGTM 🚀

PR #652, opened 2025-06-18
feat(ui): add mustache templating in pod log url

anantadwi13:
# Description
There are so many cloud providers out there. This PR is used to add templating for pod log urls. Instead of relying only on Stackdriver logs (which is a Google product), we give our users the ability to create their own log urls. There are some variables that can be used by our users.  
#### Image Builder Log
1. Available Variables
- `cluster_name` (string)
- `namespace_name` (string)
- `job_name` (string)
- `start_time` (string)
- `end_time` (string)
2. Usage
```
# merlin config.yaml

FeatureToggleConfig:
  LogConfig:
    LogImageBuilderURL: https://logviewer.sample.local/logs/viewer?cluster={{cluster_name}}&namespace={{namespace_name}}&job={{job_name}}

# it generates
# https://logviewer.sample.local/logs/viewer?cluster=caraml-cluster&namespace=caraml-namespace&job=job-caraml
```
#### Model Log
1. Available Variables
- `cluster_name` (string)
- `namespace_name` (string)
- `pod_names` (array of {`value`, `is_first`})
- `start_time` (string)
2. Usage
```
# merlin config.yaml

FeatureToggleConfig:
  LogConfig:
    LogModelURL: https://logviewer.sample.local/logs/viewer?cluster={{cluster_name}}&namespace={{namespace_name}}&pods={{#pod_names}}{{#is_first}}{{value}}{{/is_first}}{{^is_first}},{{value}}{{/is_first}}{{/pod_names}}

# it generates
# https://logviewer.sample.local/logs/viewer?cluster=caraml-cluster&namespace=caraml-namespace&pods=pod-1,pod-2,pod-3
```

# Modifications
## BE  
- add `LogImageBuilderURL` and `LogModelURL`  
## FE  
- add mustache templating
- change Stackdriver urls to custom log url with backward compatibility

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note

```

### DIFF
--- a/ui/src/components/logs/ContainerLogsView.js
+++ b/ui/src/components/logs/ContainerLogsView.js
@@ -18,6 +18,7 @@ import { useMerlinApi } from "../../hooks/useMerlinApi";
 import mocks from "../../mocks";
 import { LogsSearchBar } from "./LogsSearchBar";
 import {createLogImageBuilderUrl, createLogModelUrl} from "../../utils/createLogUrl";
+import {createStackdriverUrl} from "../../utils/createStackdriverUrl";
 
 const componentOrder = [
   "image_builder",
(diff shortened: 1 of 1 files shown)


anantadwi13:
i'm going to update this pr to add backward compatibility to the stackdriver url

ui/src/utils/createLogUrl.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const createLogModelUrl = (template, cluster, namespace, podNames, startTime) => {
+    const data = {
+        cluster_name: cluster,
+        namespace_name: namespace,
+        pod_names: podNames.map((val, idx) => (
+            {
+                value: val,
+                is_first: idx === 0,
```
deadlycoconuts, line 24: Oh wait why do we need this? In ElasticSearch, we need to give an arbitrary order to pods? 😱

deadlycoconuts (approved):
Thanks a lot for this PR! Everything looks good 🚀 I just left one comment but that's all 😅

ui/src/utils/createLogUrl.js
anantadwi13: the reason we need this is to help us create a separator between each `pod_names` value. previously we can do this by using the javascript `.join()` function.
https://github.com/caraml-dev/merlin/blob/d36645468f9a481b5c5dd2489fa2adb9ca562c2f/ui/src/components/logs/ContainerLogsView.js#L175
but since we are changing the approach to mustache templating, we need to have another workaround. what we can do is having some logic in mustache templating like this
```text
{{#pod_names}}{{#is_first}}{{value}}{{/is_first}}{{^is_first}},{{value}}{{/is_first}}{{/pod_names}}
```
if the `pod_names` values are
```json
[
    {"value": "pod-1", "is_first": true},
    {"value": "pod-2", "is_first": false},
    {"value": "pod-3", "is_first": false}
]
```
then it will generate
```text
pod-1,pod-2,pod-3
```

anantadwi13: i think we might have another workaround by getting the common prefix for all `pod_names` value. then we can use the mustache template like this
```text
{{pod_name}}*
```
if the `pod_names` values are same like above, then it will return
```text
pod-*
```

anantadwi13: but, let's have this for the next improvement 😅

PR #653, opened 2025-06-30
chore: retrieve google sa in batch job only if it's specified

bthari:
# Description
Previously, the batch-job in Merlin was using BQ table and when client tries to submit a batch-job, Merlin assumed that a Google Service Account will be provided in the job configuration. But as we have added option for using MaxCompute and it doesn't need a Google SA, I'm adding a check where it will only try to retrieve the service account if it's provided

# Modifications

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
remove google sa dependency in batch job
```


deadlycoconuts (approved):
Thanks a lot for the quick fix! LGTM! 🚀

PR #655, opened 2025-08-25
feat: rollback mechanism virtualservice patch in model endpoint

bthari:
# Description
When creating/patching/deleting VirtualService for model endpoint related action, if there's any error happened after the action is successfully run, there's a possibility of mismatch state between the resource state in Kubernetes vs what is being recorded in database (as this will not be updated). 

# Modifications
Changes: 
- Add `GetVirtualService` function to get the current state of VirtualService
- Add `cleanVirtualServiceFields` function to remove not-needed field when creating or patching resource, e.g. UUID or generation number, if this isn't set to empty/default, the Patch/Create will not succeed
- Flow, if there's any error occur after the create/patching/delete happened, rollback the changes in Kubernetes to previous state 
  - Create -> remove the newly created VirtualService
  - Patch -> re-patch the VirtualService to previous state
  - Delete -> recreate the VirtualService if previously there's an existing one

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes
```release-note
NONE
```


api/istio/istio_test.go
```diff
@@ -76,6 +77,83 @@ var (
 	}
 )
 
+func Test_client_GetVirtualService(t *testing.T) {
```
deadlycoconuts, line 80: Wow thanks for coming up with unit tests even for this 😅

api/service/model_endpoint_service.go
```diff
@@ -124,6 +124,16 @@ func (s *modelEndpointsService) DeployEndpoint(ctx context.Context, model *model
...
 
+	// if there's error during VirtualService creation, we will rollback the changes, to make sure
+	// there's no hanging state of vs and to align with the database state
+	defer func() {
+		if err != nil {
+			if err := istioClient.DeleteVirtualService(context.WithoutCancel(ctx), model.Project.Name, model.Name); err != nil {
+				log.Errorf("failed to rollback create VirtualService: %v", err)
+			}
+		}
+	}()
```
deadlycoconuts, line 135: Ahhh so if I understand correctly, we're deleting this virtual service if any of the steps following (and including) the virtual service creation fails? i.e. if any of the following functions return an error,
- `istioClient.CreateVirtualService`
- `s.modelEndpointStorage.Save`
- `s.observabilityEventProducer.ModelEndpointChangeEvent`

we delete the virtual service created? Wow this might actually be the cause of the bug 🤔 Like improper virtual service clean up because the whole set of operations aren't atomic (we also aren't reversing the other steps that have been executed like the DB update when subsequent functions fail). Thanks for digging through the code and fixing the workflow!

api/service/model_endpoint_service.go
```diff
@@ -342,6 +385,19 @@ func (c *modelEndpointsService) assignVersionEndpoint(ctx context.Context, endpo
 	return endpoint, nil
 }
 
+// cleanVirtualServiceFields reset fields that should not be sent in update request
+// otherwise, the request will be rejected by the API server
+func (c *modelEndpointsService) cleanVirtualServiceFields(vs *v1beta1.VirtualService) {
```
deadlycoconuts, line 390: Thanks for figuring this out!

api/service/model_endpoint_service_test.go
```diff
@@ -377,6 +419,44 @@ func Test_modelEndpointsService_UpdateEndpoint(t *testing.T) {
...
+
+				mockIstio.On("PatchVirtualService", context.Background(), "project-1", vs).Return(vs, nil).Once()
+
+				mockVeStorage := s.versionEndpointStorage.(*storageMock.VersionEndpointStorage)
+				mockVeStorage.On("Get", updatedModelEndpointReq.Rule.Destination[0].VersionEndpointID).Return(newVersionEndpoint, nil)
+
+				mockMeStorage := s.modelEndpointStorage.(*storageMock.ModelEndpointStorage)
+				mockMeStorage.On("Save", context.Background(), mock.AnythingOfType("*models.ModelEndpoint"), mock.AnythingOfType("*models.ModelEndpoint")).Return(fmt.Errorf("failed to save"))
+
+				mockIstio.On("PatchVirtualService", context.WithoutCancel(context.Background()), "project-1", vs).Return(vs, nil).Once()
```
deadlycoconuts: ultra nit: lol this test doesn't really check if the `vs` object used for the rollback is the the original `vs` that was retrieved from `GetVirtualService` before it gets patched. In other words, we currently have

1. `GetVirtualService` returns `vs` 
2. `PatchVirtualService` patches `vs` with `vs` (ideally the new `vs` object should be somewhat different)
3. Roll back with another `PatchVirtualService` with the same `vs` again (ideally this should be the `vs` from step 1 and not the same `vs` in step 2)

Functionally the test checks all the functions that need to be called in such a scenario though, so everything else is good 😅

deadlycoconuts (approved):
Thanks a lot for investigating the bug and for implementing all the unit tests (and for setting a fixed helm version)! Yeah I think we have this problem of not making certain workflows not atomic enough in both Merlin and Turing, especially when a lot of these workflows have a long chain of steps and when one fails, which causes some state inconsistencies because we don't rollback/clean up properly. I'll probably add a card somewhere to remind us to scan/look through the entire repo again to decide which workflows need to atomic and to refactor them to prevent any of these odd bugs from occurring.

api/service/model_endpoint_service_test.go
bthari: lmao you're right 🥲 I've added the changes for this and also the undeploy method to validate several field to match the original VirtualService retrieved

api/service/model_endpoint_service.go
bthari: The idea is yes, you're right, I rollback it to make it atomic, but just to be clear this only happens when these two returns error
- `istioClient.CreateVirtualService`
- `s.modelEndpointStorage.Save`

The `err` variable on `if err := s.observabilityEventProducer.ModelEndpointChangeEvent` is a shadowed variable, so it will only be recognized within that `if` block; this rollback function will not catch the error. I don't know is that an intended case (let the inference service running, despite the observability fail) or not, so I just keep it as it is 😅

bthari:
@deadlycoconuts I've reverted back the changes for the helm/istio, istio has fixed their rendering on version `1.25.5`, so the github runner should run okay again using all the latest version 😄 but the e2e test has another issue with `bitnami/progresql` image now, seems the image has been removed from the docker.hub 🫠 I'll try to fix this first 

Yeah, there’s a lot of things going on when modifying the Merlin and Turing resource, and there might be some flow which isn’t atomic, we should take look into it again. 

And thank you for the throughout review!

bthari:
I pin the bitnami/postgres image to the currently available in the docker hub just so the e2e test past, for the longrun might be easier to update our helm-chart, seems the base repository for postgres itself is outdated 😵‍💫 I'll create a separate backlog ticket to scan all of our dependencies of bitnami image and handle the changes

If you don't have anymore comment @deadlycoconuts, I'll merge this soon (today or tomorrow). Once again, thank you for the thorough review!

PR #657, opened 2025-12-11
feat: Propagate detailed Kubernetes pod errors to endpoint message

Vishvajeet590:
# Description

This PR improves error visibility for model deployments by propagating detailed Kubernetes pod errors (such as OOMKilled, CrashLoopBackOff, ImagePullBackOff, etc.) to users. Previously, users only saw generic error messages like "predictor is not ready" or "CrashLoopBackOff" in the CaraML dashboard, making it difficult to diagnose deployment failures. With this change, users will see specific pod failure reasons, exit codes, and messages directly in the dashboard, enabling faster troubleshooting.

# Modifications

- Enhanced error handling in the deployment flow to include pod termination reason, exit code, and message in the error output.
- Updated the deployment logic to propagate these detailed Kubernetes errors to the `VersionEndpoint.Message` field.
- Ensured that the CaraML dashboard displays these detailed errors to users for any pod failure during deployment.

```

### DIFF
--- a/api/cluster/controller.go
+++ b/api/cluster/controller.go
@@ -407,6 +407,10 @@ func (c *controller) waitInferenceServiceReady(service *kservev1beta1.InferenceS
 			err = fmt.Errorf("%w\n\nPod container status:\n%s", err, podContainerTable)
 		}
 
+		if podLastTerminationReason != "" {
+			err = fmt.Errorf("%w\n\nPod last termination reason: %s", err, podLastTerminationReason)
+		}
+
 		if podLastTerminationMessage != "" {
 			err = fmt.Errorf("%w\n\nPod last termination message:\n%s", err, podLastTerminationMessage)
 		}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: api/cluster/controller.go [hunk 1]
```diff
@@ -407,6 +407,10 @@ func (c *controller) waitInferenceServiceReady(service *kservev1beta1.InferenceS
 			err = fmt.Errorf("%w\n\nPod container status:\n%s", err, podContainerTable)
 		}
 
+		if podLastTerminationReason != "" {
```
bthari, line 410: oh we have this all this time, we just didn't expose it, thanks for figuring this out!

python/pyfunc-server/pyfuncserver/protocol/upi/server.py
```diff
@@ -67,7 +67,13 @@ def start(self):
             publisher = Publisher(kafka_producer, sampler)
             self._predict_service.set_publisher(publisher)
 
-        asyncio.get_event_loop().run_until_complete(self._run_server())
+        self._run_server_sync()
+
+    def _run_server_sync(self):
```
bthari, line 72: why are we changing the previous one with this? 😮

Vishvajeet590: The **_run_server** method in UPIServer is an async coroutine being passed to **multiprocessing.Process** without being awaited.

The fix in UPIServer would be to wrap the async coroutine call with asyncio.run() in the multiprocessing target function

bthari (approved):

PR #659, opened 2025-12-15
feat: Exposing Liveness config

Vishvajeet590:
# Description
This PR introduces several improvements and fixes to the Merlin model deployment and endpoint management logic.

# Modifications

Sending the Liveness config values from the UI.


ui/src/pages/version/components/forms/components/LivnessConfigFormGroup.js
anantadwi13, line 1: nit: typo in filename

python/pyfunc-server/pyfuncserver/protocol/upi/server.py
anantadwi13, line 1: sorry if i'm not aware. why do we need to change this? is there a relation between liveness probe with this change?

PR #660, opened 2026-01-06
feat: Exposing Liveness config

Vishvajeet590:
# Description
This PR adds support for configurable liveness and readiness probes in Merlin model deployments, allowing users to customize probe parameters through both the UI and Python SDK. This enhancement provides users with fine-grained control over health check behavior for their model deployments

# Modifications

Sending the Liveness and Readiness probes config values from the UI. 

## Backend API changes
Extended ResourceRequest model to support two new optional fields:
 - liveness_probe
 - readiness_probe
 
Updated deployment templating so provided probes are rendered into Kubernetes resources

Relevant areas:
- api/service/deployment_service.go
- api/cluster/resource/templater.go

## UI changes
Added UI controls to configure liveness & readiness probe parameters.
Updated deployment details view to display saved probe configs after deployment (like CPU limits).

## Python SDK
Added liveness and readiness probe configuration support to the Python SDK, enabling programmatic probe customization

## Screenshot of Ui changes
image

### DIFF
--- a/python/pyfunc-server/docker/local.Dockerfile
+++ b/python/pyfunc-server/docker/local.Dockerfile
@@ -7,6 +7,7 @@ COPY ${MODEL_DEPENDENCIES_URL} conda.yaml
 
 ARG MERLIN_DEP_CONSTRAINT
 RUN process_conda_env.sh conda.yaml "merlin-pyfunc-server" "${MERLIN_DEP_CONSTRAINT}"
+RUN process_conda_env.sh conda.yaml "merlin-sdk" "${MERLIN_DEP_CONSTRAINT}"
 RUN conda env create --name merlin-model --file conda.yaml
 
 # Download and dry-run user model artifacts and code
--- a/api/cluster/resource/templater.go
+++ b/api/cluster/resource/templater.go
@@ -237,7 +237,16 @@ func (t *InferenceServiceTemplater) createPredictorSpec(modelService *models.Ser
 		}
 	}
 
-	livenessProbeConfig := getLivenessProbeConfig(modelService.PredictorProtocol(), envVars, fmt.Sprintf("/v1/models/%s", modelService.Name))
+	// Get user-configured probe settings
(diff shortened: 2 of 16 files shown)


python/pyfunc-server/docker/local.Dockerfile
```diff
@@ -7,6 +7,7 @@ COPY ${MODEL_DEPENDENCIES_URL} conda.yaml
 
 ARG MERLIN_DEP_CONSTRAINT
 RUN process_conda_env.sh conda.yaml "merlin-pyfunc-server" "${MERLIN_DEP_CONSTRAINT}"
+RUN process_conda_env.sh conda.yaml "merlin-sdk" "${MERLIN_DEP_CONSTRAINT}"
```
anantadwi13, line 10: sorry i'm not familiar with this change. based on your findings, why do we need to add this? and why don't we add the same line to the `Dockerfile`? is it because this `local.Dockerfile` is only used by the pipeline, but the `Dockerfile` is used by the production service?

anantadwi13 (approved):
lgtm, but i have 1 question about the docker file

anantadwi13:
thanks for the changes

anantadwi13:
anyway, could you attach some UI previews of this probe config? thanks

python/pyfunc-server/docker/local.Dockerfile
Vishvajeet590: I have observed it was used only in `local.Dockerfile`, i.e., is used by the pipeline to test pyfunc server.

Adding `RUN process_conda_env.sh conda.yaml "merlin-sdk" "${MERLIN_DEP_CONSTRAINT}"` ensures that the merlin-sdk package and its dependencies are included in the Conda environment. The error ModuleNotFoundError: No module named 'merlin.model_schema' occurs because the merlin-sdk (which provides merlin.model_schema) was not installed.

PR #665, opened 2026-04-22
feat: implement GRPC xDS on merlin standard transformer

anantadwi13:
# Description
This MR is used to introduce xDS GRPC client load balancer to Merlin Standard Transformer.
image
Previously, the transformer needs to call istio gateway to make sure all requests are balanced across `caraml-serving-*` pods. But now we support proxyless GRPC call via xDS to reduce gateway costs while still maintaining balanced request distribution.

More about xDS can be seen here https://istio.io/latest/blog/2021/proxyless-grpc/

To enable xDS, you need to configure your merlin config below
```yaml
...

StandardTransformerConfig:
  DefaultLabels:
    sidecar.istio.io/inject: "true"
  DefaultAnnotations:
    inject.istio.io/templates: grpc-agent
    proxy.istio.io/config: '{"holdApplicationUntilProxyStarts": true }'
  ImageName: ghcr.io/caraml-dev/merlin-transformer:x.x.x # replace x.x.x with version that supports xDS
  FeastRedisConfig:
    ...
    ServingURL: xds:///caraml-serving-redis.caraml-store.svc.cluster.local:50051 # make sure to use xds:///
  FeastBigtableConfig:
    ...
    ServingURL: xds:///caraml-serving-hbase.caraml-store.svc.cluster.local:50051 # make sure to use xds:///

...
```

Once you applied above config, when you deploy/redeploy a model, you will see the InferenceService resource looks like below
```yaml
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: test-pyfunc-model-1-r2
  namespace: caraml-testing-project
  ...
spec:
  predictor:
    containers:
      ...
    ...
  transformer:
    labels:
      sidecar.istio.io/inject: "true"
    annotations:
      inject.istio.io/templates: grpc-agent
      proxy.istio.io/config: '{"holdApplicationUntilProxyStarts": true}'
    containers:
    - env:
      - name: FEAST_STORAGE_CONFIGS
        value: '{ "1": { "redisCluster": { "feastServingUrl": "xds:///caraml-serving-redis.caraml-store.svc.cluster.local:50051" } }, "2": { "bigtable": { "feastServingUrl": "xds:///caraml-serving-hbase.caraml-store.svc.cluster.local:50051" } } }'
      ...
    ...
```

# Modifications
- introduce DNS & xDS client loadbalancing on the merlin standard transformer
- introduce `DefaultLabels` & `DefaultAnnotations` to merlin config `StandardTransformerConfig`. these fields will be used in the `InferenceService` creation.

# Tests

# Checklist
- [x] Added PR label
- [x] Tested locally
- [ ] Updated documentation

# Release Notes

```release-note
NONE
```

### DIFF
--- a/api/pkg/transformer/feast/client.go
+++ b/api/pkg/transformer/feast/client.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"strings"
 
 	"github.com/caraml-dev/merlin/pkg/transformer/feast/bigtablestore"
 	"github.com/caraml-dev/merlin/pkg/transformer/feast/redis"
@@ -56,14 +57,21 @@ func createFeastServingClient(feastOptions Options, featureTableMetadata []*spec
 }
 
 func newFeastGrpcClient(url string, options Options) (*GrpcClient, error) {
-	host, port, err := net.SplitHostPort(url)
-	if err != nil {
-		return nil, errors.Errorf("Unable to parse Feast Serving host (%s): %s", url, err)
-	}
+	{
+		// validate
 
-	portInt, err := strconv.Atoi(port)
-	if err != nil {
-		return nil, errors.Errorf("Unable to parse Feast Serving port (%s): %s", url, err)
+		urlTmp := url
+		urlTmp = strings.TrimPrefix(urlTmp, "dns:///")
+		urlTmp = strings.TrimPrefix(urlTmp, "xds:///")
+		_, port, err := net.SplitHostPort(urlTmp)
+		if err != nil {
+			return nil, errors.Errorf("Unable to parse Feast Serving host (%s): %s", url, err)
+		}
+
+		_, err = strconv.Atoi(port)
+		if err != nil {
+			return nil, errors.Errorf("Unable to parse Feast Serving port (%s): %s", url, err)
+		}
 	}
 
 	dialOpts := []grpc.DialOption{}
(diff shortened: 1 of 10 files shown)


Vishvajeet590 (approved):

api/pkg/transformer/feast/client.go
```diff
@@ -74,7 +82,10 @@ func newFeastGrpcClient(url string, options Options) (*GrpcClient, error) {
 		})
 		dialOpts = append(dialOpts, keepAliveOpt)
 	}
-	client, err := newInsecureGRPCClientWithDialOptions(host, portInt, options.FeastGRPCConnCount, true, dialOpts...)
+	if strings.HasPrefix(url, "dns:///") {
```
Vishvajeet590, line 85: round_robin is only applied for `dns:///` and xds:/// handles load balancing via the control plane. But plain `host:port` URLs like `online-serving.feast.ai.golabs.io:80` will silently use pick_first, meaning all Feast requests go to a single pod with no load balancing which is exactly the problem this PR is trying to solve.

Vishvajeet590 (changes requested):

api/pkg/transformer/feast/client.go
anantadwi13: yeah that's correct, if the user doesn't apply the URL with these prefixes (`dns:///` or `xds:///`), it will use the old behaviour which is `pick_first` loadbalancing (the default one)

PR #670, opened 2026-07-06
bug: fixing the endpoint msg string causing model in pending state

Vishvajeet590:
## What

Fix Postgres `invalid byte sequence for encoding "UTF8" (SQLSTATE 22021)` error when saving a version endpoint's deployment message.

## Why

When a deployment fails, the Kubernetes error text (which often contains multi-byte box-drawing chars like `─` = `0xe2 0x94 0x80`) is stored in `version_endpoints.message`. `sanitizeEndpoint` truncated it to `maxMessageChar` by **byte** slicing, which could cut a multi-byte character in half, leaving an invalid trailing sequence (e.g. `0xe2 0x94`). Postgres then rejected the write, so the endpoint status/message never updated:

## Error example  
``` 
merlin {"level":"error","ts":1783331067.766427,"caller":"work/model_service_deployment.go:164","msg":"unable to update endpoint status for model: external-demo, version: 16, reason: ERROR: invalid byte sequence for encoding \"UTF8\": 0xe2 0x94 (SQLSTATE 22021)","stacktrace":"github.com/caraml-dev/merlin/queue/work.(*ModelServiceDeployment).Deploy.func1\n\t/src/api/queue/work/model_service_deployment.go:164\ngithub.com/caraml-dev/merlin/queue/work.(*ModelServiceDeployment).Deploy\n\t/src/api/queue/work/model_service_deployment.go:189\ngithub.com/caraml-dev/merlin/queue.(*worker).processJob\n\t/src/api/queue/worker.go:81"}

```

### DIFF
--- a/api/storage/deployment_storage.go
+++ b/api/storage/deployment_storage.go
@@ -16,9 +16,11 @@ package storage
 
 import (
 	"fmt"
+	"strings"
 
 	"gorm.io/gorm"
 
+	"github.com/caraml-dev/merlin/log"
 	"github.com/caraml-dev/merlin/models"
 )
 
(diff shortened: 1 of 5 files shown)


anantadwi13 (approved):
lgtm. looks like the unit tests are already on the main branch 👀
https://github.com/caraml-dev/merlin/commit/b0ffd2fe8e1f538a10b92efd416854af025d9d99

